PR #4, opened 2024-05-30
Automated updates for components image digest

stefanprodan:
Changes:
- Fetch the images tags and digests from the `controlplaneio-fluxcd/distribution` Git repo on every reconciliation 
- Trigger an upgrade and pin the images to the tag and digest fetched from the distribution repo
- Record the last found images and their tags/digests in the `.status.components`

The automation and digest pinning works only for upstream (`ghcr.io/fluxcd`) and for the ControlPlane (`ghcr.io/controlplaneio-fluxcd`) container registries.


internal/builder/images.go
```diff
@@ -0,0 +1,134 @@
...
+		distro = "upstream-alpine"
+	case "ghcr.io/controlplaneio-fluxcd/alpine":
+		distro = "enterprise-alpine"
+	case "ghcr.io/controlplaneio-fluxcd/distroless":
+		distro = "enterprise-distroless"
+	default:
+		return nil, fmt.Errorf("unsupported registry: %s", registry)
+	}
+
+	const ghRepo = "https://raw.githubusercontent.com/controlplaneio-fluxcd/distribution/main/images"
```
souleb, line 89: maybe this should be a parameter? we could then fetch from a test http server in our tests? otherwise tests will fail everytime change the digests upstream.

stefanprodan: The tests will not fail, as they point to the upstream CNCF Flux images (fixed version v2.3.0). I'm reluctant on exposing the URL in the API, as this function should fetch the digests from the OCI repo directly, instead of using Git like now.

souleb (approved):
LGTM!

PR #12, opened 2024-05-31
Add revision to print columns

stefanprodan:
Add revision to print columns and format the reconcile duration in status.


souleb (approved):
LGTM!

PR #13, opened 2024-05-31
Add persistent storage for Flux artifacts

stefanprodan:
Add the `.spec.storage` optional field to the API for creating a PVC named `source-controller` that binds a volume to the source-controller pod to persist the Flux internal artifacts.


souleb (approved):
LGTM!

Issue #11, opened 2024-05-31
Document how to migrate a bootstrapped cluster  


stefanprodan:
With #10 the operator is now capable of taking ownership of the Flux installation, but if the cluster is bootstrapped, the `flux-system` Kustomization will fight with the operator on the same resources, so we need to document how to safely migrate.

PR #15, opened 2024-06-01
Takeover a bootstrapped cluster

stefanprodan:
This PR enables the operator to takeover the ownership of Flux controllers when deployed on a bootstrapped cluster. Once the operator deploys Flux, it disables the reconciliation of Flux CRDs and controllers by setting the  `kustomize.toolkit.fluxcd.io/ssa: Ignore` annotation so that kustomize-controller will skip all manifests in `gotk-components.yaml`.

Fix: #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,16 @@ kubectl create secret docker-registry flux-enterprise-auth \
   --docker-password=$ENTERPRISE_TOKEN
 ```
 
+### Migration of a bootstrap cluster
+
+To migrate a cluster that was bootstrapped, after the flux-operator is installed
+and the `FluxInstance` resource is created, the following steps are required:
+
+1. Checkout the branch of the Flux repository that was used to bootstrap the cluster.
+2. Replace the contents of the `flux-system/gok-components.yaml` with the `FluxInstance` YAML manifest.
+3. Remove all controllers patches from the `flux-system/kustomization.yaml`.
+4. Commit and push the changes to the Flux repository.
+
 ## Supply Chain Security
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -140,6 +140,16 @@ kubectl create secret docker-registry flux-enterprise-auth \
   --docker-password=$ENTERPRISE_TOKEN
 ```
 
+### Migration of a bootstrap cluster
+
+To migrate a cluster that was bootstrapped, after the flux-operator is installed
+and the `FluxInstance` resource is created, the following steps are required:
+
+1. Checkout the branch of the Flux repository that was used to bootstrap the cluster.
+2. Replace the contents of the `flux-system/gok-components.yaml` with the `FluxInstance` YAML manifest.
```
souleb, line 149: does this mean that we have to keep updating the Flux repository when updating the operator, i.e. changes to the `FluxInstance spec`? Couldn't we delete the `flux kustomization` and let the operator handle updates?

stefanprodan: Sure you could have the `FluxInstance` in TF state, or applied by OperatorHub or by any other tool/marketplace automation. We'll need extensive docs for how to do this depending on platform.

stefanprodan: > Couldn't we delete the flux kustomization and let the operator handle updates?

I guess you're refering to gotk-sync.yaml? If so, we can't delete it, that syncs all Flux CRs, we can only delete the gotk-components.yaml.

souleb: I think we should have that as a follow-up. This would be a really neat transition.

souleb (approved):
LGTM!

PR #16, opened 2024-06-03
Add support for disabling the reconciliation

stefanprodan:
Allow disabling the reconciliation by annotating  a `FluxInstance` with `fluxcd.controlplane.io/reconcile: disabled`.


souleb (approved):
LGTM!

PR #20, opened 2024-06-10
Implement cluster sync feature

stefanprodan:
This PR adds an optional field called `sync` to the `FluxInstance` API. 

### Sync configuration

The `.spec.sync` field is optional and specifies the Flux sync configuration.
When set, a Flux source and a Flux Kustomization are generated to sync
the cluster state with the source repository.

The Flux objects are created in the same namespace where the FluxInstance is deployed
using the namespace name as the Flux source and Kustomization name. The naming convention
matches the one used by `flux bootstrap` to ensure compatibility with upstream, and
to allow transitioning a bootstrapped cluster to a FluxInstance managed one.

Sync fields:

- `kind`: The source kind, supported values are `GitRepository`, `OCIRepository` and `Bucket`.
- `url`: The URL of the source repository, can be a Git repository HTTP/S or SSH address, an OCI repository address or a Bucket endpoint.
- `ref`: The source reference, can be a Git ref name e.g. `refs/heads/main`, an OCI tag e.g. `latest` or a Bucket name.
- `path`: The path to the source directory containing the kustomize overlay or plain Kubernetes manifests to sync from.
- `pullSecret`: The name of the Kubernetes secret that contains the credentials to pull the source repository. This field is optional.
- `interval`: The sync interval. This field is optional, when not set the default is `1m`.

### Sync from Git over HTTP/S

Example:

```yaml
spec:
  sync:
    kind: GitRepository
    url: "https://gitlab.com/my-group/my-fleet.git"
    ref: "refs/heads/main"
    path: "clusters/my-cluster"
    pullSecret: "git-token-auth"
```

If the source repository is private, the Kubernetes secret must be created
in the same namespace where the FluxInstance is deployed, and have the following format:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: git-token-auth
  namespace: flux-system
type: Opaque
stringData:
  username: "git-username"
  password: "git-token"
```

To generate the secret with the Flux CLI:

```sh
flux create secret git git-token-auth \
  --namespace flux-system \
  --url=https://gitlab.com/my-group/my-fleet.git \
  --username=git-username \
  --password=git-token
```

### Sync from Git over SSH

Example:

```yaml
spec:
  sync:
    kind: GitRepository
    url: "ssh://git@github.com/my-org/my-fleet.git"
    ref: "refs/heads/main"
    path: "clusters/my-cluster"
    pullSecret: "git-ssh-auth"
```

If the source repository is private, the Kubernetes secret must be created
[...]

PR #25, opened 2024-06-11
s/gok-components/gotk-components/ in README

kingdonb:
Fix typo in README


stefanprodan:
@kingdonb this will go away once #20 is merged and released. With the new `.spec.sync` feature, the whole flux-system dir can be deleted.

Issue #28, opened 2024-06-13
Cannot apply kustomize patches in flux-system deployments


xsvil:
Hello,

I've recently upgraded to the FluxCD operator from Terraform FluxCD bootstrap, and I am encountering an issue related to kustomize patches.

I have followed the migration guide and created a FluxInstance. While it creates deployments with a default template, it sets the annotation **kustomize.toolkit.fluxcd.io/ssa: Ignore** on these deployments. This annotation blocks any kustomize patches from being applied. This is not the case when bootstrapping via flux cli or terraform.

For example, I have added the default affinity configuration as per the official documentation here:
[Node Affinity and Tolerations](https://fluxcd.io/flux/installation/configuration/vertical-scaling/#node-affinity-and-tolerations)

However, these settings do not affect the flux-system deployments due to the **kustomize.toolkit.fluxcd.io/ssa: Ignore**  annotation.

Is this by design and is there a way to patch all flux-system deployments?

Thanks!

stefanprodan:
@xsvil you would place the patches in the FluxInstance definition, see https://fluxcd.control-plane.io/operator/fluxinstance/#kustomize-patches

xsvil:
@stefanprodan I have added the kustomize patches. All kustomizations seem to be working, except the one specifically for affinity and tolerations:
<img width="1512" alt="Screenshot 2024-06-13 at 17 50 14" src="https://github.com/controlplaneio-fluxcd/flux-operator/assets/63684363/328eb172-6c5b-4fe0-8e34-56f3a7da04a9">

I literally migrated my kustomizations and added them on the FluxInstance. After the migration flux-system deployments are no longer on the group: system-critical node I have specifically created. And also if i describe any deployment or pod it doesn't contain affinity and tolerations.

stefanprodan:
@xsvil can you please remove the `labelSelector`, all deployments are Flux controllers so we don't need that anymore.

xsvil:
Thank you @stefanprodan ! That seems to have worked perfectly and rescheduled all flux-system deployments on the `group:system-critical` node. I thought it will still work only because all deployments still have the same label `app.kubernetes.io/part-of=flux`. Thanks again!

stefanprodan:
I need to document this before we close this issue, basically the labels are injected by the same kustomization so they don't exist in the base manifests, hence targeting should be done by `kind` and `name` only.

PR #29, opened 2024-06-13
Add `part-of` label to controllers base

stefanprodan:
Rebuild the controllers kustomize base with https://github.com/fluxcd/flux2/pull/4842

Fix: #28


souleb (approved):
LGTM!

PR #33, opened 2024-06-15
Implement automatic distribution updates from OCI artifacts

stefanprodan:
This PR adds a new optional filed named `.spec.distribution.artifact` to the FluxInstance API. When specified, the operator will pull the artifact from the registry on a regular interval to determine the latest Flux version available including CVE patches and hotfixes.

Using this feature, CNCF Flux users can keep Flux up-to-date without having to update the operator every time there is a new Flux version available. 

```yaml
kind: FluxInstance
metadata:
  name: flux
  namespace: flux-system
spec:
  distribution:
    version: "2.x"
    registry: "ghcr.io/fluxcd"
    artifact: "oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests"
```

For enterprise customers, this feature allows for automated updates of CVE patches to fixed versions and semver ranges without having to update the operator.

```yaml
kind: FluxInstance
metadata:
  name: flux
  namespace: flux-system
spec:
  distribution:
    version: "2.3.0"
    registry: "ghcr.io/controlplaneio-fluxcd/distroless"
    imagePullSecret: "flux-enterprise-auth"
    artifact: "oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests"
```


hack/vendor-flux-manifests.sh
```diff
@@ -30,4 +31,14 @@ do
     vendor "$var"
 done
 
+info "downloading distro repository"
+curl -sLO https://github.com/controlplaneio-fluxcd/distribution/archive/refs/heads/main.tar.gz
+tar xzf main.tar.gz -C "${DEST_DIR}"
+
+mkdir -p "${IMG_DIR}"
```
souleb, line 38: maybe should add a check here?

internal/builder/pull.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/fluxcd/pkg/tar"
+	"github.com/google/go-containerregistry/pkg/crane"
+)
+
+// PullArtifact downloads an artifact from an OCI repository and extracts the content
+// of the first tgz layer to the given destination directory.
```
souleb, line 16:
```suggestion
// of the first tgz layer to the given destination directory.
// The function returns the digest of the artifact.
```

internal/builder/options.go
```diff
@@ -21,6 +21,7 @@ type Options struct {
 	Patches                string
 	ArtifactStorage        *ArtifactStorage
 	Sync                   *Sync
+	ManifestsArtifactURL   string
```
souleb, line 24: This seems unused.

internal/controller/fluxinstance_controller.go
```diff
@@ -178,18 +204,53 @@ func (r *FluxInstanceReconciler) reconcile(ctx context.Context,
...
+	// Pull the latest manifests from the OCI repository.
+	if artifactURL != "" {
+		ctxPull, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		artifactDigest, err := builder.PullArtifact(ctxPull, artifactURL, tmpDir)
+		if err != nil {
+			return "", err
+		}
+		log.Info("latest manifests pulled", "url", artifactURL, "digest", artifactDigest)
```
souleb, line 222:
```suggestion
		log.Info("fetched latest manifests, "url", artifactURL, "digest", artifactDigest)
```

api/v1/fluxinstance_types.go
```diff
@@ -83,6 +83,13 @@ type Distribution struct {
 	// to use for pulling images.
 	// +optional
 	ImagePullSecret string `json:"imagePullSecret,omitempty"`
+
+	// Artifact is the URL to the OCI artifact containing
+	// the latest Kubernetes manifests for the distribution,
+	// e.g. 'oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:latest'.
+	// +kubebuilder:validation:Pattern="^oci://.*$"
+	// +optional
+	Artifact string `json:"artifact,omitempty"`
```
souleb, line 92: can we name it maybe `artifactURL`?

stefanprodan: I did consider this but given that we use `registry` and not `registryURL`, for consistency I chose to not have the URL suffix.

hack/vendor-flux-manifests.sh
stefanprodan: `-p` will not create the dir if it exists.

internal/controller/fluxinstance_controller.go
```diff
@@ -178,18 +204,55 @@ func (r *FluxInstanceReconciler) reconcile(ctx context.Context,
...
+	obj *fluxcdv1.FluxInstance, tmpDir string) (string, error) {
+	log := ctrl.LoggerFrom(ctx)
+	artifactURL := obj.Spec.Distribution.Artifact
+
+	// Pull the latest manifests from the OCI repository.
+	if artifactURL != "" {
+		ctxPull, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		artifactDigest, err := builder.PullArtifact(ctxPull, artifactURL, tmpDir)
```
souleb, line 221: If I got it correctly, at every reconciliation we fetch the manifests and don't store it locally. What do you think about storing the `artifact` and its `manifest` digest, at each reconciliation, we would first fetch the `manifest` and compare the digest with local version before attempting to fetch the `blob`?

stefanprodan: Yes we could have a similar storage & auth implementation as SC in a followup PR. I'm also considering using the layer caching I implemented in Timoni which is more close to how OverlayFS works in OCI registries.

souleb (approved):
LGTM!

PR #39, opened 2024-06-17
Add publishing procedure for manifests

stefanprodan:
Document how to publish the OCI artifacts used for automated upgrades when:
- A new Flux version is released
- A new CVE fix is released for the enterprise distribution


souleb (approved):
LGTM!

Issue #9, opened 2024-05-30
Publish the project on AWS marketplace (EKS add-on)


stefanprodan:
Push the flux-operator Helm chart to the ControlPlane AWS marketplace account and register the operator as an [EKS add-on](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html).

PR #42, opened 2024-06-18
Implement entitlement verification and AWS metering

stefanprodan:
This PR adds entitlement verification and metering for AWS Marketplace service.

Ref: #9 

PS. When the operator is not installed from the AWS Marketplace, an offline entitlement client is employed that generates
a perpetual license token.


internal/entitlement/aws.go
```diff
@@ -0,0 +1,107 @@
...
+// ControlPlane Enterprise for Flux CD AWS Marketplace product.
+// https://aws.amazon.com/marketplace/pp/prodview-ndm54wno7tayg
+type AmazonClient struct {
+	Vendor string
+	mc     *marketplacemetering.Client
+}
+
+// NewAmazonClient creates a new AmazonClient using the default
+// AWS configuration and the current region.
+func NewAmazonClient(vendor string) (*AmazonClient, error) {
```
souleb, line 48: It would be nice to have a test for this. I believe we can use `WithEndpointResolverV2()` for that purpose.

```golang
type endpointResolverV2 struct {
	URL url.URL
}

func (e *endpointResolverV2) ResolveEndpoint(ctx context.Context, params marketplacemetering.EndpointParameters) (smithyendpoints.Endpoint, error) {
	return smithyendpoints.Endpoint{
		URI: e.URL,
	}, nil
}

func TestMarketplaceClient() {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		...
	}))
	defer srv.Close()
	opt := marketplacemetering.WithEndpointResolverV2(&endpointResolverV2{URL: url.URL{Scheme: "http", Host: srv.URL}})
	mc := marketplacemetering.NewFromConfig(cfg, opt)
	...
}
```

internal/entitlement/client.go
```diff
@@ -0,0 +1,62 @@
...
+	// TokenKey is the key in the entitlement secret
+	// that holds the token.
+	TokenKey = "token"
+
+	// DefaultVendor is the default vendor name.
+	DefaultVendor = "controlplane"
+
+	// MarketplaceTypeEnvKey is the environment variable key
+	// that holds the marketplace type.
+	MarketplaceTypeEnvKey = "MARKETPLACE_TYPE"
```
souleb, line 27: I did not find this env var in the flags, how is it setup?

stefanprodan: The Helm chart used in the Marketplace sets the evn var:
- https://github.com/controlplaneio-fluxcd/charts/blob/32c902d67dd251ab1a18118b7201e420ea3c9754/charts/flux-operator/values.yaml#L86-L90
- https://github.com/controlplaneio-fluxcd/charts/blob/32c902d67dd251ab1a18118b7201e420ea3c9754/charts/flux-operator/templates/deployment.yaml#L53-L64

internal/entitlement/aws.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/marketplacemetering"
+	"github.com/golang-jwt/jwt/v4"
+)
+
+const (
+	// AmazonMarketplaceProductCode is the AWS Marketplace
+	// product code of ControlPlane Enterprise for Flux CD.
+	AmazonMarketplaceProductCode = "272knt6mdmtctbck10givwm1h"
```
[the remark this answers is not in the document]
stefanprodan: I made them private

internal/entitlement/aws.go
stefanprodan: Mocking the AWS APIs could come later after we release a version with metering and we tests it on EKS.

stefanprodan: I've tested this PR on EKS with an IRSA for `AWSMarketplaceMeteringRegisterUsage` and all works, the usage is registered and the resulting token is valid 🎉

internal/controller/entitlement_controller.go
```diff
@@ -0,0 +1,218 @@
...
+			if errNew != nil {
+				return nil, fmt.Errorf("failed to create %s: %w", secretName, errNew)
+			}
+			return newSecret, nil
+		} else {
+			return nil, fmt.Errorf("failed to init %s: %w", secretName, err)
+		}
+	}
+
+	exitingVendor, found := secret.Data[entitlement.VendorKey]
```
souleb, line 163: should it be `existingVendor`?

souleb (approved):
LGTM!

internal/controller/entitlement_controller.go
stefanprodan: Thanks, fixed!

Issue #44, opened 2024-06-20
Multitenant Service Account


blezoray:
Today, in version 0.5.0, we can change the multitenant Service Account using kustomize patch, like this:
```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: FluxInstance
spec:
...
  cluster:

    type: openshift
    multitenant: true
    networkPolicy: true
    domain: "cluster.local"
...
  kustomize:
    patches:
      - patch: |
          - op: add
            path: /spec/template/spec/containers/0/args/-
            value: --default-service-account=flux
        target:
          kind: Deployment
          name: "(kustomize-controller|helm-controller)"
```

As the result, the kustomize|helm controllers have these args (`flux` SA overwrite `default` SA):
```yaml
  spec:
    containers:
      - args:
          - --events-addr=http://notification-controller.mpms-flux-operator.svc.cluster.local./
          - --watch-all-namespaces=true
          - --log-level=info
          - --log-encoding=json
          - --enable-leader-election
          - --no-cross-namespace-refs=true
          - --no-remote-bases=true
          - --default-service-account=default
          - --concurrent=10
          - --requeue-dependency=5s
          - --default-service-account=flux
```

Is it possible to add a new parameter, like `spec.cluster.multitenantServiceAccountName` in order to specify the name of this SA and to avoid any confusion ?

Rgds, Bruno.

erikgb:
I support this! We used to patch in all changes required for multi-tenant lock-down. When enabling the multitenant mode using the operator, we get double flags for the default service account - which could be a bit confusing for some users.

````yaml
      args:
        - '--events-addr=http://notification-controller.flux-system.svc.cluster.local./'
        - '--watch-all-namespaces=true'
        - '--log-level=info'
        - '--log-encoding=json'
        - '--enable-leader-election'
        - '--no-cross-namespace-refs=true'
        - '--no-remote-bases=true'
        - '--default-service-account=default'
        - '--default-service-account=gitops-reconciler'
````

PR #63, opened 2024-07-03
Add `tenantDefaultServiceAccount` to FluxInstance API

stefanprodan:
This PR adds an optional field `.spec.cluster.tenantDefaultServiceAccount` to the FluxInstance API that allows setting the
default service account used by Flux when reconciling `Kustomization` and `HelmRelease` resources found in the tenant namespaces.

Closes: #44


blezoray:
Great. thanks a lot.

PR #98, opened 2024-09-24
Introduce `ResourceSet` API

stefanprodan:
**ResourceSet** is a declarative API for generating a group of Kubernetes objects
based on a matrix of input values and a set of templated resources.

The ResourceSet API offers a high-level abstraction for defining and managing
Flux resources and related Kubernetes objects as a single unit. It is designed
to reduce the complexity of Kustomize overlays by providing a compact way
of defining different configurations for a set of workloads per tenant and/or environment.

Use cases:

- Application definition: Bundle a set of Kubernetes resources (Flux HelmRelease, OCIRepository, Alert, Provider, Receiver, ImagePolicy) into a single deployable unit.
- Dependency management: Define dependencies between apps to ensure that the resources are applied in the correct order. The dependencies are more flexible  than in Flux, they can be for other ResourceSets, CRDs, or any other Kubernetes object.
- Multi-instance provisioning: Generate multiple instances of the same application with different configurations.
- Multi-cluster provisioning: Generate multiple instances of the same application for each target cluster that are deployed by Flux from a management cluster.
- Multi-tenancy provisioning: Generate a set of resources (Namespace, ServiceAccount, RoleBinding) for each tenant with specific roles and permissions.

## Example

The following example shows a ResourceSet that generates an application instance consisting of a
Flux HelmRelease and OCIRepository for each tenant with a specific version and replica count.

```yaml
kind: ResourceSet
metadata:
  name: podinfo
  namespace: default
  annotations:
    fluxcd.controlplane.io/reconcileEvery: "30m"
    fluxcd.controlplane.io/reconcileTimeout: "5m"
spec:
  commonMetadata:
    labels:
      app.kubernetes.io/name: podinfo
  inputs:
    - tenant: "team1"
      app:
       version: "6.7.x"
       replicas: 2
    - tenant: "team2"
      app:
       version: "6.6.x"
       replicas: 3
  resources:
    - apiVersion: source.toolkit.fluxcd.io/v1beta2
      kind: OCIRepository
      metadata:
        name: podinfo-<< inputs.tenant >>
        namespace: default
      spec:
        interval: 10m
        url: oci://ghcr.io/stefanprodan/charts/podinfo
        ref:
          semver: << inputs.app.version | quote >>
    - apiVersion: helm.toolkit.fluxcd.io/v2
      kind: HelmRelease
      metadata:
        name: podinfo-<< inputs.tenant >>
        namespace: default
      spec:
        interval: 1h
        releaseName: podinfo-<< inputs.tenant >>
        chartRef:
          kind: OCIRepository
          name: podinfo-<< inputs.tenant >>
        values:
          replicaCount: << inputs.app.replicas | int >>
```

You can run this example by saving the manifest into `podinfo.yaml`.

1. Apply the ResourceSet on the cluster:

   ```shell
   kubectl apply -f podinfo.yaml
   ```

2. Wait for the ResourceSet to reconcile the generated resources:

   ```shell
   kubectl wait resourceset/podinfo --for=condition=ready --timeout=5m
   ```

3. Run `kubectl get resourceset` to see the status of the resource:

   ```console
    $ kubectl get resourceset
    NAME      AGE   READY   STATUS
    podinfo   59s   True    Reconciliation finished in 52s
   ```

4. Run `kubectl describe resourceset` to see the reconciliation status conditions and events:

    ```console
    $ kubectl describe resourceset podinfo
    Status:
      Conditions:
        Last Transition Time:  2024-09-24T09:58:53Z
        Message:               Reconciliation finished in 52s
        Observed Generation:   1
        Reason:                ReconciliationSucceeded
        Status:                True
        Type:                  Ready
    Events:
      Type    Reason          Age   From           Message
      ----    ------          ----  ----           -------
      Normal  ApplySucceeded  72s   flux-operator  HelmRelease/default/podinfo-team1 created
                                                   HelmRelease/default/podinfo-team2 created
                                                   OCIRepository/default/podinfo-team1 created
                                                   OCIRepository/default/podinfo-team2 created
      Normal  ReconciliationSucceeded  72s  flux-operator  Reconciliation finished in 52s
    ```
 
5. Run `kubectl events` to see the events generated by the flux-operator:

    ```shell
    kubectl events --for resourceset/podinfo
    ```

6. Run `kubectl delete` to remove the ResourceSet and its generated resources:

    ```shell
    kubectl delete resourceset podinfo
    ```

## Writing a ResourceSet spec

As with all other Kubernetes config, a ResourceSet needs `apiVersion`,
`kind`, `metadata.name` and `metadata.namespace` fields.
The name of a ResourceSet object must be a valid [DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
A ResourceSet also needs a [`.spec` section](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).

### Inputs configuration

The `.spec.inputs` field is optional and specifies a list of input values
to be used in the resources templates.

Example inputs:

```yaml
spec:
  inputs:
   - tenant: team1
     role: restricted
   - tenant: team2
     role: privileged
```

An input value is a key-value pair of strings and structs, where the key is the input name
which can be referenced in the resource templates using the `<< inputs.name >>` syntax.

### Resources configuration

The `.spec.resources` field is optional and specifies the list of Kubernetes resource
to be generated and reconciled on the cluster.

Example of plain resources without any templating:

```yaml
spec:
  resources:
   - apiVersion: v1
     kind: Namespace
     metadata:
      name: apps
   - apiVersion: v1
     kind: ServiceAccount
     metadata:
      name: flux
      namespace: apps
```

#### Templating resources

The resources can be templated using the `<< inputs.name >>` syntax. The templating engine
is based on Go text template. The `<<  >>` delimiters are used instead of `{{  }}` to avoid
conflicts with Helm templating and allow ResourceSets to be included in Helm charts.

Example of templated resources:

```yaml
spec:
  inputs:
   - tenant: team1
     role: admin
   - tenant: team2
     role: cluster-admin
  resources:
   - apiVersion: v1
     kind: Namespace
     metadata:
      name: << inputs.tenant >>
   - apiVersion: v1
     kind: ServiceAccount
     metadata:
      name: flux
      namespace: << inputs.tenant >>
   - apiVersion: rbac.authorization.k8s.io/v1
     kind: RoleBinding
     metadata:
      name: flux
      namespace: << inputs.tenant >>
     subjects:
      - kind: ServiceAccount
        name: flux
        namespace: << inputs.tenant >>
        roleRef:
         kind: ClusterRole
         name: << inputs.role >>
         apiGroup: rbac.authorization.k8s.io
```

The above example will generate a `Namespace`, `ServiceAccount` and `RoleBinding` for each tenant
with the specified role.

#### Templating functions

The templating engine supports [slim-sprig](https://go-task.github.io/slim-sprig/) functions.

It is recommended to use the `quote` function when templating strings to avoid issues with
special characters e.g. `<< inputs.version | quote >>`.

When templating integers, use the `int` function to convert the string to an integer
e.g. `<< inputs.replicas | int >>`.

When templating booleans, use the `bool` function to convert the string to a boolean
e.g. `<< inputs.enabled | bool >>`.

When using integer or boolean inputs as metadata label values, use the `quote` function to convert
the value to a string e.g. `<< inputs.enabled | quote >>`.

When templating nested fields, use the `toYaml`  and `nindent` functions
to properly format the string e.g.:

```yaml
spec:
  inputs:
    - tenant: team1
      layerSelector:
        mediaType: "application/vnd.cncf.helm.chart.content.v1.tar+gzip"
        operation: copy
  resources:
    - apiVersion: source.toolkit.fluxcd.io/v1beta2
      kind: OCIRepository
      metadata:
        name: << inputs.tenant >>
      spec:
        layerSelector: << inputs.layerSelector | toYaml | nindent 4 >>
```

In addition to the slim-sprig functions, a `slugify` function is available to normalize a string for use in a Kubernetes
[label value](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
e.g. `<< inputs.tenant | slugify >>`.

#### Resource deduplication

The flux-operator deduplicates resources based on the
`apiVersion`, `kind`, `metadata.name` and `metadata.namespace` fields.

This allows defining shared resources that are applied only once, regardless of the number of inputs.

Example of a shared Flux source:

```yaml
spec:
  inputs:
    - tenant: "team1"
      replicas: "2"
    - tenant: "team2"
      replicas: "3"
  resources:
    - apiVersion: source.toolkit.fluxcd.io/v1beta2
      kind: OCIRepository
      metadata:
        name: podinfo
        namespace: default
      spec:
        interval: 10m
        url: oci://ghcr.io/stefanprodan/charts/podinfo
        ref:
          semver: '*'
    - apiVersion: helm.toolkit.fluxcd.io/v2
      kind: HelmRelease
      metadata:
        name: podinfo-<< inputs.tenant >>
        namespace: default
      spec:
        interval: 1h
        releaseName: podinfo-<< inputs.tenant >>
        chartRef:
          kind: OCIRepository
          name: podinfo
        values:
          replicaCount: << inputs.replicas | int >>
```

In the above example, the `OCIRepository` resource is created only once
and referred by all `HelmRelease` resources.

#### Conditionally resource exclusion

To exclude a resource based on input values, the `fluxcd.controlplane.io/reconcile` annotation can be set 
to `disabled` on the resource metadata. This will prevent the resource from being reconciled by the operator.

Example of excluding a resource based on an input value:

```yaml
spec:
  inputs:
    - tenant: "team1"
    - tenant: "team2"
  resources:
    - apiVersion: v1
      kind: Namespace
      metadata:
        name: << inputs.tenant >>
    - apiVersion: v1
      kind: ServiceAccount
      metadata:
        name: flux
        namespace: << inputs.tenant >>
        annotations:
          fluxcd.controlplane.io/reconcile: << if eq inputs.tenant "team1" >>enabled<< else >>disabled<< end >>
```

In the above example, the `ServiceAccount` resource is generated only for the `team1` tenant.

### Resources template

The `.spec.resourcesTemplate` field is optional and offers an alternative to the `.spec.resources`.
The `.spec.resourcesTemplate` is a single string that contains the multi-document YAML of the resources
definitions. This field can be used for complex templating scenarios with the trade-off of reduced readability.

Note that when both `.spec.resources` and `.spec.resourcesTemplate` are set, the resulting resources
are the union of the two. If duplicate resources are defined in both fields, the resources from
`.spec.resources` take precedence.

Example of a template containing conditional and repeated blocks:

```yaml
spec:
  inputs:
    - bundle: addons
      decryption: false
      components:
        - ingress-nginx
        - cert-manager
    - bundle: apps
      decryption: true
      components:
        - frontend
        - backend
  resourcesTemplate: |
    ---
    apiVersion: source.toolkit.fluxcd.io/v1beta2
    kind: OCIRepository
    metadata:
      name: << inputs.bundle >>
      namespace: flux-system
    spec:
      interval: 10m
      url: oci://registry.example.com/<< inputs.bundle >>
    <<- range $component := inputs.components >>
    ---
    apiVersion: kustomize.toolkit.fluxcd.io/v1
    kind: Kustomization
    metadata:
      name: << $component >>
      namespace: flux-system
    spec:
      interval: 1h
      prune: true
      <<- if inputs.decryption >>
      decryption:
        provider: sops
        secretRef:
          name: << inputs.bundle >>-sops
      <<- end >>
      sourceRef:
        kind: OCIRepository
        name: << inputs.bundle >>
      path: ./<< $component >>
    <<- end >>
```

The above example generates two `OCIRepository` resources (one for each bundle) and four
`Kustomization` resources (one for each component in each bundle).

### Common metadata

The `.spec.commonMetadata` field is optional and specifies common metadata to be applied to all resources.

It has two optional fields:

- `labels`: A map used for setting [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)
  on an object. Any existing label will be overridden if it matches with a key in
  this map.
- `annotations`: A map used for setting [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
  on an object. Any existing annotation will be overridden if it matches with a key
  in this map.

Example common metadata:

```yaml
spec:
  commonMetadata:
    labels:
      app.kubernetes.io/name: podinfo
    annotations:
      fluxcd.controlplane.io/prune: disabled
```

In the above example, all resources generated by the ResourceSet
will not be pruned by the [garbage collection](#garbage-collection) process as
the `fluxcd.controlplane.io/prune` annotation is set to `disabled`.

### Dependency management

`.spec.dependsOn` is an optional list used to refer to Kubernetes
objects that the ResourceSet depends on. If specified, then the ResourceSet
is reconciled after the referred objects exist in the cluster.

A dependency is a reference to a Kubernetes object with the following fields:

- `apiVersion`: The API version of the referred object (required).
- `kind`: The kind of the referred object (required).
- `name`: The name of the referred object (required).
- `namespace`: The namespace of the referred object (optional).
- `ready`: A boolean indicating if the referred object must have the `Ready` status condition set to `True` (optional, default is `false`).
- `readyExpr`: A [CEL expression](#cel-readiness-expressions) that evaluates to a boolean indicating if the referred object is ready (optional).

Example of conditional reconciliation based on the existence of CustomResourceDefinitions
and the readiness of a ResourceSet:

```yaml
spec:
  dependsOn:
    - apiVersion: apiextensions.k8s.io/v1
      kind: CustomResourceDefinition
      name: helmreleases.helm.toolkit.fluxcd.io
      ready: true
    - apiVersion: apiextensions.k8s.io/v1
      kind: CustomResourceDefinition
      name: servicemonitors.monitoring.coreos.com
    - apiVersion: fluxcd.controlplane.io/v1
      kind: ResourceSet
      name: cluster-addons
      namespace: flux-system
      ready: true
```

Note that is recommended to define dependencies on CustomResourceDefinitions if the ResourceSet
deploys Flux HelmReleases which contain custom resources.

When the dependencies are not met, the flux-operator will reevaluate the requirements
every five seconds and reconcile the ResourceSet when the dependencies are satisfied.
Failed dependencies are reported in the ResourceSet `Ready` [status condition](#ResourceSet-Status),
in log messages and Kubernetes events.

#### CEL readiness expressions

The `readyExpr` field allows for more complex readiness checks and
can be used for gating the reconciliation of a ResourceSet based on the evaluation
of the [CEL](https://cel.dev) expression.

The expression is evaluated in the context of the referred object and has access to all the fields of the object,
including the status conditions and the status subfields. The expression must evaluate to a boolean value, any syntax
or runtime errors will be reported in the ResourceSet status conditions.

Example readiness expression:

```yaml
spec:
  dependsOn:
    - apiVersion: cluster.x-k8s.io/v1beta1
      kind: Cluster
      name: my-cluster
      namespace: dev
      ready: true
      readyExpr: |
        metadata.generation == status.observedGeneration &&
        status.controlPlaneReady == true
    - apiVersion: v1
      kind: Secret
      name: my-gate
      namespace: dev
      ready: true
      readyExpr: |
        string(base64.decode(data.gate)) == 'opened'
```

For testing the CEL expressions, you can use the [CEL playground](https://playcel.undistro.io/).

### Reconciliation configuration

The reconciliation of behaviour of a ResourceSet can be configured using the following annotations:

- `fluxcd.controlplane.io/reconcile`: Enable or disable the reconciliation loop. Default is `enabled`, set to `disabled` to pause the reconciliation.
- `fluxcd.controlplane.io/reconcileEvery`: Set the reconciliation interval used for drift detection and correction. Default is `1h`.
- `fluxcd.controlplane.io/reconcileTimeout`: Set the reconciliation timeout including health checks. Default is `5m`.

### Health check configuration

The `.spec.wait` field is optional and instructs the flux-operator to perform
a health check on all applied resources and waits for them to become ready. The health
check is disabled by default and can be enabled by setting the `.spec.wait` field to `true`.

The health check is performed for the following resources types:

- Kubernetes built-in kinds: Deployment, DaemonSet, StatefulSet,
  PersistentVolumeClaim, Service, Ingress, CustomResourceDefinition.
- Flux kinds: HelmRelease, OCIRepository, Kustomization, GitRepository, etc.
- Custom resources that are compatible with [kstatus](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus).

By default, the wait timeout is `5m` and can be changed with the
`fluxcd.controlplane.io/reconcileTimeout` annotation, set on the ResourceSet object.

### Role-based access control

The `.spec.serviceAccountName` field is optional and specifies the name of the
Kubernetes ServiceAccount used by the flux-operator to reconcile the ResourceSet.
The ServiceAccount must exist in the same namespace as the ResourceSet
and must have the necessary permissions to create, update and delete
the resources defined in the ResourceSet.

On multi-tenant clusters, it is recommended to use a dedicated ServiceAccount per tenant namespace
with the minimum required permissions. To enforce a ServiceAccount for all ResourceSets,
the `--default-service-account=flux-operator`flag can be set in the flux-operator container arguments.
With this flag set, only the ResourceSets created in the same namespace as the flux-operator
will run with cluster-admin permissions.

### Garbage collection

The operator performs garbage collection of the resources previously generated by a ResourceSet
that are no longer present in the current revision. Garbage collection is also performed
when a ResourceSet object is deleted, triggering a removal of all Kubernetes
objects previously applied on the cluster.

The garbage collection process removes stale resources in stages, first it deletes the Flux
custom resources and waits for the Flux Kustomizations and HelmReleases to be finalized by
the controllers. After the Flux resources are removed (max wait is one minute), the operator
proceeds with the deletion of the remaining Kubernetes objects. This ensures that the
Flux controllers have a chance to clean up the resources they manage before the operator
deletes the Kubernetes ServiceAccount and RoleBinding used by Flux impersonation.

After the garbage collection process is completed, the operator issues a Kubernetes event
containing the list of removed resources and the duration of the cleanup.

The garbage collection is enabled by default and can be disabled for certain resources
by setting the `fluxcd.controlplane.io/prune` annotation to `disabled`. To fully disable
the garbage collection for a ResourceSet, the annotation must be set on all resources
using the [`.spec.commonMetadata`](#common-metadata) field.

### Conditions

A ResourceSet enters various states during its lifecycle, reflected as Kubernetes Conditions.
It can be [reconciling](#reconciling-fluxinstance) while applying the
resources on the cluster, it can be [ready](#ready-fluxinstance), or it can [fail during
reconciliation](#failed-fluxinstance).

The ResourceSet API is compatible with the **kstatus** specification,
and reports `Reconciling` and `Stalled` conditions where applicable to
provide better (timeout) support to solutions polling the ResourceSet to
become `Ready`.

#### Reconciling ResourceSet

The flux-operator marks a ResourceSet as _reconciling_ when it starts
the reconciliation of the same. The Condition added to the ResourceSet's
`.status.conditions` has the following attributes:

- `type: Reconciling`
- `status: "True"`
- `reason: Progressing` | `reason: ProgressingWithRetry`

The Condition `message` is updated during the course of the reconciliation to
report the action being performed at any particular moment such as
building manifests, detecting drift, etc.

The `Ready` Condition's `status` is also marked as `Unknown`.

#### Ready ResourceSet

The flux-operator marks a ResourceSet as _ready_ when the resources were
built and applied on the cluster and all health checks are observed to be passing.

When the ResourceSet is "ready", the flux-operator sets a Condition with the
following attributes in the ResourceSet’s `.status.conditions`:

- `type: Ready`
- `status: "True"`
- `reason: ReconciliationSucceeded`

#### Failed ResourceSet

The flux-operator may get stuck trying to reconcile and apply a
ResourceSet without completing. This can occur due to some of the following factors:

- The dependencies are not ready.
- The templating of the resources fails.
- The resources are invalid and cannot be applied.
- Garbage collection fails.
- Running health checks fails.

When this happens, the flux-operator sets the `Ready` Condition status to False
and adds a Condition with the following attributes to the ResourceSet’s
`.status.conditions`:

- `type: Ready`
- `status: "False"`
- `reason: DependencyNotReady | BuildFailed | ReconciliationFailed | HealthCheckFailed`

The `message` field of the Condition will contain more information about why
the reconciliation failed.

While the ResourceSet has one or more of these Conditions, the flux-operator
will continue to attempt a reconciliation with an
exponential backoff, until it succeeds and the ResourceSet is marked as [ready](#ready-fluxinstance).

### Inventory status

In order to perform operations such as drift detection, garbage collection, upgrades, etc.,
the flux-operator needs to keep track of all Kubernetes objects that are
reconciled as part of a ResourceSet. To do this, it maintains an inventory
containing the list of Kubernetes resource object references that have been
successfully applied and records it in `.status.inventory`. The inventory
records are in the format `Id: <namespace>_<name>_<group>_<kind>, V: <version>`.

Example:

```text
Status:
  Inventory:
    Entries:
      Id: default_podinfo__ServiceAccount
      V:  v1
      Id: default_podinfo__Service
      V:  v1
      Id: default_podinfo_apps_Deployment
      V:  v1
```

### DIFF
--- a/api/v1/resourceset_types.go
+++ b/api/v1/resourceset_types.go
@@ -33,6 +33,14 @@ type ResourceSetSpec struct {
 	// +optional
 	Resources []*apiextensionsv1.JSON `json:"resources,omitempty"`
 
+	// ResourcesTemplate is a Go template that generates the list of
+	// Kubernetes resources to reconcile. The template is rendered
+	// as multi-document YAML, the resources should be separated by '---'.
+	// When both Resources and ResourcesTemplate are set, the resulting
+	// objects are merged and deduplicated, with the ones from Resources taking precedence.
+	// +optional
+	ResourcesTemplate string `json:"resourcesTemplate,omitempty"`
+
 	// DependsOn specifies the list of Kubernetes resources that must
 	// exist on the cluster before the reconciliation process starts.
 	// +optional
--- a/docs/api/v1/resourceset.md
+++ b/docs/api/v1/resourceset.md
@@ -152,7 +152,6 @@ spec:
 An input value is a key-value pair of strings and structs, where the key is the input name
 which can be referenced in the resource templates using the `<< inputs.name >>` syntax.
 
-
 ### Resources configuration
 
 The `.spec.resources` field is optional and specifies the list of Kubernetes resource
@@ -328,6 +327,67 @@ spec:
 ```
 
 In the above example, the `ServiceAccount` resource is generated only for the `team1` tenant.
+
+### Resources template
+
+The `.spec.resourcesTemplate` field is optional and offers an alternative to the `.spec.resources`.
+The `.spec.resourcesTemplate` is a single string that contains the multi-document YAML of the resources
+definitions. This field can be used for complex templating scenarios with the trade-off of reduced readability.
+
+Note that when both `.spec.resources` and `.spec.resourcesTemplate` are set, the resulting resources
+are the union of the two. If duplicate resources are defined in both fields, the resources from
+`.spec.resources` take precedence.
+
+Example of a template containing conditional and repeated blocks:
+
+```yaml
+spec:
+  inputs:
+    - bundle: addons
+      decryption: false
+      components:
+        - ingress-nginx
+        - cert-manager
+    - bundle: apps
+      decryption: true
+      components:
+        - frontend
+        - backend
+  resourcesTemplate: |
+    ---
+    apiVersion: source.toolkit.fluxcd.io/v1beta2
+    kind: OCIRepository
+    metadata:
+      name: << inputs.bundle >>
+      namespace: flux-system
+    spec:
+      interval: 10m
+      url: oci://registry.example.com/<< inputs.bundle >>
+    <<- range $component := inputs.components >>
+    ---
+    apiVersion: kustomize.toolkit.fluxcd.io/v1
+    kind: Kustomization
+    metadata:
+      name: << $component >>
+      namespace: flux-system
+    spec:
+      interval: 1h
+      prune: true
+      <<- if inputs.decryption >>
+      decryption:
+        provider: sops
+        secretRef:
+          name: << inputs.bundle >>-sops
+      <<- end >>
+      sourceRef:
+        kind: OCIRepository
+        name: << inputs.bundle >>
+      path: ./<< $component >>
+    <<- end >>
+```
+
+The above example generates two `OCIRepository` resources (one for each bundle) and four
+`Kustomization` resources (one for each component in each bundle).
 
 ### Common metadata
 
--- a/internal/builder/resourceset.go
+++ b/internal/builder/resourceset.go
@@ -22,7 +22,7 @@ import (
 
 // BuildResourceSet builds a list of Kubernetes resources
 // from a list of JSON templates using the provided inputs.
-func BuildResourceSet(templates []*apix.JSON, jsonInputs []fluxcdv1.ResourceSetInput) ([]*unstructured.Unstructured, error) {
+func BuildResourceSet(yamlTemplate string, templates []*apix.JSON, jsonInputs []fluxcdv1.ResourceSetInput) ([]*unstructured.Unstructured, error) {
 	inputs := make([]map[string]any, 0, len(jsonInputs))
 	for i, ji := range jsonInputs {
 		inp := make(map[string]any, len(ji))
@@ -37,6 +37,8 @@ func BuildResourceSet(templates []*apix.JSON, jsonInputs []fluxcdv1.ResourceSetI
 	}
 
 	var objects []*unstructured.Unstructured
+
+	// build resources from JSON templates
 	for i, tmpl := range templates {
 		if len(inputs) == 0 {
 			object, err := BuildResource(tmpl, nil)
@@ -60,18 +62,38 @@ func BuildResourceSet(templates []*apix.JSON, jsonInputs []fluxcdv1.ResourceSetI
 			}
 
 			// deduplicate objects
-			found := false
-			for _, obj := range objects {
-				if obj.GetAPIVersion() == object.GetAPIVersion() &&
-					obj.GetKind() == object.GetKind() &&
-					obj.GetNamespace() == object.GetNamespace() &&
-					obj.GetName() == object.GetName() {
-					found = true
-					break
-				}
+			if !containsObject(objects, object) {
+				objects = append(objects, object)
+			}
+		}
+	}
+
+	// build resources from multi-doc YAML template
+	if yamlTemplate != "" {
+		var objectsFromTemplate []*unstructured.Unstructured
+		if len(inputs) == 0 {
+			objs, err := BuildResourcesFromYAML(yamlTemplate, nil)
+			if err != nil {
+				return nil, fmt.Errorf("failed to build resource: %w", err)
 			}
 
-			if !found {
+			objectsFromTemplate = append(objectsFromTemplate, objs...)
+		}
+		for _, input := range inputs {
+			objs, err := BuildResourcesFromYAML(yamlTemplate, input)
+			if err != nil {
+				return nil, fmt.Errorf("failed to build resources: %w", err)
+			}
+			objectsFromTemplate = append(objectsFromTemplate, objs...)
+		}
+
+		for _, object := range objectsFromTemplate {
+			// exclude object based on annotations
+			if val := object.GetAnnotations()[fluxcdv1.ReconcileAnnotation]; val == fluxcdv1.DisabledValue {
+				continue
+			}
+			// deduplicate objects
+			if !containsObject(objects, object) {
 				objects = append(objects, object)
 			}
 		}
@@ -83,19 +105,14 @@ func BuildResourceSet(templates []*apix.JSON, jsonInputs []fluxcdv1.ResourceSetI
 // BuildResource builds a Kubernetes resource from a JSON template using the provided inputs.
 // Template functions are provided by the slim-sprig library https://go-task.github.io/slim-sprig/.
 // In addition, the slugify function is available to generate slugs from strings using https://github.com/gosimple/slug/.
+// And for readability, a toYaml function is available to encode an input value into a YAML string.
 func BuildResource(tmpl *apix.JSON, inputs map[string]any) (*unstructured.Unstructured, error) {
-	ymlTemplate, err := yaml.JSONToYAML(tmpl.Raw)
+	yamlTemplate, err := yaml.JSONToYAML(tmpl.Raw)
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert template to YAML: %w", err)
 	}
 
-	tp, err := template.New("res").
-		Delims("<<", ">>").
-		Funcs(sprig.HermeticTxtFuncMap()).
-		Funcs(template.FuncMap{"slugify": slug.Make}).
-		Funcs(template.FuncMap{"inputs": func() any { return inputs }}).
-		Funcs(template.FuncMap{"toYaml": toYaml, "mustToYaml": mustToYaml}).
-		Parse(string(ymlTemplate))
+	tp, err := newTemplate(string(yamlTemplate), inputs)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse template: %w", err)
 	}
@@ -112,6 +129,56 @@ func BuildResource(tmpl *apix.JSON, inputs map[string]any) (*unstructured.Unstru
 	}
 
 	return object, nil
+}
+
+// BuildResourcesFromYAML builds a list of Kubernetes resources from a multi-doc YAML template
+// using the same templating functions as BuildResource.
+func BuildResourcesFromYAML(yamlTemplate string, inputs map[string]any) ([]*unstructured.Unstructured, error) {
+	tp, err := newTemplate(yamlTemplate, inputs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse multi-doc YAML template: %w", err)
+	}
+
+	b := &strings.Builder{}
+	err = tp.Execute(b, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute multi-doc YAML template: %w", err)
+	}
+
+	objects, err := ssautil.ReadObjects(bytes.NewReader([]byte(b.String())))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read objects from multi-doc YAML: %w", err)
+	}
+
+	return objects, nil
+}
+
+func newTemplate(yamlTemplate string, inputs map[string]any) (*template.Template, error) {
+	tp, err := template.New("res").
+		Delims("<<", ">>").
+		Funcs(sprig.HermeticTxtFuncMap()).
+		Funcs(template.FuncMap{"slugify": slug.Make}).
+		Funcs(template.FuncMap{"inputs": func() any { return inputs }}).
+		Funcs(template.FuncMap{"toYaml": toYaml, "mustToYaml": mustToYaml}).
+		Parse(yamlTemplate)
+	if err != nil {
+		return nil, err
+	}
+	return tp, nil
+}
+
+func containsObject(objects []*unstructured.Unstructured, object *unstructured.Unstructured) bool {
+	found := false
+	for _, obj := range objects {
+		if obj.GetAPIVersion() == object.GetAPIVersion() &&
+			obj.GetKind() == object.GetKind() &&
+			obj.GetNamespace() == object.GetNamespace() &&
+			obj.GetName() == object.GetName() {
+			found = true
+			break
+		}
+	}
+	return found
 }
 
 // init initializes the slugify Go template function with the default settings.
--- a/internal/controller/resourceset_controller.go
+++ b/internal/controller/resourceset_controller.go
@@ -141,7 +141,7 @@ func (r *ResourceSetReconciler) reconcile(ctx context.Context,
 	}
 
 	// Build the resources.
-	buildResult, err := builder.BuildResourceSet(obj.Spec.Resources, obj.GetInputs())
+	buildResult, err := builder.BuildResourceSet(obj.Spec.ResourcesTemplate, obj.Spec.Resources, obj.GetInputs())
 	if err != nil {
 		msg := fmt.Sprintf("build failed: %s", err.Error())
 		conditions.MarkFalse(obj,
--- a/config/crd/bases/fluxcd.controlplane.io_resourcesets.yaml
+++ b/config/crd/bases/fluxcd.controlplane.io_resourcesets.yaml
@@ -119,6 +119,14 @@ spec:
                 items:
                   x-kubernetes-preserve-unknown-fields: true
                 type: array
+              resourcesTemplate:
+                description: |-
+                  ResourcesTemplate is a Go template that generates the list of
+                  Kubernetes resources to reconcile. The template is rendered
+                  as multi-document YAML, the resources should be separated by '---'.
+                  When both Resources and ResourcesTemplate are set, the resulting
+                  objects are merged and deduplicated, with the ones from Resources taking precedence.
+                type: string
               serviceAccountName:
                 description: |-
                   The name of the Kubernetes service account to impersonate
--- a/internal/builder/resourceset_test.go
+++ b/internal/builder/resourceset_test.go
@@ -48,6 +48,11 @@ func TestBuildResourceSet(t *testing.T) {
 			srcFile:    filepath.Join(testdataRoot, "nestedinputs.yaml"),
 			goldenFile: filepath.Join(testdataRoot, "nestedinputs.golden.yaml"),
 		},
+		{
+			name:       "multi-doc-template",
+			srcFile:    filepath.Join(testdataRoot, "multi-doc-template.yaml"),
+			goldenFile: filepath.Join(testdataRoot, "multi-doc-template.golden.yaml"),
+		},
 	}
 
 	for _, tt := range tests {
@@ -61,7 +66,7 @@ func TestBuildResourceSet(t *testing.T) {
 			err = yaml.Unmarshal(data, &rg)
 			g.Expect(err).ToNot(HaveOccurred())
 
-			objects, err := BuildResourceSet(rg.Spec.Resources, rg.GetInputs())
+			objects, err := BuildResourceSet(rg.Spec.ResourcesTemplate, rg.Spec.Resources, rg.GetInputs())
 			g.Expect(err).ToNot(HaveOccurred())
 
 			manifests, err := ssautil.ObjectsToYAML(objects)
@@ -92,7 +97,7 @@ func TestBuildResourceSet_Empty(t *testing.T) {
 	err = yaml.Unmarshal(data, &rg)
 	g.Expect(err).ToNot(HaveOccurred())
 
-	objects, err := BuildResourceSet(rg.Spec.Resources, rg.GetInputs())
+	objects, err := BuildResourceSet(rg.Spec.ResourcesTemplate, rg.Spec.Resources, rg.GetInputs())
 	g.Expect(err).ToNot(HaveOccurred())
 	g.Expect(objects).To(BeEmpty())
 }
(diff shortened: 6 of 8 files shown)


internal/builder/resourceset.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute template: %w", err)
+	}
+
+	object, err := ssautil.ReadObject(bytes.NewReader([]byte(b.String())))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read object: %w", err)
+	}
+
+	return object, nil
```
matheuscscp, line 113: Right before returning here you could default the namespace of object to the namespace of the `ResourceSet` if not specified, and if the resource is namespaced (can use the discovery client like mentioned in the comment for the dependencies).

api/v1/resourceset_types.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// The name of the Kubernetes service account to impersonate
+	// when reconciling the generated resources.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// Wait instructs the controller to check the health
+	// of all the reconciled resources.
+	// +optional
+	Wait bool `json:"wait,omitempty"`
```
matheuscscp, line 49: The first thing that comes across my mind reading this field is "custom healthchecks".

matheuscscp (commented):
First batch of comments :hugs:

api/v1/resourceset_types.go
matheuscscp: Also there should be a timeout field.

api/v1/resourceset_types.go
```diff
@@ -0,0 +1,191 @@
...
+	// +optional
+	Ready bool `json:"ready,omitempty"`
+
+	// ReadyExpr checks if the resource satisfies the given CEL expression.
+	// +optional
+	ReadyExpr string `json:"readyExpr,omitempty"`
+}
+
+// ResourceSetInput defines the key-value pairs of the resource group input.
+type ResourceSetInput map[string]string
```
matheuscscp, line 80: Why not make it `map[string]any` for allowing nested fields? I think it improves a lot your ability to organize your inputs.

internal/builder/resourceset.go
stefanprodan: I'm not for calling the discovery client here, it will fail if the RS contains both a CRD and a CR. I'm also not for defaulting the namespace in a RS, for the same reason a Flux Kustomization does not set any default namespace anywhere, it's opaque and can result in major issues, better to error out and have users set the right namespace.

api/v1/resourceset_types.go
stefanprodan: The timeout is set via annotations as explained in the docs, this is the convention used by all Flux Operator APIs.

internal/controller/resourceset_controller.go
```diff
@@ -0,0 +1,643 @@
...
+					ext.Sets(),
+					ext.Encoders(),
+				}
+
+				env, err := cel.NewEnv(envOptions...)
+				if err != nil {
+					return fmt.Errorf("failed to create CEL env: %w", err)
+				}
+
+				expr, issues := env.Parse(dep.ReadyExpr)
```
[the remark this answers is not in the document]
stefanprodan: Eval runs the check, and if I call it before passing the object it will always fail because the expression contains object paths.

internal/builder/resourceset.go
```diff
@@ -0,0 +1,114 @@
...
+
+			// exclude object based on annotations
+			if val := object.GetAnnotations()[fluxcdv1.ReconcileAnnotation]; val == fluxcdv1.DisabledValue {
+				continue
+			}
+
+			// deduplicate objects
+			found := false
+			for _, obj := range objects {
+				if obj.GetAPIVersion() == object.GetAPIVersion() &&
```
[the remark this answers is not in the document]
stefanprodan: If this happens, then it's up the Kubernetes API to override the previous version using whatever conversion webhook is available.

docs/api/v1/resourceset.md
```diff
@@ -0,0 +1,637 @@
...
+          name: << inputs.app >>-sops
+      <<- end >>
+      sourceRef:
+        kind: OCIRepository
+        name: << inputs.app >>
+      path: ./
+```
+
+When both `.spec.resources` and `.spec.resourcesTemplate` are set, the resulting resources are the union of the two.
+If duplicate resources are defined in both fields, the resources from `.spec.resources` take precedence.
```
matheuscscp, line 381: Let's document these two lines in the struct as well.

docs/api/v1/resourceset.md
```diff
@@ -0,0 +1,637 @@
...
+          fluxcd.controlplane.io/reconcile: << if eq inputs.tenant "team1" >>enabled<< else >>disabled<< end >>
+```
+
+In the above example, the `ServiceAccount` resource is generated only for the `team1` tenant.
+
+### Resources template
+
+The `.spec.resourcesTemplate` field is optional and offers an alternative to the `.spec.resources`.
+The `.spec.resourcesTemplate` is a single string that contains the multi-document YAML of the resources
+definitions. This field can be used for complex templating scenarios with the trade-off of reduced readability.
```
matheuscscp, line 335:
```suggestion
definitions. This field can be used for complex templating scenarios with the trade-off of reduced readability,
e.g. when conditional or repeated blocks are needed.
```

internal/builder/resourceset.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert template to YAML: %w", err)
+	}
+
+	tp, err := template.New("res").
+		Delims("<<", ">>").
+		Funcs(sprig.HermeticTxtFuncMap()).
+		Funcs(template.FuncMap{"slugify": slug.Make}).
+		Funcs(template.FuncMap{"inputs": func() any { return inputs }}).
+		Funcs(template.FuncMap{"toYaml": toYaml, "mustToYaml": mustToYaml}).
```
matheuscscp, line 119: In order to keep the templating engine consistent, let's extract this part to a function that takes `inputs` as argument and call it in both `BuildResource` and `BuildResourcesFromYAML`.

internal/controller/resourceset_controller.go
```diff
@@ -0,0 +1,643 @@
...
+				Name:          "kubectl",
+				OperationType: metav1.ManagedFieldsOperationUpdate,
+			},
+		},
+	}
+
+	resultSet := ssa.NewChangeSet()
+
+	// Apply the resources to the cluster.
+	changeSet, err := resourceManager.ApplyAllStaged(ctx, objects, applyOpts)
```
matheuscscp, line 360: I see that kustomize-controller instead of using this function breaks down the manifests in the stages:

* CRDs + Namespaces
* Classes
* Everything else

While `ApplyAllStaged` merges classes with everything else.

What's the use for kustomize-controller to do that? Do we need it here too?

I see that `FluxInstance` also uses `ApplyAllStaged` but it doesn't deploy any classes so it's trivially fine in its case.

matheuscscp (commented):
LGTM, after fixing the remaining comments I think we can merge :+1:

internal/builder/resourceset.go
```diff
@@ -0,0 +1,205 @@
...
+		if len(inputs) == 0 {
+			objs, err := BuildResourcesFromYAML(yamlTemplate, nil)
+			if err != nil {
+				return nil, fmt.Errorf("failed to build resource: %w", err)
+			}
+
+			objectsFromTemplate = append(objectsFromTemplate, objs...)
+		}
+		for _, input := range inputs {
+			objs, err := BuildResourcesFromYAML(yamlTemplate, input)
```
matheuscscp, line 83: Also I just realized that we can build and parse the template just once in this case and execute it for each `input`.

matheuscscp: Actually this also true for `.spec.resources`.

stefanprodan: Hmm not really, inputs can differ in structure.

matheuscscp: Oh yeah, totally, my bad!

internal/controller/resourceset_controller.go
```diff
@@ -0,0 +1,643 @@
...
+	"github.com/controlplaneio-fluxcd/flux-operator/internal/inventory"
+)
+
+// ResourceSetReconciler reconciles a ResourceSet object
+type ResourceSetReconciler struct {
+	client.Client
+	kuberecorder.EventRecorder
+
+	APIReader             client.Reader
+	Scheme                *runtime.Scheme
```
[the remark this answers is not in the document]
stefanprodan: These will be used once we'll query the cluster for input providers.

matheuscscp (approved):
:rocket: :rocket: :rocket:

Issue #121, opened 2024-12-13
Allow a shorter reconciliation interval for the artifact


matheuscscp:
Today we have the annotation `fluxcd.controlplane.io/reconcileEvery` for configuring the interval with which flux-operator will reconcile a `FluxInstance`, which defaults to `1h`. The server-side apply dry-run can be resource intensive, so this interval should indeed be kept around this amount of time.

However, for pulling the OCI artifact containing the latest images (i.e. with the latest CVE patches), users may want a shorter interval, e.g. `2m`.

# Proposal

Create a new reconciler that runs on a separate interval controlled by the annotation `fluxcd.controlplane.io/reconcileArtifactEvery` defaulting to `10m`. This reconciler will compare the digest of the artifact with the one stored in the `FluxInstance` status and annotate the object to request a reconciliation in case of mismatch.

stefanprodan:
A first step towards this, would be to introduce a new optional field in status called `lastArtifactRevision` with the values set from: https://github.com/controlplaneio-fluxcd/flux-operator/blob/3d483b72342f0d7d667e87608fe15160864dcd75/internal/controller/fluxinstance_controller.go#L234

This status fields should only be set if `.spec.distribution.artifact` exists, when the spec changes, and the artifact field is removed, so should the status field. This status field will act as a enabler of the artifact reconciler, if status does not contain `lastArtifactRevision`, then the reconciler should exit early.

PR #127, opened 2024-12-17
Add FluxInstance artifact reconciler

matheuscscp:
Closes: #121


matheuscscp:
Will investigate the e2e timeout tomorrow.

config/olm/bundle/manifests/flux-operator.clusterserviceversion.yaml
```diff
@@ -15,6 +15,7 @@ metadata:
             "namespace": "flux-system",
             "annotations": {
               "fluxcd.controlplane.io/reconcileEvery": "1h",
+              "fluxcd.controlplane.io/reconcileArtifactEvery": "10m",
```
stefanprodan, line 18: We don't set the artifact in the OLM example so this annotation can be removed.

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	// Skip reconciliation if the object has the reconcile annotation set to 'disabled'.
+	if obj.IsDisabled() {
+		return ctrl.Result{}, nil
+	}
+
+	// Skip reconciliation if the object does not have a distribution artifact specified.
+	if obj.Spec.Distribution.Artifact == "" {
+		return ctrl.Result{}, nil
+	}
+
```
stefanprodan, line 52: With this we avoid a race condition when the FluxInstance is first created, we don't want this reconciler to run then.
```suggestion
if obj.Status == nil || obj.Status.LastArtifactRevision == "" {
return requeueArtifactAfter(obj *fluxcdv1.FluxInstance), nil
}
```

internal/controller/fluxinstance_artifact_manager.go
```diff
@@ -0,0 +1,35 @@
...
+
+// FluxInstanceArtifactReconcilerOptions contains options for the reconciler.
+type FluxInstanceArtifactReconcilerOptions struct {
+	RateLimiter workqueue.TypedRateLimiter[reconcile.Request]
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *FluxInstanceArtifactReconciler) SetupWithManager(mgr ctrl.Manager, opts FluxInstanceArtifactReconcilerOptions) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&fluxcdv1.FluxInstance{},
```
stefanprodan, line 25: Wouldn't this prevent the main reconciler from ever running due to the FluxInstance being removed from the queue?

matheuscscp: Maybe this is the reason for the e2e timeout?

config/olm/bundle/manifests/flux-operator.clusterserviceversion.yaml
matheuscscp: fixed, thanks!

internal/controller/fluxinstance_artifact_controller.go
matheuscscp: fixed, thanks! (btw `obj.Status` is not a pointer)

internal/controller/fluxinstance_artifact_manager.go
matheuscscp: I'm trying to prepend `Named("flux_instance_artifact").` before the `Watches(...)` to see what happens. In any case, the description of `For(...)` says it is equivalent to the `Watches(...)` I'm using now :cry: :

```
For defines the type of Object being *reconciled*, and configures the ControllerManagedBy to respond to create / delete / update events by *reconciling the object*.

This is the equivalent of calling Watches(source.Kind(cache, &Type{}, &handler.EnqueueRequestForObject{})).
```

matheuscscp: lol, but it works! the e2e passed locally

matheuscscp: Looks like `For` creates the name `tolower(Kind)`, so `fluxinstance`. I'm gonna change the name to `fluxinstance_artifact`.

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	// Skip reconciliation if the object has the reconcile annotation set to 'disabled'.
+	if obj.IsDisabled() {
+		return ctrl.Result{}, nil
+	}
+
+	// Skip reconciliation if the object does not have a last artifact revision to avoid race condition.
+	if obj.Status.LastArtifactRevision == "" {
+		return ctrl.Result{}, nil
```
stefanprodan, line 50: This need `requeueArtifactAfter`, either wise you'll be disabling the reconciliation forever since the watcher skips status update.

stefanprodan: We should also exit and requeue when IsReady is false

matheuscscp: fixed both, thanks!

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Fetch the latest digest of the distribution manifests.
+	artifactURL := obj.Spec.Distribution.Artifact
+	artifactDigest, err := builder.HeadArtifact(ctx, artifactURL)
+	if err != nil {
+		msg := fmt.Sprintf("fetch failed: %s", err.Error())
+		r.Event(obj, corev1.EventTypeWarning, meta.ArtifactFailedReason, msg)
+		return ctrl.Result{}, err
+	}
+	log.Info("fetched latest manifests", "url", artifactURL, "digest", artifactDigest)
```
stefanprodan, line 78: This will fill the log with duplicate messages every 10 minutes.

matheuscscp: changed `log.Info` to `log.V(1).Info`, thanks!

internal/builder/head.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+)
+
+// HeadArtifact looks up an artifact from an OCI repository and returns the digest of the artifact.
+func HeadArtifact(ctx context.Context, ociURL string) (string, error) {
+	desc, err := crane.Head(strings.TrimPrefix(ociURL, "oci://"), crane.WithContext(ctx))
```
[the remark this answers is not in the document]
matheuscscp: fixed, thanks!

docs/api/v1/fluxinstance.md
```diff
@@ -23,6 +23,7 @@ metadata:
   annotations:
     fluxcd.controlplane.io/reconcile: "enabled"
     fluxcd.controlplane.io/reconcileEvery: "1h"
+    fluxcd.controlplane.io/reconcileArtifactEvery: "10m"
```
matheuscscp, line 26: We don't set the artifact in this example, so need to remove the annotation here too.

matheuscscp: fixed

internal/controller/fluxinstance_artifact_manager.go
```diff
@@ -0,0 +1,38 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *FluxInstanceArtifactReconciler) SetupWithManager(mgr ctrl.Manager, opts FluxInstanceArtifactReconcilerOptions) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("fluxinstance_artifact").
+		Watches(&fluxcdv1.FluxInstance{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				predicate.Or(
+					predicate.GenerationChangedPredicate{},
```
stefanprodan, line 31: We can remove the generation predicate and only react to annotations changes, that's where disabled and the interval can be set.

matheuscscp: fixed, thanks

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #131, opened 2024-12-21
Enable alerting for `FluxInstance` events

stefanprodan:
This PR integrates Flux Operator with notification-controller. Users can define Flux Alerts and get notified when the operator emits events during the FluxInstance reconciliation.

## Setup alerting

Define an Alert for FluxInstance events:

```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta3
kind: Provider
metadata:
  name: slack-bot
  namespace: flux-system
spec:
  type: slack
  channel: general
  address: https://slack.com/api/chat.postMessage
  secretRef:
    name: slack-bot-token
---
apiVersion: notification.toolkit.fluxcd.io/v1beta3
kind: Alert
metadata:
  name: slack-alert
  namespace: flux-system
spec:
  providerRef:
    name: slack-bot
  eventSeverity: info
  eventSources:
    - kind: FluxInstance
      name: flux
```

## Notifications examples

The operator will forward events to notification-controller after the install operation succeed, after a successful update (version bump, config change, CVE patch, etc) or when the reconciliation fails. 

Screenshot 2024-12-21 at 13 06 53

Screenshot 2024-12-21 at 13 09 14

Screenshot 2024-12-21 at 13 24 17

### DIFF
--- a/internal/controller/fluxinstance_controller.go
+++ b/internal/controller/fluxinstance_controller.go
@@ -307,6 +307,10 @@ func (r *FluxInstanceReconciler) build(ctx context.Context,
 		options.Patches += builder.GetMultitenantProfile(obj.GetCluster().TenantDefaultServiceAccount)
 	}
 
+	if builder.ContainElementString(options.Components, options.NotificationController) {
+		options.Patches += builder.GetNotificationPatch(options.Namespace)
+	}
+
 	if obj.Spec.Sharding != nil {
 		options.ShardingKey = obj.Spec.Sharding.Key
 		options.Shards = obj.Spec.Sharding.Shards
--- a/internal/controller/fluxinstance_controller.go
+++ b/internal/controller/fluxinstance_controller.go
@@ -154,7 +154,7 @@ func (r *FluxInstanceReconciler) reconcile(ctx context.Context,
 			meta.ReadyCondition,
 			meta.ArtifactFailedReason,
 			"%s", msg)
-		r.EventRecorder.Event(obj, corev1.EventTypeWarning, meta.ArtifactFailedReason, msg)
+		r.notify(ctx, obj, meta.ArtifactFailedReason, corev1.EventTypeWarning, msg)
 		return ctrl.Result{}, err
 	}
 
@@ -528,7 +528,9 @@ func (r *FluxInstanceReconciler) apply(ctx context.Context,
 			action = "installed"
 		}
 
-		msg := fmt.Sprintf("Flux %s revision %s\n%s", action, buildResult.Revision, applyLog)
+		ver := strings.Split(buildResult.Revision, "@")[0]
+
+		msg := fmt.Sprintf("Flux %s %s\n%s", ver, action, applyLog)
 		r.notify(ctx, obj, meta.ReconciliationSucceededReason, corev1.EventTypeNormal, msg)
 	}
 
@@ -641,7 +643,7 @@ func (r *FluxInstanceReconciler) recordMetrics(obj *fluxcdv1.FluxInstance) error
 
 func (r *FluxInstanceReconciler) notify(ctx context.Context, obj *fluxcdv1.FluxInstance, reason, eventType, msg string) {
 	notificationAddress := ""
-	if builder.ContainElementString(obj.GetComponents(), builder.MakeDefaultOptions().NotificationController) {
+	if os.Getenv("NOTIFICATIONS_DISABLED") == "" && builder.ContainElementString(obj.GetComponents(), builder.MakeDefaultOptions().NotificationController) {
 		notificationAddress = fmt.Sprintf("http://%s.%s.svc.%s/",
 			builder.MakeDefaultOptions().NotificationController,
 			obj.GetNamespace(),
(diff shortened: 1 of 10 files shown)


internal/controller/fluxinstance_controller.go
```diff
@@ -516,6 +525,20 @@ func (r *FluxInstanceReconciler) apply(ctx context.Context,
...
 
+	// Send event to notification-controller only if the server-side apply resulted in changes.
+	applyLog := strings.TrimSuffix(changeSetLog.String(), "\n")
+	if applyLog != "" {
+		action := "updated"
+		if len(oldInventory.Entries) == 0 {
+			action = "installed"
+		}
+
+		msg := fmt.Sprintf("Flux %s revision %s\n%s", action, buildResult.Revision, applyLog)
```
matheuscscp, line 536:
```suggestion
		msg := fmt.Sprintf("Flux revision %s %s\n%s", buildResult.Revision, action, applyLog)
```

I think `Flux revision X installed` reads better than `Flux installed revision X` since Flux is the one being installed, and not the one installing anything. Same for `updated`.

internal/controller/fluxinstance_controller.go
```diff
@@ -622,3 +645,28 @@ func (r *FluxInstanceReconciler) recordMetrics(obj *fluxcdv1.FluxInstance) error
...
+	}
+
+	eventRecorder, err := events.NewRecorderForScheme(
+		r.Scheme,
+		r.EventRecorder,
+		ctrl.Log,
+		notificationAddress,
+		r.StatusManager)
+	if err != nil {
+		return err
```
matheuscscp, line 666: Instead of returning this error you could simply log it here since you already have `msg` to indicate what the event is for and don't need to be specific about it on every call to `notify()` like you did.

internal/controller/fluxinstance_controller.go
stefanprodan: How about using the verb first, like `Installed Flux revision X`, I want to place the revision last as it contains a SHA256 which is so long.

matheuscscp: Sounds good 👍

stefanprodan: I went with "Flux <version> [installed|updated]" and moved the revision to the event metadata.

internal/controller/fluxinstance_controller.go
```diff
@@ -622,3 +640,34 @@ func (r *FluxInstanceReconciler) recordMetrics(obj *fluxcdv1.FluxInstance) error
 	reporter.RecordMetrics(unstructured.Unstructured{Object: rawMap})
 	return nil
 }
+
+func (r *FluxInstanceReconciler) notify(ctx context.Context, obj *fluxcdv1.FluxInstance, reason, eventType, msg string) {
+	notificationAddress := ""
+	if os.Getenv("NOTIFICATIONS_DISABLED") == "" && builder.ContainElementString(obj.GetComponents(), builder.MakeDefaultOptions().NotificationController) {
```
matheuscscp, line 646: I'd prefer avoiding an env var here and introduce this as a boolean in the `FluxInstanceReconciler` struct instead.

internal/controller/fluxinstance_controller_test.go
```diff
@@ -219,6 +229,7 @@ func TestFluxInstanceReconciler_LifeCycle(t *testing.T) {
 
 func TestFluxInstanceReconciler_FetchFail(t *testing.T) {
 	g := NewWithT(t)
+	t.Setenv("NOTIFICATIONS_DISABLED", "yes")
```
matheuscscp, line 232: Since we are disabling notifications in most places in the tests I'd prefer disabling them by default inside the `getFluxInstanceReconciler()` function and enabling only in the tests that need it (if any).

internal/controller/fluxinstance_controller.go
```diff
@@ -622,3 +640,34 @@ func (r *FluxInstanceReconciler) recordMetrics(obj *fluxcdv1.FluxInstance) error
...
+		r.StatusManager)
+	if err != nil {
+		log.Error(err, "failed to create event recorder")
+		return
+	}
+
+	annotations := map[string]string{}
+	if obj.Status.LastAttemptedRevision != "" {
+		annotations[fluxcdv1.RevisionAnnotation] = obj.Status.LastAttemptedRevision
+	}
```
matheuscscp, line 670: Since in this case we are losing an event it'd be really nice to enhance the context of this log with the context of the lost event.

```suggestion
	annotations := map[string]string{}
	if obj.Status.LastAttemptedRevision != "" {
		annotations[fluxcdv1.RevisionAnnotation] = obj.Status.LastAttemptedRevision
	}

	eventRecorder, err := events.NewRecorderForScheme(
		r.Scheme,
		r.EventRecorder,
		log,
		notificationAddress,
		r.StatusManager)
	if err != nil {
		log.Error(err, "failed to create event recorder for event", "event", map[string]any{
			"instance":    map[string]any{"name": obj.Name, "namespace": obj.Namespace},
			"type":        eventType,
			"reason":      reason,
			"message":     msg,
			"annotations": annotations})
		return
	}
```

internal/controller/fluxinstance_controller.go
stefanprodan: Notifications should work for ResourceGroups and any other new APIs we'll add in the future, so I want to avoid hard coupling it to the FluxInstance. The env var is more of a kill switch for when you don't want the operator to send any requests to the notification-controller HTTP endpoint.

internal/controller/fluxinstance_controller.go
stefanprodan: The messages are already logged by the operator at each step. This would just duplicate tones of text in the controller logs.

internal/controller/fluxinstance_controller_test.go
stefanprodan: Done! Thanks

matheuscscp (approved):
LGTM :ok_hand:

Issue #140, opened 2025-01-12
`ResourceSet` dynamic inputs for preview environments automation


stefanprodan:
With the [introduction of the ResourceSet API](#98), the Flux Operator can generate a set of Kubernetes resources based on a **static input list** defined by users in the `ResourceSet` spec. There are use-cases when the inputs are not known ahead of time and can't be set in a static manner. One such use-case is deploying app code and/or config changes made in a GitHub Pull Request or GitLab Merge Request. To accommodate this use-case, the Flux Operator should allow users to **define inputs in a dynamic manner** using a CRD named `ResourceSetInputProvider`. Based on this CRD the operator would call an external system (e.g. GitHub, GitLab, AzureDevOps APIs) and transform the response for use as `ResouceSet` inputs. 

## Example

We'll use GitHub and the `podinfo` demo app to demonstrate how the new API should work, but the `ResourceSetInputProvider` could support GitLab, Azure DevOps, Gitea and any other Git service.

### Preview environments workflow

- A developer makes code changes to the podinfo source code and its Helm chart and opens a PR in GitHub
- The CI builds the new app container image and pushes a tag to GHCR with the Git commit SHA
- Another developer reviews the changes and adds a label to the PR `deploy/flux-preview`
- Flux Operator running in the preview cluster scans the repository and finds the new PR using the label filter
- Flux Operator installs a Helm release using the PR number and the commit SHA inputs to deploy the app and chart changes in the cluster
- The developers iterate over changes, with each push to the PR branch triggering a Helm release upgrade in the cluster
- Once the PR is approved and merged, the Flux Operator uninstalls the Helm release form the cluster

### Flux Operator Config

To enable the above workflow, we'll define a series of Flux Operator custom resources in the preview cluster.

First we'll create a dedicated namespace called `podinfo-preview` where all the app instances generated from PRs will be deployed. In that namespace, we'll create a Kubernetes Secret containing a GitHub token that grants read access to the podinfo repository and PRs.

In the `podinfo-preview` namespace, we'll create a `ResourceSetInputProvider` that tells Flux Operator to scan the repository for PRs labeled with `deploy/flux-preview`:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSetInputProvider
metadata:
  name: podinfo-pull-requests
  namespace: podinfo-preview
  annotations:
    fluxcd.controlplane.io/reconcileEvery: "10m"
spec:
  type: GitHubPullRequest
[...]

PR #143, opened 2025-01-13
Introduce `ResourceSetInputProvider` API

stefanprodan:
This PR introduces the `ResourceSetInputProvider` API with support for two providers: `GitHubPullRequest` and `GitLabMergeRequest`.

 For the motivation behind this new API see #140 

Closes: #140

---

**ResourceSetInputProvider** is a declarative API for generating a set of input values
for use within [ResourceSet](resourceset.md) definitions. The input values are fetched from external
services such as GitHub or GitLab, and can be used to parameterize the resources templates
defined in ResourceSets.

## Example

The following example shows a provider that fetches input values from
GitHub Pull Requests labeled with `deploy/flux-preview`:

```yaml
metadata:
  name: flux-appx-prs
  namespace: default
  annotations:
spec:
  type: GitHubPullRequest
  url: https://github.com/controlplaneio-fluxcd/flux-appx
  filter:
    labels:
      - "deploy/flux-preview"
  defaultValues:
    chart: "charts/flux-appx"
```

You can run this example by saving the manifest into `flux-appx-prs.yaml`.

1. Apply the ResourceSetInputProvider on the cluster:

   ```shell
   kubectl apply -f flux-appx-prs.yaml.yaml
   ```

2. Wait for the ResourceSetInputProvider to reconcile:

   ```shell
   kubectl wait rsip/flux-appx-prs --for=condition=ready --timeout=5m
   ```

3. Run `kubectl get -o yaml` to see the exported inputs generated in the ResourceSetInputProvider status:

   ```console
   $ kubectl get rsip/flux-appx-prs -o yaml | yq .status.exportedInputs
   - author: stefanprodan
     branch: kubernetes/helm-set-limits
     chart: charts/flux-appx
     id: "4"
     sha: bf5d6e01cf802734853f6f3417b237e3ad0ba35d
     title: 'kubernetes(helm): Add default resources limits'
   - author: stefanprodan
     branch: feat/ui-footer
     chart: charts/flux-appx
     id: "3"
     sha: 8492c0b5b2094fe720776c8ace1b9690ff258f53
     title: 'feat(ui): Add footer'
   - author: stefanprodan
     branch: feat/ui-color-scheme
     chart: charts/flux-appx
     id: "2"
     sha: 8166bdecd6b078b9e5dd14fa3b7b67a847f76893
     title: 'feat(ui): Default color scheme'
    ```

4. Run `kubectl delete` to remove the provider from the cluster:

    ```shell
    kubectl delete rsip/flux-appx-prs
    ```

## Writing a ResourceSetInputProvider spec

As with all other Kubernetes config, a ResourceSet needs `apiVersion`,
`kind`, `metadata.name` and `metadata.namespace` fields.
The name of a ResourceSet object must be a valid [DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
A ResourceSet also needs a [`.spec` section](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).

### Type

The `.spec.type` field is required and specifies the type of the provider.

The following types are supported:

- `GitHubPullRequest`: fetches input values from opened GitHub Pull Requests.
- `GitLabMergeRequest`: fetches input values from opened GitLab Merge Requests.

For both `GitHubPullRequest` and `GitLabMergeRequest` types, flux-operator will
export in `.status.exportedInputs` a set of input values for each
Pull Request or Merge Request that matches the [filter](#filter) criteria.

The [exported inputs](#exported-inputs-status) structure is:

- `id`: the ID number of the PR/MR (type string).
- `sha`: the commit SHA of the PR/MR (type string).
- `branch`: the branch name of the PR/MR (type string).
- `author`: the author username of the PR/MR (type string).
- `title`: the title of the PR/MR (type string).

### URL

The `.spec.url` field is required and specifies the HTTP/S URL of the provider.
For Git services, the URL should contain the GitHub repository or the GitLab project address.

### Filter

The `.spec.filter` field is optional and specifies the filter criteria for the input values.

The following filters are supported:

- `limit`: limit the number of input values fetched (default is 100).
- `labels`: filter GitHub Pull Requests or GitLab Merge Requests by labels.
- `includeBranch`: regular expression to include branches by name.
- `excludeBranch`: regular expression to exclude branches by name.

Example of a filter configuration for GitLab Merge Requests:

```yaml
spec:
  filter:
    limit: 10
    labels:
      - "deploy::flux-preview"
    includeBranch: "^feat/.*"
    excludeBranch: "^feat/not-this-one$"
```

### Default values

The `.spec.defaultValues` field is optional and specifies the default values for the exported inputs.
This field can be used to set values that are common to all the exported inputs.

Example:

```yaml
spec:
  defaultValues:
    env: "staging"
    tenants:
      - "tenant1"
      - "tenant2"
```

### Authentication configuration

The `.spec.secretRef` field is optional and specifies the Kubernetes Secret containing
the authentication credentials used for connecting to the external service.
Note that the secret must be created in the same namespace as the ResourceSetInputProvider.

For Git services, the secret should contain the `username` and `password` keys, with the password
set to a personal access token that grants access for listing Pull Requests or Merge Requests
and Git branches.

Example secret:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: github-pat
  namespace: default
stringData:
    username: flux
    password: <GITHUB PAT>
```

Example secret reference:

```yaml
spec:
  secretRef:
    name: github-pat
```

### TLS certificate configuration

The `.spec.certSecretRef` field is optional and specifies the Kubernetes Secret containing the
TLS certificate used for connecting to the external service.
Note that the secret must be created in the same namespace as the ResourceSetInputProvider.

For Git services that use self-signed certificates, the secret should contain the `ca.crt` key.

Example secret:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: gitlab-ca
  namespace: default
stringData:
    ca.crt: |
      -----BEGIN CERTIFICATE-----
      MIIDpDCCAoygAwIBAgIUI7z
      ...
      -----END CERTIFICATE-----
```

Example certificate reference:

```yaml
spec:
  certSecretRef:
    name: gitlab-ca
```

### Reconciliation configuration

The reconciliation of behaviour of a ResourceSet can be configured using the following annotations:

- `fluxcd.controlplane.io/reconcile`: Enable or disable the reconciliation loop. Default is `enabled`, set to `disabled` to pause the reconciliation.
- `fluxcd.controlplane.io/reconcileEvery`: Set the reconciliation interval used for calling external services. Default is `10m`.
- `fluxcd.controlplane.io/reconcileTimeout`: Set the timeout for calling external services. Default is `1m`.

### Conditions

A ResourceSetInputProvider enters various states during its lifecycle, reflected as Kubernetes Conditions.
It can be [reconciling](#reconciling-fluxinstance) while fetching data from external services,
it can be [ready](#ready-fluxinstance), or it can [fail during reconciliation](#failed-fluxinstance).

The ResourceSetInputProvider API is compatible with the **kstatus** specification,
and reports `Reconciling` and `Stalled` conditions where applicable to
provide better (timeout) support to solutions polling the ResourceSetInputProvider to
become `Ready`.

#### Reconciling ResourceSetInputProvider

The flux-operator marks a ResourceSetInputProvider as _reconciling_ when it starts
the reconciliation of the same. The Condition added to the ResourceSetInputProvider's
`.status.conditions` has the following attributes:

- `type: Reconciling`
- `status: "True"`
- `reason: Progressing` | `reason: ProgressingWithRetry`

The Condition `message` is updated during the course of the reconciliation to
report the action being performed at any particular moment such as
fetching data from external services.

The `Ready` Condition's `status` is also marked as `Unknown`.

#### Ready ResourceSetInputProvider

The flux-operator marks a ResourceSetInputProvider as _ready_ when the
data fetching from external services is successful.

When the ResourceSet is "ready", the flux-operator sets a Condition with the
following attributes in the ResourceSet’s `.status.conditions`:

- `type: Ready`
- `status: "True"`
- `reason: ReconciliationSucceeded`

#### Failed ResourceSetInputProvider

The flux-operator may get stuck trying to reconcile and apply a
ResourceSetInputProvider without completing. This can occur due to some of the following factors:

- The authentication to the external service fails.
- The external service is unreachable.
- The data fetched from the external service is invalid.

When this happens, the flux-operator sets the `Ready` Condition status to False
and adds a Condition with the following attributes to the ResourceSet’s
`.status.conditions`:

- `type: Ready`
- `status: "False"`
- `reason: ReconciliationFailed`

The `message` field of the Condition will contain more information about why
the reconciliation failed.

While the ResourceSetInputProvider has one or more of these Conditions, the flux-operator
will continue to attempt a reconciliation with an
exponential backoff, until it succeeds and the ResourceSetInputProvider is marked as [ready](#ready-fluxinstance).

### Exported inputs status

After a successful reconciliation, the ResourceSetInputProvider status contains a list of exported inputs
that can be used in the ResourceSet templates.

Example:

```yaml
status:
   exportedInputs:
   - author: stefanprodan
     branch: kubernetes/helm-set-limits
     id: "4"
     sha: bf5d6e01cf802734853f6f3417b237e3ad0ba35d
     title: 'kubernetes(helm): Add default resources limits'
   - author: stefanprodan
     branch: feat/ui-footer
     id: "3"
     sha: 8492c0b5b2094fe720776c8ace1b9690ff258f53
     title: 'feat(ui): Add footer'
   - author: stefanprodan
     branch: feat/ui-color-scheme
     id: "2"
     sha: 8166bdecd6b078b9e5dd14fa3b7b67a847f76893
     title: 'feat(ui): Default color scheme'
```


internal/builder/profiles.go
```diff
@@ -89,6 +89,15 @@ const tmpNotificationPatch = `
     - op: add
       path: /spec/versions/2/schema/openAPIV3Schema/properties/spec/properties/resources/items/properties/kind/enum/-
       value: FluxInstance
+    - op: add
```
matheuscscp, line 92: Don't we want to send any notifications about `ResourceSet` or `ResourceSetInputProvider` as well like we do for `FluxInstance`?

api/v1/resourceset_types.go
```diff
@@ -158,18 +180,6 @@ func (in *ResourceSet) GetTimeout() time.Duration {
 	return timeout
 }
 
-// GetInputs returns the ResourceSet inputs array.
-func (in *ResourceSet) GetInputs() []ResourceSetInput {
```
matheuscscp, line 162: Instead of deleting this function we could actually put it to good use like you did for `(*ResourceSetInputProvider).GetDefaultInputs() (map[string]any, error)`.

We can move the code from here:

https://github.com/controlplaneio-fluxcd/flux-operator/blob/a06fccd81ace6c15343a35bb3a524c8dbcefa77d/internal/builder/resourceset.go#L26-L37

The signature would become:

```go
func (in *ResourceSet) GetInputs() ([]map[string]any, error)
```

In the current state this function doesn't have much value but if we parsed the JSONs then it would be useful.

api/v1/resourceset_types.go
```diff
@@ -57,6 +63,22 @@ type ResourceSetSpec struct {
...
+type InputsProvider struct {
+	// APIVersion of the inputs provider resource.
+	// When not set, the APIVersion of the ResourceSet is used.
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
+
+	// Kind of the inputs provider resource.
+	// +kubebuilder:validation:Enum=ResourceSetInputProvider
+	// +required
+	Kind string `json:"kind"`
```
matheuscscp, line 75: If we plan to support multiple object types here, shouldn't we define an interface `interface{ GetInputs() ([]map[string]any, error) }` and implement it for `ResourceSetInputProvider`? (Similar to source-controller's `Source` interface containing `GetArtifact() *Artifact`)

internal/controller/resourceset_controller.go
```diff
@@ -262,6 +280,29 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
...
+		key := client.ObjectKey{
+			Namespace: obj.GetNamespace(),
+			Name:      inputSource.Name,
+		}
+
+		if err := r.Get(ctx, key, &provider); err != nil {
+			return nil, fmt.Errorf("failed to get inputs from %s/%s: %w", key.Namespace, key.Name, err)
+		}
+
+		inputs = append(inputs, provider.Status.ExportedInputs...)
```
matheuscscp, line 297: As per my comment on the file `api/v1/resourceset_types.go`, if we plan to support multiple types as input providers, here we should not assume the type `fluxcdv1.ResourceSetInputProvider` and the last line here should actually be something like this:

```go
providerInputs, err := provider.GetInputs()
if err != nil {
	return nil, fmt.Errorf("failed to get inputs from %s/%s: %w", key.Namespace, key.Name, err)
}
inputs = append(inputs, providerInputs)
```

Also as per my comment about not deleting the method `(*ResourceSet).GetInputs()` and remaking it such that it parses the JSONs, both `*ResourceSetInputProvider` and `*ResourceSet` would implement the `InputProvider interface { GetInputs() ([]map[string]any, error) }`. In this case, we would separate this `for` loop in two: one for listing the objects implementing the `InputProvider` interface, and one for iterating over them calling `.GetInputs()`. If we do this, the `obj` variable can be appended to the end of this list and the static inputs from `*ResourceSet` can be fetched in the second `for` loop alongside the inputs from the referenced objects.

internal/controller/resourceset_manager.go
```diff
@@ -29,7 +43,66 @@ func (r *ResourceSetReconciler) SetupWithManager(mgr ctrl.Manager, opts Resource
...
+		}
+
+		return results
+	}
+}
+
+var exportedInputsChangePredicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		oldObj := e.ObjectOld.(*fluxcdv1.ResourceSetInputProvider)
+		newObj := e.ObjectNew.(*fluxcdv1.ResourceSetInputProvider)
```
matheuscscp, line 103: I'm not quite sure but it seems to be a pattern checking if `e.ObjectOld` or `e.ObjectNew` are `nil` and returning `false` in that case. I'm also an adept of not using defensive programming if you can avoid it, but I've seen this in many predicates of the Flux codebase. Why do we do that? Can these values actually be `nil` here? Would controller-runtime actually pass `nil` values here?

matheuscscp (commented):
First pass, still need to go through the `ResourceSetInputProvider` controller code and everything under `internal/gitprovider/`.

api/v1/resourceset_types.go
```diff
@@ -57,6 +63,22 @@ type ResourceSetSpec struct {
 	Wait bool `json:"wait,omitempty"`
 }
 
+type InputsProvider struct {
+	// APIVersion of the inputs provider resource.
+	// When not set, the APIVersion of the ResourceSet is used.
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
```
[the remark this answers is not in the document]
stefanprodan: Yes, we'll probably have to support ConfigMaps and Secret.

internal/builder/profiles.go
stefanprodan: Yes but not in this PR. This change is for enabling webhook receivers.

api/v1/resourceset_types.go
stefanprodan: Other kinds (Kubernetes ConfigMaps and Secrets) will not subscribe to this interface. I'm also not for premature optimisation, since now we only support one type, there is no need to introduce a interface.

internal/controller/resourceset_manager.go
stefanprodan: Here we cast to a pointer while in Flux we cast to a value.

api/v1/resourceset_types.go
matheuscscp: But we already do support two types here: `*ResourceSet`, which provides static inputs, and `*ResourceSetInputProvider`, which provides dynamic inputs. See the next comment below.

matheuscscp: Also, when we implement ConfigMaps and Secrets in another PR, we could easily have simple wrappers on top of those types to implement the interface:

```go
type configMapInputProvider struct { *corev1.ConfigMap }

func (in configMapInputProvider) GetInputs() ([]map[string]any, error)
```

internal/controller/resourceset_manager.go
stefanprodan: I know we have those in Flux, but I don't see how an update event can send nil values, update means you always have old/new.

internal/gitprovider/result.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, r := range results {
+		list = append(list, r.ToMapWithDefaults(defaults))
+	}
+
+	for _, item := range list {
+		input := make(map[string]*apiextensionsv1.JSON)
+		for k, v := range item {
+			b, err := json.Marshal(v)
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal value %v: %w", v, err)
```
matheuscscp, line 67:
```suggestion
				return nil, fmt.Errorf("failed to marshal value to JSON %v: %w", v, err)
```

api/v1/resourcesetinputprovider_types.go
```diff
@@ -0,0 +1,196 @@
...
+	// When connecting to a Git provider that uses self-signed certificates, the CA certificate
+	// must be set in the Secret under the 'ca.crt' key to establish the trust relationship.
+	// +optional
+	CertSecretRef *meta.LocalObjectReference `json:"certSecretRef,omitempty"`
+
+	// DefaultValues contains the default values for the inputs.
+	// These values are used to populate the inputs when the provider
+	// response does not contain them.
+	// +optional
+	DefaultValues ResourceSetInput `json:"defaultValues,omitempty"`
```
matheuscscp, line 58: What kind of inputs are you expecting providers not to have? Do you have anything specific in mind?

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -0,0 +1,399 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to call provider: %w", err)
+	}
+
+	// Update the object status with the exported inputs.
+	data, err := yaml.Marshal(exportedInputs)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to marshal exported inputs: %w", err)
+	}
+	obj.Status.ExportedInputs = exportedInputs
+	obj.Status.LastExportedRevision = digest.FromBytes(data).String()
```
matheuscscp, line 149: I assume `yaml.Marshal()` is deterministic, i.e. it sorts map keys?

internal/gitprovider/gitlab.go
```diff
@@ -0,0 +1,162 @@
...
+			break
+		}
+		glOpts.Page = resp.NextPage
+	}
+
+	return results, nil
+}
+
+// parseGitHubURL parses a GitLab URL and returns the host and project.
+func parseGitLabURL(ghURL string) (string, string, error) {
```
matheuscscp, line 150:
```suggestion
func parseGitLabURL(glURL string) (string, string, error) {
```

internal/gitprovider/gitlab_test.go
```diff
@@ -0,0 +1,230 @@
...
+	tests := []struct {
+		name       string
+		opts       Options
+		want       []Result
+		wantErrMsg string
+	}{
+		{
+			name: "filters branches by regex",
+			opts: Options{
+				Token: os.Getenv("GITLAB_TOKEN"),
```
matheuscscp, line 25: Are we setting this in our GitHub secrets?

internal/gitprovider/gitlab.go
```diff
@@ -0,0 +1,162 @@
...
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: opts.CertPool,
+			},
+		}
+		rtClient.HTTPClient.Transport = tr
+	}
+	glOpts = append(glOpts, gitlab.WithHTTPClient(rtClient.HTTPClient))
+
+	if host != "https://gitlab.com" {
```
matheuscscp, line 43: I assume we don't have a way to test this, do we?

matheuscscp (commented):
A few more comments, still need to review the ResourceSetInputProvider controller code and GitHub.

api/v1/resourcesetinputprovider_types.go
```diff
@@ -0,0 +1,196 @@
...
+	// 'username' and 'password'.
+	// When connecting to a Git provider, the password should be a personal access token
+	// that grants read-only access to the repository.
+	// +optional
+	SecretRef *meta.LocalObjectReference `json:"secretRef,omitempty"`
+
+	// CertSecretRef specifies the Kubernetes Secret containing either or both of
+	//
+	// - a PEM-encoded CA certificate (`ca.crt`)
+	// - a PEM-encoded client certificate (`tls.crt`) and private key (`tls.key`)
```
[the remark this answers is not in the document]
matheuscscp: (The mTLS part.)

api/v1/resourcesetinputprovider_types.go
stefanprodan: Anything app specific e.g. chart name, env, perms etc. See the docs for some example.

internal/controller/resourcesetinputprovider_controller.go
stefanprodan: Yeap it's alphabetical sort

internal/gitprovider/gitlab.go
stefanprodan: No we rely on all Flux users running GitHub EE, it's the same logic as in NC/CLI.

internal/gitprovider/gitlab_test.go
stefanprodan: No this is for local dev, I've been running these tests a lot and with the tokens I'm avoiding rate limits.

matheuscscp: Oh right. So what's gonna happen in our CI exactly?

api/v1/resourcesetinputprovider_types.go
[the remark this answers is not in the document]
stefanprodan: Not for Git, as documented in the spec Git services only accept a CA. The client certs would be used for when calling 3rd-party HTTPS services.

internal/gitprovider/gitlab_test.go
stefanprodan: CI uses anonymous access

stefanprodan: So we are making 20 API calls per Git provider in every CI run, which shouldn't trigger any rate limits.

Issue #139, opened 2025-01-11
Need authentication to pull the flux-operator-manifests from a private OCI registry


ahothan:
In the case flux-operator-manifests is stored in a private OCI repository, FluxInstance fails to retrieve the manifest and there is no way to provide the auth token in the latest flux-operator version (v0.13.0).
The imagePullSecret field is only used for flux controller images.

In our use case, we will use the same OCI auth token to pull the flux-operator-manifests and flux controller images as they are stored in the same OCI registry but this will likely require a different field to be more generic, e.g. add an extra field "manifestImagePullSecret" under distribution?

stefanprodan:
I propose we use `.spec.distribution.manifestPullSecret`.

For the implementation we could do the following:

- import `github.com/google/go-containerregistry/pkg/authn/k8schain`
- add an `authn.Keychain` arg to the  [builder.PullArtifact](https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/internal/builder/pull.go) function
- if `manifestPullSecret` is specified, we create a keychain using `k8schain.NewFromPullSecrets`
- inside `builder.PullArtifact` we use `crane.WithAuthFromKeychain`

@ahothan let me know if you can contribute this feature and I'll assigned it to you. Thanks!

heytrav:
I've noticed that since today I seem to be unable to deploy the helm chart:

```
helm install flux-operator oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator \
  --namespace flux-system
Found existing alias for "helm install". You should use: "hin"
Error: INSTALLATION FAILED: GET "https://ghcr.io/v2/controlplaneio-fluxcd/charts/flux-operator/tags/list": GET "https://ghcr.io/token?scope=repository%3Acontrolplaneio-fluxcd%2Fcharts%2Fflux-operator%3Apull&service=ghcr.io": unexpected status code 403: denied: denied
```

Is this related?

ahothan:
@stefanprodan I should be able to get a PR shortly thanks for the implementation suggestions

PR #144, opened 2025-01-14
Add manifest pull secret to distribution

ahothan:


ahothan:
addresses https://github.com/controlplaneio-fluxcd/flux-operator/issues/139

PR #147, opened 2025-01-15
Update cluster type documentation to reflect schema validation

plobsing:
The [validation spec](https://github.com/controlplaneio-fluxcd/flux-operator/blob/e916d852131d2b19b0beac180e89d5cdd36298d3/api/v1/fluxinstance_types.go#L165) declared for the operator declares these using the cloud vendor names rather than the k8s implementation names.

This has already been updated in the distribution docs by https://github.com/controlplaneio-fluxcd/distribution/pull/145


stefanprodan (approved):
LGTM

Thanks @plobsing

PR #159, opened 2025-01-28
Use `fluxcd/pkg/runtime/cel` for ResourceSet dependencies ready expression evaluation

matheuscscp:

PR #160, opened 2025-02-01
Adhere to namespace/name standard in error/log messages in ResourceSet

matheuscscp:
Followup: #159


stefanprodan:
You would usually depend on CRDs and those have no namespace, better to use `ssautil.FmtObjMetadata`

stefanprodan (approved):

PR #202, opened 2025-03-01
Add caching for GitHub App tokens

matheuscscp:
This PR introduces a token cache in the operator. At the moment, we use it for GitHub App tokens in the `ResourceSetInputProvider` API.

This token cache will report the following metrics:

* Counter `flux_token_cache_events_total`
  * Help: `Total number of cache retrieval events for a Gitops Toolkit resource reconciliation.`
  * Labels: `event_type`, `kind`, `name`, `exported_namespace`. The values of `event_type` can be `cache_miss` or `cache_hit`.
* Counter `flux_token_cache_requests_total`
  * Help: `Total number of cache requests partioned by success or failure.`
  * Labels: `status`. The values of `status` can be `success` or `failure`.
* Counter `flux_token_cache_evictions_total`
  * Help: `Total number of cache evictions.`
  * No labels.
* Gauge `flux_token_cached_items`
  * Help: `Total number of items in the cache.`
  * No labels.


cmd/operator/main.go
```diff
@@ -158,6 +166,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	var tokenCache *cache.TokenCache
+	if tokenCacheMaxSize > 0 {
+		tokenCache = cache.NewTokenCache(tokenCacheMaxSize,
+			cache.WithMaxDuration(tokenCacheMaxDuration),
+			cache.WithMetricsRegisterer(reporter.Registerer()))
```
stefanprodan, line 173: What's the metric name and labels, I can't tell from this config, but it needs to be `flux_` prefixed.

matheuscscp: Maybe we should add this `flux_` prefix in `fluxcd/pkg/cache`? See this PR: https://github.com/fluxcd/pkg/pull/877

https://github.com/fluxcd/pkg/pull/877/files#diff-1eb58a7a2b6b183a0788cff2b8e29b60c7be6afff5bb2e07068b842a8744ddd6R73

In this PR I'm hard-coding `token_` internally in the `cache.TokenCache` constructor since it's already a "terminal" use case, we won't have specialized token caches, only a single token cache. Maybe this prefix should be `flux_cache_`?

(Please stamp this 877 and https://github.com/fluxcd/pkg/pull/878)

matheuscscp: Names and labels are here: https://github.com/fluxcd/pkg/blob/ff04927148b1b1407f07f58ffc3bcded59eb7a14/cache/metrics.go#L46

stefanprodan: We cant' hardcode the metric name in the package, Flux Operator needs `flux_` while the Flux Controller would use the GitOps Toolkit prefix `gotk_`.

Please post here the full metric name and labels, this needs to be documented in: https://fluxcd.control-plane.io/operator/monitoring/

matheuscscp: Cool I'll remove this from 877. Can you please stamp it anyway (and also 878)? I think it'd be nice to have a `token-cache-max-duration` flag as well. (For OIDC tokens for example users may want to decrease this max duration as the token caches an impersonation permission that could have been revoked).

matheuscscp: I added the metrics to the PR description, thanks!

stefanprodan:
All metrics exported by Flux Operator use [exported_namespace](https://github.com/controlplaneio-fluxcd/flux-operator/blob/cc6818a7e89ea2825ef5fa718b0da5fe94947dc5/internal/reporter/metrics.go#L144) to avoid collision with Prometheus Operator collector where the namespace is set to the pod/service namespace. Can we change that label?

matheuscscp:
> All metrics exported by Flux Operator use [exported_namespace](https://github.com/controlplaneio-fluxcd/flux-operator/blob/cc6818a7e89ea2825ef5fa718b0da5fe94947dc5/internal/reporter/metrics.go#L144) to avoid collision with Prometheus Operator collector where the namespace is set to the pod/service namespace. Can we change that label?

Sure, we need a change in the cache lib

matheuscscp:
> All metrics exported by Flux Operator use [exported_namespace](https://github.com/controlplaneio-fluxcd/flux-operator/blob/cc6818a7e89ea2825ef5fa718b0da5fe94947dc5/internal/reporter/metrics.go#L144) to avoid collision with Prometheus Operator collector where the namespace is set to the pod/service namespace. Can we change that label?

Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #204, opened 2025-03-05
Add timeout to ResourceSetInputProvider

matheuscscp:


internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -282,6 +282,9 @@ func (r *ResourceSetInputProviderReconciler) callProvider(ctx context.Context,
 		return nil, err
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, obj.GetTimeout())
+	defer cancel()
```
stefanprodan, line 286: Move this to line 166 and please don't override the controller runtime ctx, give it another name. We need to pass the timeout ctx to both newGitProvider and callProvider.

PR #223, opened 2025-03-26
Update fluxcd/pkg/runtime to v0.58.0

matheuscscp:
Includes https://github.com/fluxcd/pkg/pull/891

Also conditionally adding the service account impersonation like in the KC PR.


stefanprodan (approved):
LGTM

Thanks @matheuscscp

PR #230, opened 2025-04-08
Fully integrate ResourceSet APIs with notification-controller

### LINKED ISSUE #224: Support ResourceSet in Flux notifications
As a user of Flux and Flux Operator, I want to enable Flux notifications for all Flux Operator resources. `FluxInstance` is currently supported, but not `ResourceSet` (and probably not `ResourceSetInputProvider`). I tried `ResourceSet`, and got the following error: `Alert/flux-tenants/stas-teams-error dry-run failed (Invalid): Alert.notification.toolkit.fluxcd.io "stas-teams-error" is invalid: spec.eventSources[10].kind: Unsupported value: "ResourceSet": supported values: "Bucket", "GitRepository", "Kustomization", "HelmRelease", "HelmChart", "HelmRepository", "ImageRepository", "ImagePolicy", "ImageUpdateAutomation", "OCIRepository", "FluxInstance"`

#### Background

We had a bug in one of our ResourceSet templates: `build failed: failed to build resources[0]: failed to execute template: template: resourceset:4:11: executing "resourceset" at <inputs>: map has no entry for key "name"`. This can happen, and we had to check the resource inside the cluster to understand what was going on. As we have notifications already set up in this namespace, we would prefer the ability to get errors like this forwarded to our notification provider.

matheuscscp:
Closes #224

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -273,6 +273,7 @@ func main() {
 
 	if err = (&controller.ResourceSetInputProviderReconciler{
 		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
 		StatusManager: controllerName,
 		EventRecorder: mgr.GetEventRecorderFor(controllerName),
 		TokenCache:    tokenCache,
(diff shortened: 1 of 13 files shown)


matheuscscp:
I tested this manually and it works:

Screenshot from 2025-04-10 10-46-31

By the way, TIL: server-side apply does not return a diff if the value of a `ConfigMap` `.data` key has changed. It returns a diff if the key itself changed or was added/removed, but if only a value is changing then no diff is returned :exploding_head:

erikgb:
Thanks @matheuscscp! 🏅

PR #250, opened 2025-05-09
Fix copyFrom not copying secret type

### LINKED ISSUE #248: ResourceSet copyFrom Secrets type
ResourceSet is not handling the type field for Secrets with copyFrom... it creates them with type `Opaque` which does not work with `secretRef` (for `OCIRepository` it should be `kubernetes.io/dockerconfigjson`)...

As discovered [here](https://github.com/controlplaneio-fluxcd/d2-fleet/issues/7)

matheuscscp:
Fixes #248

### DIFF
--- a/internal/controller/resourceset_controller.go
+++ b/internal/controller/resourceset_controller.go
@@ -550,6 +550,18 @@ func (r *ResourceSetReconciler) copyResources(ctx context.Context,
 				if err := kubeClient.Get(ctx, sourceName, secret); err != nil {
 					return fmt.Errorf("failed to copy data from Secret/%s: %w", source, err)
 				}
+				_, ok, err := unstructured.NestedString(objects[i].Object, "type")
+				if err != nil {
+					return fmt.Errorf("type field of Secret/%s is not a string: %w", source, err)
+				}
+				if !ok {
+					if secret.Type == "" {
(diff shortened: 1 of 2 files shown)


internal/controller/resourceset_controller.go
```diff
@@ -550,6 +550,12 @@ func (r *ResourceSetReconciler) copyResources(ctx context.Context,
 				if err := kubeClient.Get(ctx, sourceName, secret); err != nil {
 					return fmt.Errorf("failed to copy data from Secret/%s: %w", source, err)
 				}
+				if secret.Type == "" {
+					secret.Type = corev1.SecretTypeOpaque
+				}
+				if err := unstructured.SetNestedField(objects[i].Object, string(secret.Type), "type"); err != nil {
+					return fmt.Errorf("failed to copy type from Secret/%s: %w", source, err)
+				}
```
stefanprodan, line 558: We need to set the type only if it's not present in the ResourceSet https://github.com/controlplaneio-fluxcd/d2-fleet/blob/main/tenants/apps.yaml#L41-L48

So instead of checking the original secret, check the unstructured object first.

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #251, opened 2025-05-09
Add force reconcile option to ResourceSet

stefanprodan:
This PR introduces the `fluxcd.controlplane.io/force` annotation for ResourceSets.

When a ResourceSet is annotated with `fluxcd.controlplane.io/force: enabled`, the controller will replace the generated resources if the reconciliation fails due to immutable field changes.


matheuscscp (approved):
LGTM :ok_hand:

Issue #203, opened 2025-03-05
Decouple static `ResourceSet` instances from `ResourceSet`


matheuscscp:
This idea came from @cicci8ino, he has this use case in his home lab.

As a user I'd like to be able to create static instances of a `ResourceSet` template that are decoupled from the `ResourceSet` itself. This way I don't need to touch the `ResourceSet` anymore when I want to create more static instances, i.e. I don't touch the resource containing the template anymore. This is a reasonable choice from the perspective of using `ResourceSets` as templates, if I have a resource that encapsulates my template, I don't like the idea of having to specify the entire list of instances directly on the template resource itself.

## Proposal

One way to do this would be adding a field `matchLabels` to `resourceset.spec.inputsFrom` and making `resourceset.spec.inputsFrom.name` not be required anymore, but rather requiring one of `name` or `matchLabels`. For example:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSet
...
spec:
  ...
  inputsFrom:
  - kind: ResourceSetInputProvider
    matchLabels:
      tenant: my-tenant
```

On the `ResourceSetInputProvider` side, we could introduce a new value for `spec.type`, `Static`, and copy `spec.defaultValues` over to `status.exportedInputs` when this new type is used. For example:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSetInputProvider
...
spec:
  type: Static
  defaultValues:
    some: value
status:
  exportedInputs:
  - id: <object UID>
    some: value
```

### Alternative

Another way to achieve this would be:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSet
...
spec:
  ...
  inputsFrom:
  - kind: ConfigMap
    matchLabels:
      tenant: my-tenant
```

And then look for a key e.g. `inputs.yaml` in the `ConfigMap` and parse it from YAML. This alternative has a few disadvantages:

* Users have to write YAML in a multi-line string, while in the `ResourceSetInputProvider` alternative we can provide a way for users to write real YAML.
* We need to introduce support for a new kind `ConfigMap` in `ResourceSet`.
* Users may want to define which key to look at instead of `inputs.yaml`, so we would need something specific for `ConfigMap` in the `inputsFrom` structure.

stefanprodan:
[...]

PR #252, opened 2025-05-09
Decouple static ResourceSet instances from ResourceSet

matheuscscp:
Closes: #203


api/v1/resourceset_types.go
```diff
@@ -65,6 +65,8 @@ type ResourceSetSpec struct {
 	Wait bool `json:"wait,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="self.name != '' || has(self.selector)", message="at least one of name or selector must be set for input provider references"
+// +kubebuilder:validation:XValidation:rule="self.name == '' || !has(self.selector)", message="cannot set both name and selector for input provider references"
```
stefanprodan, line 69:
```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.name) || !has(self.selector)", message="at least one of name or selector must be set for input provider references"
// +kubebuilder:validation:XValidation:rule="has(self.name) || has(self.selector)", message="cannot set both name and selector for input provider references"
```

api/v1/resourcesetinputprovider_types.go
```diff
@@ -15,24 +15,31 @@ import (
...
 	InputProviderGitLabBranch       = "GitLabBranch"
 	InputProviderGitLabMergeRequest = "GitLabMergeRequest"
+
+	ReasonInvalidDefaultValues  = "InvalidDefaultValues"
+	ReasonInvalidExportedInputs = "InvalidExportedInputs"
 )
 
 // ResourceSetInputProviderSpec defines the desired state of ResourceSetInputProvider
+// +kubebuilder:validation:XValidation:rule="self.type != 'Static' || self.url == ''", message="spec.url must be empty when spec.type is 'Static'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'Static' || self.url != ''", message="spec.url must not be empty when spec.type is not 'Static'"
```
stefanprodan, line 32:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.type != 'Static' || !has(self.url)", message="spec.url must be empty when spec.type is 'Static'"
// +kubebuilder:validation:XValidation:rule="self.type == 'Static' || has(self.url)", message="spec.url must not be empty when spec.type is not 'Static'"
```

api/v1/resourceset_types.go
matheuscscp: Good catch! I have no idea why `go fmt ./...` transforms `''` in `”`.

Btw the logic in XValidation is inverted, for "at least one" we need "has(...) || has(...)" and for "cannot set both" we need "!has(...) || !has(...)"

api/v1/resourcesetinputprovider_types.go
```diff
@@ -191,6 +204,10 @@ func (in *ResourceSetInputProvider) GetInputs() ([]map[string]any, error) {
 			}
 			inp[k] = data
 		}
+		if withProviderRef {
+			inp["providerName"] = in.GetName()
```
matheuscscp, line 208: What happens for ResourceSets containing both spec.inputs and spec.inputsFrom? Would it then be mandatory for templates to branch on the presence of these fields? I think we should add the entire reference for all objects providing inputs, including ResourceSet itself via spec.inputs:

* `providerAPIVersion`
* `providerKind`
* `providerName`
* `providerNamespace`

This would allow us to remove this ugly `GetInputsWithProviderRef()` and this logic altogether from the `api` package and do this inside `(r *ResourceSetReconciler) getInputs()`:

```go
	// List inputs from all providers.
	var inputs []map[string]any
	for _, provider := range providers {
		exportedInputs, err := provider.GetInputs()
		if err != nil {
			return nil, fmt.Errorf("failed to get inputs from %s/%s: %w",
				provider.GroupVersionKind().Kind, provider.GetName(), err)
		}
		providerType := provider.GroupVersionKind()
		providerAPIVersion := providerType.GroupVersion().String()
		providerKind := providerType.Kind
		providerName := provider.GetName()
		providerNamespace := provider.GetNamespace()
		for _, input := range exportedInputs {
			input["providerAPIVersion"] = providerAPIVersion
			input["providerKind"] = providerKind
			input["providerName"] = providerName
			input["providerNamespace"] = providerNamespace
		}
		inputs = append(inputs, exportedInputs...)
	}
```

internal/controller/resourceset_controller_test.go
```diff
@@ -641,6 +641,266 @@ spec:
...
+			InputsFrom: []fluxcdv1.InputProviderReference{{
+				Kind: fluxcdv1.ResourceSetInputProviderKind,
+			}},
+		},
+	})
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("at least one of name or selector must be set for input provider references"))
+}
+
+func TestResourceSetReconciler_LabelSelector(t *testing.T) {
```
stefanprodan, line 685: Please add a new test with for GC with just label selectors:
- Start with no RSIPs - RSET should reconcile with an empty inventory
- Add a couple of RSIPs that match the selector - check RSET inventory is filled 
- Delete one RSIP - check RSET inventory was updated
- Delete all RSIPs - RSET should reconcile with an empty inventory and GC should delete all items

internal/controller/resourceset_controller.go
```diff
@@ -296,36 +296,103 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
 
 func (r *ResourceSetReconciler) getInputs(ctx context.Context,
```
stefanprodan, line 297: Please add a comment to this function to explain what it does including deduplication

internal/controller/resourceset_controller.go
```diff
@@ -296,36 +296,103 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
...
+	getInputProviderKey := func(provider fluxcdv1.InputProvider) inputProviderKey {
+		return inputProviderKey{
+			GVK:       provider.GroupVersionKind(),
+			Name:      provider.GetName(),
+			Namespace: provider.GetNamespace(),
 		}
+	}
+
+	// Initialize the map of providers with the ResourceSet itself.
+	obj.SetGroupVersionKind(fluxcdv1.GroupVersion.WithKind(fluxcdv1.ResourceSetKind))
```
stefanprodan, line 317: Do we really need to mutate the RSET here, it doesn't contain the GVK?

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -86,12 +86,19 @@ The `.spec.type` field is required and specifies the type of the provider.
...
 - `GitHubBranch`: fetches input values from GitHub repository branches.
 - `GitLabMergeRequest`: fetches input values from opened GitLab Merge Requests.
 - `GitLabBranch`: fetches input values from GitLab project branches.
 
-For all types, the flux-operator will export in `.status.exportedInputs` a
+For the `Static` type, the flux-operator will export in `.status.exportedInputs` a
+single input map with the values from the field `.spec.defaultValues` and the
+additional value:
+
+- `id`: the UID of the ResourceSetInputProvider (type string).
```
stefanprodan, line 99:
```suggestion
- `id`: the Adler-32 checksum of the ResourceSetInputProvider UID (type string).
```

Shorter IDs means users don't have to worry about hitting the max length of Flux object names.

internal/controller/resourceset_controller.go
matheuscscp: Yes, this is always the case when you `.Get()` with a concrete type, the tests fail without this

matheuscscp:
All fixed, thanks!

internal/controller/resourceset_controller.go
```diff
@@ -294,38 +294,121 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
...
+	}
+
+	// getInputProviderKey returns the key for the input provider.
+	getInputProviderKey := func(provider fluxcdv1.InputProvider) inputProviderKey {
+		return inputProviderKey{
+			GVK:       provider.GroupVersionKind(),
+			Name:      provider.GetName(),
+			Namespace: provider.GetNamespace(),
 		}
+	}
```
stefanprodan, line 320: I suggest moving the `InputProviderKey` type and all the logic around it to `internal/inputs`. The `Get` function could accept the controller-runtime `Object` so it will work for both RSET and RSIP.

internal/controller/resourceset_controller.go
```diff
@@ -294,38 +294,121 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
...
+		providerAPIVersion := providerType.GroupVersion().String()
+		providerKind := providerType.Kind
+		providerName := provider.GetName()
+		providerNamespace := provider.GetNamespace()
+		for _, input := range exportedInputs {
+			input["providerAPIVersion"] = providerAPIVersion
+			input["providerKind"] = providerKind
+			input["providerName"] = providerName
+			input["providerNamespace"] = providerNamespace
+		}
```
stefanprodan, line 411: This could be move to `internal/inputs` into a function for `InputProviderKey`.

matheuscscp:
All fixed, thanks!

docs/api/v1/resourceset.md
```diff
@@ -152,6 +152,25 @@ spec:
       name: podinfo-pull-requests
 ```
 
+Example of inputs generated from multiple ResourceSetInputProviders via
```
stefanprodan, line 155: We need a sub-section in the inputs with the builtin ones providerNamespace and providerName, we also need to explain how these come from RS for inline inputs vs RSIP.

matheuscscp: Fixed, thanks!

docs/api/v1/resourceset.md
```diff
@@ -374,6 +393,25 @@ spec:
...
+#### Built-in input fields
+
+For every set of inputs, the reference of the object providing those inputs is injected
+by Flux Operator when templating the resources. The reference can be used in templates
+through the `inputs` object, through the following fields:
+
+- `inputs.provider.apiVersion`: The API version of the object providing the inputs.
+- `inputs.provider.kind`: The kind of the object providing the inputs.
+- `inputs.provider.name`: The name of the object providing the inputs.
+- `inputs.provider.namespace`: The namespace of the object providing the inputs.
```
stefanprodan, line 405: What happens if I have a RSET with in-line inputs like:

```yaml
spec:
  inputs:
   - provider: something
```

matheuscscp: The `provider` key is entirely replaced:

```go
	input["provider"] = map[string]any{
		"apiVersion": providerType.GroupVersion().String(),
		"kind":       providerType.Kind,
		"name":       provider.GetName(),
		"namespace":  provider.GetNamespace(),
	}
```

I added the following to the end of the docs section:

```
The `inputs.provider` field is always overwritten with the data above and cannot
be set by the user.
```

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #253, opened 2025-05-10
Implement `apply_kubernetes_manifest` MCP tool

stefanprodan:
This tool allows creating or updating Kubernetes resources in the cluster.
If the resources already exist and are managed by Flux, the tool will error out unless explicitly told to overwrite them.

### apply_kubernetes_manifest

Applies a YAML manifest on the cluster using Kubernetes server-side apply.

**Parameters:**

- `yaml_content` (required): The multi-doc YAML content
- `overwrite` (optional): Whether to overwrite resources managed by Flux (default: false)

**Output:**

The list of applied resources in the format `kind/namespace/name [created|updated|unchanged]`.

### DIFF
--- a/docs/mcp/tools.md
+++ b/docs/mcp/tools.md
@@ -188,7 +188,8 @@ Confirmation message indicating the resource has been resumed.
 ## Apply Tool
 
 This tool allows creating or updating Kubernetes resources in the cluster.
-Note that if the resources are managed by Flux, the changes will be undone on the next reconciliation.
+If the resources already exist and are managed by Flux, the tool will error out unless
+explicitly told to overwrite them.
 
 ### apply_kubernetes_manifest
 
@@ -196,7 +197,8 @@ Applies a YAML manifest on the cluster using Kubernetes server-side apply.
 
 **Parameters:**
 
-- `yaml` (required): The multi-doc YAML content
+- `yaml_content` (required): The multi-doc YAML content
+- `overwrite` (optional): Whether to overwrite resources managed by Flux (default: false)
 
 **Output:**
 
--- a/docs/mcp/tools.md
+++ b/docs/mcp/tools.md
@@ -185,6 +185,23 @@ Resumes the reconciliation of a previously suspended Flux resource.
 
 Confirmation message indicating the resource has been resumed.
 
+## Apply Tool
+
+This tool allows creating or updating Kubernetes resources in the cluster.
+Note that if the resources are managed by Flux, the changes will be undone on the next reconciliation.
+
+### apply_kubernetes_manifest
+
+Applies a YAML manifest on the cluster using Kubernetes server-side apply.
+
+**Parameters:**
+
+- `yaml` (required): The multi-doc YAML content
+
+**Output:**
+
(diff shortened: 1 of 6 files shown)


stefanprodan:
This tool is great for prototyping 😄 

Example prompt:

```test
Generate a test namespace. Then copy the podinfo Helm release and its source from the frontend namespace to the test namespace. Make sure to also copy the flux service account and RBAC.

...

Change the replica count to 2 in the podinfo values and check the deployment.

...

Delete the test namespace.
```

docs/mcp/tools.md
```diff
@@ -185,6 +185,23 @@ Resumes the reconciliation of a previously suspended Flux resource.
 
 Confirmation message indicating the resource has been resumed.
 
+## Apply Tool
+
+This tool allows creating or updating Kubernetes resources in the cluster.
+Note that if the resources are managed by Flux, the changes will be undone on the next reconciliation.
```
matheuscscp, line 191: This is a line I have always thought about never crossing when building a UI/assistant for Flux. Here we're giving people a tool that strays from the path of GitOps, especially with this new mindset of letting AI do everything on your behalf/vibe coding ( :vomiting_face: ).

I'd love to see tools for helping you write Flux YAML manifests, though, that you can copy and paste in a PR for somebody to review following the GitOps principle, or that you can choose to apply with `kubectl` yourself in the cluster and be accountable for it.

stefanprodan: There are many valid use-cases for this tool. During an incident, you would suspend Flux and instead of using kubectl or some UI to patch in-cluster objects, you would use this tool. The major advantage is that you can tell it to port the changes to the files in the repo once the incident is solved. Another use-case is experimentation, you tell it to copy Flux objects to a temp namespace, make edits than dump them to the repo. And yet another use-case, is to bootstrap the operator with natural language.

matheuscscp:
> This tool is great for prototyping 😄
> 
> Example prompts:
> 
> * Generate a test namespace. Then copy the podinfo Helm release and its source from the frontend namespace to the test namespace. Make sure to also copy the flux service account and RBAC.
> * Change the replica count to 2 in the podinfo values and check the deployment.
> * Delete the test namespace.

I think the same would work for an AI app that has `kubectl` access without our help, I'd prefer people to have this idea and try it on their own

docs/mcp/tools.md
matheuscscp: AIs hallucinate, I'm not comfortable with letting them apply things to clusters, especially on incidents. And as I wrote in the other comment, this would probably be possible without our help anyway. But as a security company I think we should not advertise this kind of practice, at least not yet

matheuscscp: Well, we already have delete/suspend/resume/reconcile anyway...

matheuscscp (approved):

matheuscscp:
Well, if people will try this anyway then maybe if we can help make it safer/saner it's probably for the best :shrug:

docs/mcp/tools.md
stefanprodan: All these tools can be disabled with the —read-only flag. Which is what we recommend im docs for production envs.

Issue #262, opened 2025-05-16
Feature: Support In-Cluster Deployment of Flux MCP Servers


ChrisJBurns:
We at Stacklok have developed [ToolHive](https://github.com/stacklok/toolhive) and we have created an Operator that allows people to deploy MCP servers into Kubernetes. Would it be possible to add the feature to the Flux MCP server that allows for InCluster config so that it can be deployed inside of the cluster? This would allow agents that live inside of the cluster to perform tasks using the Flux Operator MCP server. It also allows for users to connect to the Flux Operator MCP via SSE as well as ToolHive stands up a transparency proxy to the backend `stdio`/`sse`MCP servers.

If this is possible. We would be more than happy to write a blog post on how to deploy the Flux Operator MCP server into Kubernetes using our Operator.

As was [discussed previously](https://www.linkedin.com/feed/update/urn:li:ugcPost:7321477756561571840?commentUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7321477756561571840%2C7321484054845943809%29&replyUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7321477756561571840%2C7321485430921568256%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287321484054845943809%2Curn%3Ali%3AugcPost%3A7321477756561571840%29&dashReplyUrn=urn%3Ali%3Afsd_comment%3A%287321485430921568256%2Curn%3Ali%3AugcPost%3A7321477756561571840%29) running Flux Operator MCP locally inside of a Docker container wouldn't work greatly as you have to use the cloud provider binaries and add those into the image. We are starting to see that running MCP servers locally are great for developers, but for enterprise hosting them inside of a secured boundary such as Kubernetes is where we see the real traction being made.

Always happy to discuss more on this.

ChrisJBurns:
Yep, that sounds about right!

I'm happy to help with this. I already mocked some code up locally (not tested yet though) to see what it would look like with the `sigs.k8s.io/controller-runtime/pkg/client/config` package. In ToolHive we use [`rest.InClusterConfig`](https://github.com/stacklok/toolhive/blob/main/pkg/container/kubernetes/client.go#L56) but happy to mock something up with the  `sigs.k8s.io/controller-runtime/pkg/client/config` package on a local Kind cluster.

Using `sigs.k8s.io/controller-runtime/pkg/client/config` - not tested or even compiled yet. Just seeing how it may look
[...]

PR #265, opened 2025-05-24
Add support for in-cluster deployment of Flux MCP Server

stefanprodan:
This PR adds support for running the Flux MCP Server in SSE mode inside Kubernetes. The major difference from running the server locally is that the kubeconfig context switching tools are disabled, so comparing deployments across clusters is not possible. 

The server multi-arch container image will be available at `ghcr.io/controlplaneio-fluxcd/flux-operator-mcp`. The image embeds SBOMs and provenance attestations and is sign with Cosign and GitHub OIDC.

A Kubernetes deployment and service (port `9090`) are included in the Flux Operator distro, which allows deploying the MCP Server in the flux-system namespace:

```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: OCIRepository
metadata:
  name: flux-operator-manifests
  namespace: flux-system
spec:
  interval: 120m
  url: oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests
  ref:
    tag: latest
---
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: flux-operator-mcp
  namespace: flux-system
spec:
  interval: 60m
  wait: true
  timeout: 5m
  retryInterval: 5m
  prune: true
  sourceRef:
    kind: OCIRepository
    name: flux-operator-manifests
  path: ./flux-operator-mcp
  patches:
    - patch: |
        - op: add
          path: /spec/template/spec/containers/0/args/-
          value: --read-only=false
      target:
        kind: Deployment
```

Closes: #262

### DIFF
--- a/cmd/mcp/main.go
+++ b/cmd/mcp/main.go
@@ -71,7 +71,6 @@ func init() {
 }
 
 func main() {
-	log.SetFlags(0)
 	ctrllog.SetLogger(logr.New(ctrllog.NullLogSink{}))
 
 	if err := rootCmd.Execute(); err != nil {
@@ -157,7 +156,7 @@ var serveCmd = &cobra.Command{
 func serveCmdRun(cmd *cobra.Command, args []string) error {
 	inCluster := os.Getenv("KUBERNETES_SERVICE_HOST") != ""
 	if inCluster {
-		log.Printf("Starting in-cluster MCP server with read-only mode: %t", rootArgs.readOnly)
+		log.Printf("Starting the MCP server in-cluster with read-only mode set to %t", rootArgs.readOnly)
 	}
 
 	if os.Getenv("KUBECONFIG") == "" && !inCluster {
--- a/cmd/mcp/main.go
+++ b/cmd/mcp/main.go
@@ -155,7 +155,12 @@ var serveCmd = &cobra.Command{
 }
 
 func serveCmdRun(cmd *cobra.Command, args []string) error {
-	if os.Getenv("KUBECONFIG") == "" {
+	inCluster := os.Getenv("KUBERNETES_SERVICE_HOST") != ""
+	if inCluster {
+		log.Printf("Starting in-cluster MCP server with read-only mode: %t", rootArgs.readOnly)
+	}
+
+	if os.Getenv("KUBECONFIG") == "" && !inCluster {
 		return errors.New("KUBECONFIG environment variable is not set")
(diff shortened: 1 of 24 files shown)


cmd/mcp/main.go
```diff
@@ -155,7 +155,12 @@ var serveCmd = &cobra.Command{
 }
 
 func serveCmdRun(cmd *cobra.Command, args []string) error {
-	if os.Getenv("KUBECONFIG") == "" {
+	inCluster := os.Getenv("KUBERNETES_SERVICE_HOST") != ""
+	if inCluster {
+		log.Printf("Starting in-cluster MCP server with read-only mode: %t", rootArgs.readOnly)
```
matheuscscp, line 160:
```suggestion
		log.Printf("Starting in-cluster MCP server with read-only mode set to %t", rootArgs.readOnly)
```

The current version of this message is a bit confusing, I read this on the beginning of the review and understood that MCP in-cluster would run only in read-only mode, then I came back here and read the message again and understood that it's not the case

config/mcp/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+                 values:
+                   - linux
+      containers:
+      - name: server
+        image: flux-operator-mcp:latest
+        imagePullPolicy: IfNotPresent
+        args:
+          - serve
+          - --transport=sse
+          - --port=8080
```
matheuscscp, line 36: Maybe we should have `--read-only=true` as the default here and let users write a patch for allowing writes?

stefanprodan: No, flipping the default from when you run it locally would be super confusing.

config/mcp/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: flux-operator-mcp
+  labels:
+    app.kubernetes.io/name: flux-operator-mcp
+spec:
+  ports:
+  - name: http
+    port: 9090
```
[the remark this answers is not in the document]
stefanprodan: Changing the port would result in more confusing for users when it comes to allowing ingress traffic. Which port would would add the network policy 80 or 9090?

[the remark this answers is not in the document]
matheuscscp: Fair enough :+1:

Issue #148, opened 2025-01-16
`ResourceSet` time-based deploy window for Git branches


stefanprodan:
This proposal is an extension of #141 where instead of scanning Git branches for the latest commit continuously, we want to deploy changes from Git to the production clusters within a predefined time window. This caters to use-cases such as: maintenance windows, planned releases, No Deploy Friday.

## Example

Assuming we want to deploy changes to production every weekday from Monday to Thursday between 9:00 AM and 11:00 AM London time.

### Flux Operator Config

We can define a `ResourceSetInputProvider` that resolves the latest commit pushed to the `main` branch in the GitHub repository from where Flux reconciles apps to the production cluster with a `.spec.schedule`:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSetInputProvider
metadata:
  name: apps-main-branch
  namespace: flux-system
  annotations:
    fluxcd.controlplane.io/reconcile: "enabled"
    fluxcd.controlplane.io/reconcileEvery: "5m"
spec:
  schedule:
  - cron: 0 9 * * 1-4
    timeZone: Europe/London
    window: 2h
  type: GitHubBranch # can be GitLabBranch
  url: https://github.com/my-org/my-apps-repo
  secretRef:
    name: gh-token-apps
  filter:
    includeBranch: "^main$"
```

This provider will export a single set of inputs that contains the latest commit SHA of the main branch:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSetInputProvider
metadata:
  name: apps-main-branch
  namespace: flux-system
status:
  exportedInputs:
  - id: "1433470881"
    branch: "main"
    sha: "2dd3a8d2088457e5cf991018edf13e25cbd61380"
```

Outside of the schedule window, the exported inputs will remain unchanged, the SHA will be updated to the latest main HEAD only within the schedule window. If a GitHub/GitLab webhook is setup to trigger the provider, the webhook call will be ignored if it's outside the schedule window. 

To trigger a scan outside the schedule, users can annotate the `ResourceSetInputProvider` with
`reconcile.fluxcd.io/forceAt: "<current timestamp>"` and Flux Operator will reconcile only once then pause.

To deploy the apps, we can define a `ResourceSet` that pins the Flux `GitRepository` to the the commit SHA exported by the provider:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSet
metadata:
  name: apps
[...]

PR #288, opened 2025-06-11
Introduce schedule for ResourceSetInputProvider

matheuscscp:
Part of: #148 

Works well locally. Created the following RSIP and ran `./cli reconcile rsip test --force`:

```yaml
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"fluxcd.controlplane.io/v1","kind":"ResourceSetInputProvider","metadata":{"annotations":{},"name":"test","namespace":"flux-system"},"spec":{"schedule":[{"cron":"0 0 29 2 *","timeZone":"UTC","window":"10h"}],"type":"Static"}}
    reconcile.fluxcd.io/forceAt: 2025-06-12 11:32:01.600194055 +0100 IST m=+0.030152249
    reconcile.fluxcd.io/requestedAt: 2025-06-12 11:32:01.600194055 +0100 IST m=+0.030152249
  creationTimestamp: "2025-06-12T10:30:04Z"
  finalizers:
  - fluxcd.controlplane.io/finalizer
  generation: 1
  name: test
  namespace: flux-system
  resourceVersion: "4721994"
  uid: 0becd27f-e2c9-4da6-a789-9560ba5dc376
spec:
  schedule:
  - cron: 0 0 29 2 *
    timeZone: UTC
    window: 10h
  type: Static
status:
  conditions:
  - lastTransitionTime: "2025-06-12T10:32:01Z"
    message: Reconciliation finished in 2ms
    observedGeneration: 1
    status: "True"
    type: Ready
  exportedInputs:
  - id: "3173452256"
  lastExportedRevision: sha256:6a6bdea84fff5785d0aa9495a0eb8978c16f2f6612e8530d6b79c93abc430dec
  lastHandledForceAt: 2025-06-12 11:32:01.600194055 +0100 IST m=+0.030152249
  lastHandledReconcileAt: 2025-06-12 11:32:01.600194055 +0100 IST m=+0.030152249
  nextSchedule:
    cron: 0 0 29 2 *
    timeZone: UTC
    when: "2028-02-29T00:00:00Z"
    window: 10h0m0s
```


api/v1/schedule_types.go
```diff
@@ -0,0 +1,41 @@
...
+// Schedule defines a schedule for something to run.
+type Schedule struct {
+	// Cron specifies the cron expression for the schedule.
+	// +required
+	Cron string `json:"cron"`
+
+	// TimeZone specifies the time zone for the cron schedule. Defaults to UTC.
+	// +kubebuilder:default:="UTC"
+	// +optional
+	TimeZone string `json:"timeZone"`
```
matheuscscp: k8s `CronJob` also uses `timeZone`, not `tz`

cmd/cli/build_instance.go
```diff
@@ -202,10 +202,9 @@ func buildInstanceCmdRun(cmd *cobra.Command, args []string) error {
 		ssautil.SetCommonMetadata(objects, instance.Spec.CommonMetadata.Labels, instance.Spec.CommonMetadata.Annotations)
 	}
 
-	ownerGroup := fmt.Sprintf("%s", fluxcdv1.GroupVersion.Group)
 	ssautil.SetCommonMetadata(objects, map[string]string{
-		fmt.Sprintf("%s/name", ownerGroup):      instance.GetName(),
-		fmt.Sprintf("%s/namespace", ownerGroup): instance.GetNamespace(),
+		fmt.Sprintf("%s/name", fluxcdv1.GroupVersion.Group):      instance.GetName(),
+		fmt.Sprintf("%s/namespace", fluxcdv1.GroupVersion.Group): instance.GetNamespace(),
```
matheuscscp: Fixing linter warning

cmd/cli/get_inputprovider.go
```diff
@@ -0,0 +1,97 @@
...
+	err = kubeClient.List(ctx, &list, lsOpts)
+	if err != nil {
+		return err
+	}
+
+	var rows [][]string
+	for _, obj := range list.Items {
+		inputsCount := 0
+		if obj.Status.ExportedInputs != nil {
+			inputsCount = len(obj.Status.ExportedInputs)
```
matheuscscp:
```
./cli get rsip test
NAME	INPUTS	READY	MESSAGE                        
test	1     	True 	Reconciliation finished in 2ms	
```

cmd/cli/reconcile_inputprovider.go
```diff
@@ -0,0 +1,60 @@
...
+		return fmt.Errorf("name is required")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
+	defer cancel()
+
+	now := metav1.Now().String()
+
+	if !reconcileInputProviderArgs.force {
+		return annotateResource(ctx,
```
matheuscscp:
```
./cli reconcile rsip test
```

From `k events -w`:
```
Reconciliation skipped, next scheduled at 2028-02-29T00:00:00Z
```

cmd/cli/reconcile_inputprovider.go
```diff
@@ -0,0 +1,60 @@
...
+			fluxcdv1.ResourceSetInputProviderKind, args[0],
+			*kubeconfigArgs.Namespace,
+			meta.ReconcileRequestAnnotation,
+			now)
+	}
+
+	return annotateResourceWithMap(ctx,
```
matheuscscp:
```
./cli reconcile rsip test --force
```

From `k events -w`:
```
Reconciliation finished in 1ms
```

cmd/cli/reconcile_inputprovider.go
stefanprodan: How about we add a `—wait` flag that defaults to `true` for all reconcile commands and we wait and print the ready status?

stefanprodan:
Why is the Stalled condition present if the last generation is valid?

matheuscscp:
> Why is the Stalled condition present if the last generation is valid?

Looks like a bug in `finalizeStatus()`:

https://github.com/controlplaneio-fluxcd/flux-operator/blob/716e37c086f1bbfc1581129315474bcef320c331/internal/controller/resourcesetinputprovider_controller.go#L525-L550

It should remove Stalled when Ready is True.

stefanprodan:
It should't be in the `finalizeStatus()`, when we set `ReconcilingCondition` at the beginning we should remove Stalled.

matheuscscp:
> It should't be in the `finalizeStatus()`, when we set `ReconcilingCondition` at the beginning we should remove Stalled.

I think in the image automation controllers we use a library from fluxcd/pkg that we don't use anywhere else which does everything needed for status finalization, this sounds like the kind of behavior we want everywhere that shouldn't be placed in a part of the code we repeat for each controller

cmd/cli/reconcile_inputprovider.go
matheuscscp: Should also have `--timeout`

stefanprodan: We do have that https://github.com/controlplaneio-fluxcd/flux-operator/blob/716e37c086f1bbfc1581129315474bcef320c331/cmd/cli/main.go#L48

stefanprodan:
We already use `conditions.MarkReconciling` which does the right thing (removes Stalled and sets Reconciling), but for `Static` we exit every time before we reach this, and that's how the bug was introduced.

PS. I pushed a fix for this, please update the example.

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -117,6 +119,33 @@ func (r *ResourceSetInputProviderReconciler) reconcile(ctx context.Context,
 	log := ctrl.LoggerFrom(ctx)
 	reconcileStart := time.Now()
 
+	// Handle force reconciliation requests.
+	force := meta.ShouldHandleForceRequest(obj)
+
+	// Validate schedule.
+	schedule, err := schedule.New(obj.Spec.Schedule, obj.GetTimeout())
```
stefanprodan: Variable `schedule` collides with imported package name

cmd/cli/reconcile_inputprovider.go
matheuscscp: We will do this in a separate PR.

api/v1/schedule_types.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2024 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package v1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+const (
+	ReasonInvalidSchedule      = "InvalidSchedule"
+	ReasonSkippedDueToSchedule = "SkippedDueToSchedule"
```
stefanprodan, line 10: Please add these to the API docs under the `Status` section.

api/v1/schedule_types.go
stefanprodan:
```suggestion
	// +optional
	TimeZone string `json:"timeZone,omitempty"`
```

api/v1/schedule_types.go
```diff
@@ -0,0 +1,41 @@
...
+	// +optional
+	TimeZone string `json:"timeZone"`
+
+	// Window defines the time window during which the execution is allowed.
+	// Defaults to 0s, meaning no window is applied.
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m|h))+$"
+	// +kubebuilder:default:="0s"
+	// +optional
+	Window metav1.Duration `json:"window"`
```
stefanprodan, line 30:
```suggestion
	// +optional
	Window metav1.Duration `json:"window,omitempty"`
```

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -129,6 +158,10 @@ func (r *ResourceSetInputProviderReconciler) reconcile(ctx context.Context,
 		return ctrl.Result{}, nil
 	}
 
+	// Mark the object as reconciling.
+	const msgReconciling = "Reconciliation in progress"
```
stefanprodan: We can delete this, I realised we already have a global const called `msgInProgress` in the controller pkg.

internal/controller/resourcesetinputprovider_controller_test.go
```diff
@@ -145,6 +148,98 @@ func TestResourceSetInputProviderReconciler_reconcile_InvalidDefaultValues(t *te
...
+			Schedule: []fluxcdv1.Schedule{{
+				// This cron only happens once every 4 years, so most
+				// of the time the reconciliation will be skipped.
+				Cron:   "0 0 29 2 *",
+				Window: metav1.Duration{Duration: time.Second},
+			}},
+		},
+	}
+
+	schedule, err := schedule.Parse("0 0 29 2 *", "UTC")
```
stefanprodan:
```suggestion
	sched, err := schedule.Parse("0 0 29 2 *", "UTC")
```

internal/controller/resourcesetinputprovider_controller_test.go
```diff
@@ -842,3 +937,112 @@ spec:
...
+
+			obj := &fluxcdv1.ResourceSetInputProvider{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						fluxcdv1.ReconcileEveryAnnotation: tt.interval.String(),
+					},
+				},
+			}
+
+			schedule, err := schedule.New(tt.schedules, tt.timeout)
```
stefanprodan:
```suggestion
			sched, err := schedule.New(tt.schedules, tt.timeout)
```

internal/schedule/schedule.go
```diff
@@ -0,0 +1,213 @@
...
+	// cronLookBack is the maximum time range a previous cron trigger
+	// could have occurred in the past. It's 5 years because a cron
+	// schedule can be defined for Feb 29, which only occurs every 4
+	// years (0 0 29 2 *).
+	cronLookBack = 5 * 365 * 24 * time.Hour
+)
+
+// Schedule computes when reconciliations should happen according
+// to a list of schedule configurations.
+type Schedule struct {
```
stefanprodan:
```suggestion
type Scheduler struct {
```

internal/schedule/schedule.go
```diff
@@ -0,0 +1,213 @@
...
+	schedules []schedule
+	timeout   time.Duration
+}
+
+type schedule struct {
```
stefanprodan: This name collides with the package name, please rename.

internal/testing/time.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2025 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package pkgtesting
```
stefanprodan: Please rename the package to `testutils`

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -285,6 +285,70 @@ spec:
...
+When a schedule is specified with a non-zero duration window, flux-operator will only reconcile
+the ResourceSetInputProvider when the time point `time.Now().Add(obj.GetTimeout())` falls within
+the time window defined by the schedule. This check is performed not only at the start of the
+reconciliation, but also when scheduling the next reconciliation according to the interval defined
+by the `fluxcd.controlplane.io/reconcileEvery` [annotation](#reconciliation-configuration).
+
+To force a one-off out-of-schedule reconciliation, annotate the ResourceSetInputProvider with
+`reconcile.fluxcd.io/requestedAt: "<current timestamp>"` and
+`reconcile.fluxcd.io/forceAt: "<current timestamp>"`. Both annotations must be annotated
+with the exact same string, otherwise the reconciliation will not be triggered. This can
```
stefanprodan:
```suggestion
`reconcile.fluxcd.io/forceAt: "<current timestamp>"`. Both annotations values
must be the exact same string, otherwise the reconciliation will not be triggered. This can
```

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -285,6 +285,70 @@ spec:
...
+scheduled reconciliation. For example:
+
+```yaml
+status:
+  nextSchedule:
+    cron: 0 8 * * 1-5
+    timeZone: Europe/London
+    when: "2025-06-29T00:00:00Z"
+    window: 8h
+```
```
stefanprodan: This should be moved to the Status section.

The explanation can be simplified:

When `.spec.schedule` is set, the next scheduled reconciliation is reported in `.status. nextSchedule`.

api/v1/schedule_types.go
matheuscscp: Fixed, thanks!

internal/schedule/schedule.go
matheuscscp: Fixed, thanks!

api/v1/schedule_types.go
matheuscscp: A struct is never empty even if all the values inside are zero, so this has no effect what-so-ever. This is useful only for pointers, and here we want a default value anyway so there's no point in using a pointer either

internal/controller/resourcesetinputprovider_controller.go
matheuscscp: Used it everywhere, thanks!

api/v1/schedule_types.go
matheuscscp: Fixed, thanks!

internal/controller/resourcesetinputprovider_controller_test.go
matheuscscp: Fixed, thanks!

internal/controller/resourcesetinputprovider_controller_test.go
matheuscscp: Changed to `scheduler`, thanks!

internal/schedule/schedule.go
matheuscscp: Fixed, thanks!

internal/testing/time.go
matheuscscp: Fixed, thanks!

docs/api/v1/resourcesetinputprovider.md
matheuscscp: Fixed, thanks!

docs/api/v1/resourcesetinputprovider.md
matheuscscp: That's not true, though. During the window we don't report anything, the `.status.nextSchedule` field is empty. The interval is what is scheduling the next reconciliation, so there's no point in reporting this in the status during the window.

I moved this section to Status and modified the explanation so it starts simpler, and then we have a paragraph at the end explaining what happens during a window. Thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #290, opened 2025-06-12
Introduce get/suspend/resume/reconcile commands for ResourceSetInputProvider

matheuscscp:
Closes: #148

### DIFF
--- a/cmd/cli/get_inputprovider.go
+++ b/cmd/cli/get_inputprovider.go
@@ -1,0 +1,102 @@
+// Copyright 2025 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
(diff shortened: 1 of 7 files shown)


cmd/cli/get_inputprovider.go
```diff
@@ -0,0 +1,97 @@
...
+			ready,
+			conditions.GetMessage(&obj, "Ready"),
+		}
+		if getInputProviderArgs.allNamespaces {
+			row = append([]string{obj.Namespace}, row...)
+		}
+		rows = append(rows, row)
+	}
+
+	header := []string{"Name", "Inputs", "Ready", "Message"}
```
stefanprodan: Let's add here `Next Schedule`, if it's not in status can be empty.

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #292, opened 2025-06-13
cli: Add `--wait` flag to reconcile commands

stefanprodan:
This PR adds a `--wait` flag (default `true`) to the `flux-operator reconcile` commands. The CLI commands will now output the ready message either when the reconciliation was successful or if it resulted in an error.

```console
$ flux-operator reconcile rset apps -n flux-system
► Reconciliation triggered
◎ Waiting for reconciliation...
✔ Reconciliation finished in 57ms

$ flux-operator reconcile instance flux -n flux-system
► Reconciliation triggered
◎ Waiting for reconciliation...
✗ Reconciliation is disabled for flux
```

### DIFF
--- a/cmd/cli/reconcile_inputprovider.go
+++ b/cmd/cli/reconcile_inputprovider.go
@@ -18,16 +18,30 @@ var reconcileInputProviderCmd = &cobra.Command{
 	Use:     "inputprovider",
 	Aliases: []string{"rsip", "resourcesetinputprovider"},
 	Short:   "Trigger ResourceSetInputProvider reconciliation",
-	RunE:    reconcileInputProviderCmdRun,
+	Example: `  # Trigger the reconciliation of a ResourceSetInputProvider
+  flux-operator -n flux-system reconcile rsip my-inputprovider
+
+  # Force the reconciliation of a ResourceSetInputProvider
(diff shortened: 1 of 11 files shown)


cmd/cli/reconcile_inputprovider.go
```diff
@@ -18,16 +18,30 @@ var reconcileInputProviderCmd = &cobra.Command{
...
+  # Trigger the reconciliation of a ResourceSetInputProvider without waiting for it to become ready
+  flux-operator -n flux-system reconcile rsip my-inputprovider --wait=false
+`,
+	RunE: reconcileInputProviderCmdRun,
 }
 
-var reconcileInputProviderArgs struct {
+type reconcileInputProviderFlags struct {
 	force bool
+	wait  bool
```
matheuscscp, line 35: Can't we add this to the `reconcile` command and use in all the subcommands?

matheuscscp (approved):

cmd/cli/reconcile_inputprovider.go
stefanprodan: I regret doing this in the Flux CLI, it's really bad UX as all common flags endup at the bottom mixed with the kube ones which are a lot.

PR #293, opened 2025-06-13
Refactor common status operations

stefanprodan:
Introduce the `FluxObject` interface and refactor the common operations against status in all reconcilers.

This PR also fixes the initialisation of suspended resources and the initialisation of RSIP with a schedule in the future.


internal/controller/resourcesetinputprovider_controller_test.go
```diff
@@ -220,15 +220,15 @@ func TestResourceSetInputProviderReconciler_reconcile_SkippedDueToSchedule(t *te
 
 	expectedRequeueAfter := time.Until(sched.Next(time.Now())).Seconds()
 
-	g.Expect(requeueAfter).To(BeNumerically("~", expectedRequeueAfter, 0.01))
+	g.Expect(requeueAfter).To(BeNumerically("~", expectedRequeueAfter, 2*time.Second))
```
matheuscscp: Please remove the .Seconds() everywhere as it converts time.Duration to float64. The units are incorrect, `2*time.Second` is being converted to `2*10^9`, we are giving 2 billion seconds of tolerance. We need all three values in this line to be on the same unit

matheuscscp (commented):
Just one comment, the rest LGTM

matheuscscp:
We should review how we deal with initializing suspended objects in the Flux controllers

internal/controller/resourcesetinputprovider_controller_test.go
stefanprodan: Without `.Seconds()`

```
=== RUN   TestResourceSetInputProviderReconciler_reconcile_SkippedDueToSchedule
I0614 00:19:31.724991   32026 resourcesetinputprovider_controller.go:135] "msg"="Reconciliation skipped, next scheduled at 2028-02-29T00:00:00+02:00" 
    resourcesetinputprovider_controller_test.go:223: 
        Expected
            <time.Duration>: 85538428276618000
        to be within 2s of ~
            <time.Duration>: 85545628274848000
--- FAIL: TestResourceSetInputProviderReconciler_reconcile_SkippedDueToSchedule (0.01s)
```

PS. My local time is not UTC 0, so this test will always fail...

matheuscscp: Okay we need to normalize the time... Funny that my local time is UTC+1 which is Irish Summer Time, and the test still passes. I think the CI is in UTC. The difference in the 2 durations is exactly 2 hours, which I think is the  difference between UTC and your time zone.

I'll push a definitive fix to the branch

stefanprodan: Yes if i print the 2 durations it's 2h diffrence `requeueAfter: 23760h13m19.555277s  expected: 23762h13m19.553466s`

matheuscscp: Funny, the difference between UTC and Bucharest right now is 3h.

matheuscscp: Change "UTC" to "" on line 213... I forgot we don't have the default UTC set by API server during admission in this test case

matheuscscp: Yeah this is definitely the issue, the time.Now() calls all have the same time zone (your local time zone in this case), the only difference is in the time zones being passed to the cron lib, which are "" in the RSIP object and "UTC" in the test. "" is your local time zone.

After fixing this and using time.Duration for everything and 2*time.Second for tolerance it should work in your computer

stefanprodan: The actual issue is that not specifying `TimeZone: "UTC"` in spec, results in local time.

matheuscscp: Either way it should work for this test to pass, "" is a valid time zone for the cron lib. I guess the alternative is more "realistic" since in runtime we have the UTC default, so you can just pass UTC in the object spec above.

stefanprodan: Everywhere else in tests we persist the objects in etcd to run the whole Kubernetes API pipeline including the admission controller that would set the defaults. But here you call directly the private `reconcile` function bypassing the API.

matheuscscp: You have a good point, we can remove UTC in the object spec and create it instead:

```go
g.Expect(testEnv.Create(ctx, obj)).NotTo(HaveOccurred())
g.Expect(testEnv.Get(ctx, client.ObjectKeyFromObject(obj), obj)).NotTo(HaveOccurred())
```

Then we test the default as well.

stefanprodan: Well now that I've changed the test to use the actual reconciler shows how broken things are. Creating a RSIP with a schedule that doesn't trigger right away leaves the object in a transitional state:

```yaml
         conditions:
        - lastTransitionTime: "2025-06-14T06:58:35Z"
          message: Reconciliation in progress
          observedGeneration: 1
          reason: Progressing
          status: "True"
          type: Reconciling
        - lastTransitionTime: "2025-06-14T06:58:35Z"
          message: Reconciliation in progress
          observedGeneration: 1
          reason: Progressing
          status: Unknown
          type: Ready
        nextSchedule:
          cron: 0 0 29 2 *
          timeZone: UTC
          when: "2028-02-29T00:00:00Z"
          window: 1s
```

matheuscscp: Another great point, this is the same bug we are fixing in this PR but for time-based disabling. I'm trying to think how we can fix it... We need some sort of additional disabled check during initialization. Maybe we introduce a method in the RSIP object, we create an interface with this method and we test for this interface during the initialization internal library.

I'll push a fix

matheuscscp:
> We should review how we deal with initializing suspended objects in the Flux controllers

Another action point after this PR is standardizing how we signal reconciliations that were skipped (after creation) due to objects being suspended/disabled: In KC we print an `info` log and don't send an event. In flux-operator we print an `error` log and send a `Warning` event if the object is disabled via the annotation, but for RSIP disabled via schedule we print `info` and send `Normal`.

Don't think this is urgent though, we can address this after the next flux-operator release.

internal/controller/resourcesetinputprovider_controller_test.go
```diff
@@ -192,52 +186,79 @@ func TestResourceSetInputProviderReconciler_reconcile_SkippedDueToSchedule(t *te
...
+  type: Static
+  schedule:
+  - cron: "0 0 29 2 *" # This cron only happens once every 4 years.
+    window: 1s
+  defaultValues:
+    env: test
+`, ns.Name)
 
-	sched, err := schedule.Parse("0 0 29 2 *", "UTC")
+	// Create the ResourceSetInputProvide
```
matheuscscp:
```suggestion
	// Create the ResourceSetInputProvider
```

PR #295, opened 2025-06-14
Update some inconsistencies in FluxInstance docs

bgdnlp:
Creating first FluxInstance and going through docs, I noticed these. I'm not sure I'm doing any of it right.
- The `components` list defaults to 4 out of 6, it is correct in docs, but the API says "all".
- Cluster `domain` and `networkPolicy` are marked as required, but they have default values. This causes at least VSCode yaml-spec extension to complain when editing the manifest if those properties are not specified. In documentation they are marked as optional.
- Cluster `multitenant` defaults to false, but that's not documented.

### DIFF
--- a/docs/api/v1/fluxinstance.md
+++ b/docs/api/v1/fluxinstance.md
@@ -320,6 +320,7 @@ The supported values are `kubernetes` (default), `openshift`, `azure`, `aws` and
 
 The `.spec.cluster.multitenant` field is optional and specifies whether to enable Flux
 [multi-tenancy lockdown](https://fluxcd.io/flux/installation/configuration/multitenancy/).
+By default, it is `false` (disabled).
 
 The `.spec.cluster.tenantDefaultServiceAccount` is optional and specifies the default
 service account used by Flux when reconciling `Kustomization` and `HelmRelease`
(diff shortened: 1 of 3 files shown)


stefanprodan (commented):
Can you please run `make` to run the codegen and regenerate the CRDs. After this please push all changes.

docs/api/v1/fluxinstance.md
```diff
@@ -320,6 +320,7 @@ The supported values are `kubernetes` (default), `openshift`, `azure`, `aws` and
 
 The `.spec.cluster.multitenant` field is optional and specifies whether to enable Flux
 [multi-tenancy lockdown](https://fluxcd.io/flux/installation/configuration/multitenancy/).
+By default, it is 'false' (disabled).
```
matheuscscp:
```suggestion
By default, it is `false` (disabled).
```

bgdnlp:
Done, changed quotes and ran make,  but it says that merging is blocked because "commits must have verified signatures". I don't have that set up, do I need to?

matheuscscp:
We would appreciate it very much, please 🙏

https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key

https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

bgdnlp:
I tried to squash rebase for a prettier history and made a mess, so I started over. Opened a new PR, #296. Hopefully that one is good enough.

PR #297, opened 2025-06-16
Introduce `GitHubTag` and `GitLabTag` input providers

stefanprodan:
This PR Introduce `GitHubTag` and `GitLabTag` types for the ResourceSetInputProvider API and a new `semver` filter.

The new providers can be used to fetch the latest Git tag corresponding to a semver rage, for example:

```yaml
metadata:
  name: podinfo-releases
  namespace: apps
spec:
  type: GitHubTag
  url: https://github.com/stefanprodan/podinfo
  filter:
    limit: 1
    semver: ">=1.0.0"
status:
  conditions:
  - lastTransitionTime: "2025-06-16T10:43:47Z"
    message: Reconciliation finished in 202ms
    observedGeneration: 1
    status: "True"
    type: Ready
  exportedInputs:
  - id: "50462972"
    sha: fb3b01be30a3f353b221365cd3b4f9484a0885ea
    tag: 6.9.0
  lastExportedRevision: sha256:738cf4279af53ccbd7d392cdbb2a7e6d2113878fb38828fd0b1fe3b7b979c5f9
```


docs/api/v1/resourcesetinputprovider.md
```diff
@@ -136,6 +144,7 @@ The following filters are supported:
 - `labels`: filter GitHub Pull Requests or GitLab Merge Requests by labels.
 - `includeBranch`: regular expression to include branches by name.
 - `excludeBranch`: regular expression to exclude branches by name.
+- `semver`: sematic version ranger to filter and order tags.
```
matheuscscp:
```suggestion
- `semver`: sematic version range to filter and order tags.
```

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -149,6 +158,15 @@ spec:
     excludeBranch: "^feat/not-this-one$"
 ```
 
+Example of a filter configuration for fetching the latest Git tag according to semver:
```
matheuscscp:
```suggestion
Example of a filter configuration for fetching only the latest Git tag according to semver:
```

internal/gitprovider/github.go
```diff
@@ -64,18 +64,65 @@ func NewGitHubProvider(ctx context.Context, opts Options) (*GitHubProvider, erro
...
+
 func (p *GitHubProvider) ListBranches(ctx context.Context, opts Options) ([]Result, error) {
 	ghOpts := &github.BranchListOptions{
 		ListOptions: github.ListOptions{
 			PerPage: 100,
 		},
 	}
 
-	var results []Result
+	results := make([]Result, 0)
```
matheuscscp, line 121: Is the linter suggesting this form? I think `var x []something` is more idiomatic, as the zero value for a slice (nil) is always a valid slice with zero elements

internal/gitprovider/options.go
```diff
@@ -50,3 +54,33 @@ func matchLabels(opt Options, labels []string) bool {
...
+	var versions []*semver.Version
+	for _, tag := range tags {
+		if v, err := semver.NewVersion(tag); err == nil {
+			if constraint.Check(v) {
+				versions = append(versions, v)
+			}
+		}
+	}
+
+	if len(tags) == 0 || len(versions) == 0 {
```
matheuscscp, line 75: `if len(tags) == 0` can be moved earlier

internal/controller/resourcesetinputprovider_controller_git_test.go
```diff
@@ -0,0 +1,906 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/yaml"
+
+	fluxcdv1 "github.com/controlplaneio-fluxcd/flux-operator/api/v1"
+	"github.com/controlplaneio-fluxcd/flux-operator/internal/testutils"
+)
+
+func TestResourceSetInputProviderReconciler_GitTag_LifeCycle(t *testing.T) {
```
matheuscscp, line 29:
```suggestion
func TestResourceSetInputProviderReconciler_GitLabTag_LifeCycle(t *testing.T) {
```

Or rename the next two tests below to `...GitBranch...` and `...GitChangeRequest...`

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -292,8 +310,8 @@ The `.spec.schedule` field is optional and can be used to specify a list of `Sch
 Each `Schedule` object has the following fields:
 - `.cron`: a required string representing the cron schedule in the format accepted by
   [cron](https://crontab.guru/).
-- `.timeZone`: a string representing the time zone in which the cron schedule should be interpreted.
-  This is optional and defaults to `UTC`.
+- `.timeZone`: a string representing the [time zone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
```
matheuscscp, line 313: Nice :+1:

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -304,21 +322,29 @@ Example:
...
-one has a zero duration window, flux-operator will always reconcile the ResourceSetInputProvider upon
-any requests.
+When a schedule is specified with `window: 0s`, flux-operator will make the best effort to reconcile
+the ResourceSetInputProvider at the scheduled time; but it may not be able to guarantee that the
+reconciliation will start exactly at that time, especially if the operator is too busy or if
+the cluster is under heavy load.
+
+If multiple schedules are specified and at least one has a zero-duration window,
+flux-operator will always reconcile the ResourceSetInputProvider upon any requests, e.g.,
+updating the `.spec` or when the CLI is used to trigger a reconciliation.
```
matheuscscp:
```suggestion
updating the `.spec`, when the CLI is used to trigger a reconciliation, or when flux-operator restarts.
```

internal/gitprovider/github.go
matheuscscp: Unless you're thinking about serialization i.e. `[]` in JSON for example, in which case this change makes sense :+1:

internal/controller/resourcesetinputprovider_controller_git_test.go
stefanprodan: This test covers both GitLabTag and GitHubTag unlike the others which are for a single provider.

matheuscscp: Oh ok, I missed it! Thanks

internal/gitprovider/github.go
stefanprodan: Yes linter fails also my IDE highlights this, we shouldn't be using `array == nil` anywhere, we should always check `len(array) > 0`

matheuscscp: Yes, `array == nil` is bad, always prefer `len(array) == 0`. But I think `var x []something` is more idiomatic as it highlights the use of Go zero values. But for serializing to `[]`, yes, either `make([]something, 0)` or `[]something{}`

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -136,6 +144,7 @@ The following filters are supported:
 - `labels`: filter GitHub Pull Requests or GitLab Merge Requests by labels.
 - `includeBranch`: regular expression to include branches by name.
 - `excludeBranch`: regular expression to exclude branches by name.
+- `semver`: sematic version range to filter and order tags.
```
matheuscscp:
```suggestion
- `semver`: sematic version range to filter and sort tags.
```

PR #307, opened 2025-06-18
Introduce alphabetical and numerical ordering for tags

matheuscscp:
I thought about first introducing sorting alphabetically and numerically to make sure we iterate on the RSIP API in a consistent way. We introduced `semver` in `filter`, now it's time for these other two orders, which can be useful in practice to select latest tags as a simple mechanism.


matheuscscp:
@stefanprodan CI is failing because I need you to please push the following tags to podinfo GitHub and GitLab:

* `123456`
* `234567`
* `345678`

matheuscscp:
Closing as we decided in a different direction

Issue #151, opened 2025-01-19
`ResourceSet` dynamic inputs from OCI Repository Tags


stefanprodan:
With the introduction of [ResourceSet input providers](https://github.com/controlplaneio-fluxcd/flux-operator/pull/143), Flux Operator can dynamically deploy applications based on data provided by Git services e.g. Pull Requests or Git branches.

There are cases when the Kubernetes manifests are pushed from Git to a container registry and stored as OCI artifacts. One such example is a mono-repo with multiple microservices, where changes to a specific microservice, trigger a new OCI artifact push and Flux running in the cluster should deploy only the affected microservice.

In this scenario, the OCI repository tags could be used as inputs for the `ResourceSet` templates and thus decouple the deployment process from Git. To support this use case, a new `ResourceSetInputProvider` type will be introduced called `OCIRepositoryTag`.

## Example

We assume that the Kubernetes manifests of all microservices are stored in the same OCI repository. When a commit is pushed to the Git mono-repo, the CI pipeline determines which microservice has changed and pushes the Kubernetes manifests belonging to the that microservice using a tag that matches the format `app-<microservice name>`.

```shell
flux push artifact oci://ghcr.io/<org>/<mono-repo>:app-microservice1 --path ./microservice1
```

### Flux Operator Config

We define a `ResourceSetInputProvider` that fetches the OCI repository tags from the specified URL and filters them by a regular expression:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSetInputProvider
metadata:
  name: app-microservices-tags
  namespace: default
  annotations:
    fluxcd.controlplane.io/reconcile: "enabled"
    fluxcd.controlplane.io/reconcileEvery: "5m"
spec:
    type: OCIRepositoryTag
    url: oci://ghcr.io/<org>/<mono-repo>
    secretRef:
      name: ghcr-pull-secret
    filter:
      includeTag: "^app-.*"
```

Assuming we have three microservices in the mono-repo, the `ResourceSetInputProvider` will export the following inputs:

```yaml
status:
  exportedInputs:
    - id: "1433470881"
      digest: "sha256:d7e924b4882e55b97627355c7b3d2e711e9b54303afa2f50c25377f4df66a83b"
      tag: 'app-microservice1'
    - id: "1433536418"
      digest: "sha256:3b6cdcc7adcc9a84d3214ee1c029543789d90b5ae69debe9efa3f66e982875de"
[...]

PR #302, opened 2025-06-17
feature: Add support for Azure DevOps to `ResourceSetInputProvider`

paulvlasml:
`ResourceSetInputProvider` currently has support for the GitHub and GitLab git providers. Since the interface for the git providers only has 2 functions(now 3), I would like to contribute by adding support for Azure Dev ops. I have made the commits for this contribution in the following forked repo:

PR #310, opened 2025-06-19
Introduce OCI artifact tags for ResourceSetInputProvider

matheuscscp:
Part of: #151 

In this PR we are implementing the ability to list tags from generic OCI repositories, with or without a semver filter. For example:

```yaml
metadata:
  name: app-microservices-tags
  namespace: default
  annotations:
spec:
    type: OCIArtifactTag
    url: oci://ghcr.io/<org>/<mono-repo>
    secretRef:
      name: ghcr-pull-secret
    filter:
      semver: ">=1.0.0 <2.0.0"
      limit: 1
```

This will export only the latest tag in the semver range `>=1.0.0 <2.0.0`:

```yaml
status:
  exportedInputs:
  - id: "059e0169"
    tag: v1.3.2
    digest: sha256:639798659bf761152f5e1a0555e5d22fd3221e6801c7ce089878f562412344fe
```

Without `.spec.filter` all the tags would be exported.

### DIFF
--- a/internal/controller/resourcesetinputprovider_controller.go
+++ b/internal/controller/resourcesetinputprovider_controller.go
@@ -5,9 +5,11 @@ package controller
 
 import (
 	"context"
+	"crypto/tls"
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"net/http"
 	"regexp"
 	"slices"
 	"strings"
@@ -19,6 +21,9 @@ import (
 	"github.com/fluxcd/pkg/git/github"
 	"github.com/fluxcd/pkg/runtime/conditions"
 	"github.com/fluxcd/pkg/runtime/patch"
+	"github.com/fluxcd/pkg/version"
+	"github.com/google/go-containerregistry/pkg/authn/k8schain"
+	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/opencontainers/go-digest"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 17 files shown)


internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -219,28 +226,42 @@ func (r *ResourceSetInputProviderReconciler) reconcile(ctx context.Context,
...
-			msg := fmt.Sprintf("failed to create provider %s", err.Error())
-			conditions.MarkFalse(obj,
-				meta.ReadyCondition,
-				meta.ReconciliationFailedReason,
-				"%s", msg)
-			r.notify(ctx, obj, corev1.EventTypeWarning, meta.ReconciliationFailedReason, msg)
-			return ctrl.Result{}, err
-		}
+		switch {
+		case strings.HasPrefix(obj.Spec.Type, "Git"):
```
stefanprodan: This will break #302 

We now have 2 level deep switches, let's create a function `reconcileExternalProvider` that returns the inputs.

matheuscscp: Thanks! It made more sense to call it `callExternalProvider` instead and return just the exported inputs instead of a reconcile result. We were handling all the errors in this part the same way, so now we just return the error in `callExternalProvider` and then update the conditions and call `r.notify()` in the caller.

stefanprodan:
We could set the `GITHUB_TOKEN` env var in the test step to avoid rate limits.

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -316,19 +346,14 @@ func (r *ResourceSetInputProviderReconciler) newGitProvider(ctx context.Context,
...
-		Filters: gitprovider.Filters{
-			Limit: 100,
-		},
 	}
 
 	if obj.Spec.Filter != nil {
-		if obj.Spec.Filter.Limit > 0 {
-			opts.Filters.Limit = obj.Spec.Filter.Limit
-		}
+		opts.Filters.Limit = obj.GetFilterLimit()
```
stefanprodan: This changes the behaviour, the intention here was to always have a limit no matter if filters are set or not.

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #311, opened 2025-06-19
Introduce workload identity for ResourceSetInputProvider

matheuscscp:
Closes: #151 

TODO:
* [ ] Tests
* [ ] Docs


matheuscscp:
@stefanprodan We use `.Requeue` in tons of places:

https://github.com/controlplaneio-fluxcd/flux-operator/actions/runs/15767491421/job/44446560359?pr=311

stefanprodan:
Let's add `Requeue` to the `staticcheck` ignore rules for now, but we'll need to deal with this before controller-runtime removes it.

api/v1/resourcesetinputprovider_types.go
```diff
@@ -244,6 +260,17 @@ func (in *ResourceSetInputProvider) GetFilterLimit() int {
 	return 100 // default limit
 }
 
+// GetInvolvedObject returns the involved object for the input provider
+// for cache operations.
+func (in *ResourceSetInputProvider) GetInvolvedObject() cache.InvolvedObject {
```
stefanprodan: Please move this out of the API package, we shouldn’t have here any deps beyond Kubernetes

matheuscscp: Fixed, thanks!

cmd/operator/main.go
```diff
@@ -111,6 +112,8 @@ func main() {
 		setupLog.Info("RUNTIME_NAMESPACE env var not set, defaulting to " + fluxcdv1.DefaultNamespace)
 	}
 
+	auth.EnableObjectLevelWorkloadIdentity()
```
matheuscscp, line 116: @stefanprodan I thought it would be a good idea not to add the feature gate here since:

- The operator does not have any feature gates at the moment.
- Cluster admins would need to enable it in the FluxInstance and in the operator (two places).
- Cluster admins have other ways to block the feature if they really need it (Kyverno, admission CEL, etc.), which we can document.
- The operator is a good place for experimenting this feature.

Let me know if you prefer otherwise.

Idea: It would be nice if the operator had an API to more easily apply and manage an operator instance, in a way that is notably simpler than a HelmRelease in the GitOps repo but still allows GitOps control over the instance, including easily adding feature gates. Tying the lifecycle of an operator instance to FluxInstance is bad e.g. if you want to delete the FluxInstance from the cluster and want to keep the operator. But at the same time it could be a good avenue for a simplified UX. Maybe with an annotation and proper help from the `flux-operator` CLI we could make this possible, e.g. `flux-operator delete instance --keep-operator` first annotates the instance with `fluxcd.controlplane.io/delete: Instance` (default being `InstanceAndOperator`) and then deletes the instance (e.g. same as `flux uninstall`, completely wipes all pods). Could also be the opposite if we prefer being less destructive, i.e. `delete instance --and-operator`, annotation `delete: InstanceAndOperator`, default `Instance`. We could add the `flux-operator` component (`spec.components`), and the underlying object could still be an officially distributed (and hard-coded) HelmRelease instance template with support for `spec.cluster.objectLevelWorkloadIdentity` and `spec.kustomize.patches`, like the objects applied by `spec.sync`. We could eventually add an optional `spec.operator` with specific features if requested.

stefanprodan (commented):
In OCIRepository we use the `.spec.serviceAccountName` to load attached [image pull secrets](https://github.com/controlplaneio-fluxcd/d2-fleet/blob/0cc2ef99775fad4565ecc083155d900916309c04/tenants/apps.yaml#L55). We should do the same here for `OCIArtifactTag` instead of erroring out at admission.

matheuscscp:
> In OCIRepository we use the `.spec.serviceAccountName` to load attached [image pull secrets](https://github.com/controlplaneio-fluxcd/d2-fleet/blob/0cc2ef99775fad4565ecc083155d900916309c04/tenants/apps.yaml#L55). We should do the same here for `OCIArtifactTag` instead of erroring out at admission. 

While this may be a widespread pattern in Kubernetes, in practice it's just bad in terms of performance. We need one more lookup to get the secret name versus using something that is already in the object. This is true even in the Pod spec, which *also* allows image pull secrets to be specified directly. The service account is just holding a secret name, there's no real authn/authz going on when it's used only for holding image pull secrets. Would it be okay for us to wait on somebody asking for this feature? If somebody asks for it, removing the admission validation would not be a breaking change.

stefanprodan:
> in practice it's just bad in terms of performance. We need one more lookup to get the secret name versus using something that is already in the object.

This is not a good reason to break a well establish standard in Kubernetes. One lookup doesn't matter, we should implement this since both OCIRepo and ImageRepo supports it.

matheuscscp:
> > in practice it's just bad in terms of performance. We need one more lookup to get the secret name versus using something that is already in the object.
> 
> This is not a good reason to break a well establish standard in Kubernetes. One lookup doesn't matter, we should implement this since both OCIRepo and ImageRepo supports it.

Fixed, thanks!

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -220,16 +227,83 @@ spec:
 
 ### Authentication configuration
 
+#### Secret-less
```
stefanprodan: Let's swap the order, `Secret-based` being first. For workload ID we'll need a dedicated doc in the Flux Operator website. Once we have this we can trim down the explanation here, but this can come later on.

cmd/operator/main.go
stefanprodan: Enabling IAM role bindings for Flux Operator SA is super easy, just set `commonAnnotations` in the chart. If we'll add feature gates, these will be in the chart as well.

matheuscscp: Should we introduce feature gates in this PR?

stefanprodan: I think it's Ok to have the object-level WID enabled by default with no opt-out for now. The `serviceAccountName` acts as a gate because we didn't had this before in the API, so we don't risk like in Flux to break any existing configuration.

docs/api/v1/resourcesetinputprovider.md
matheuscscp: Fixed, thanks!

matheuscscp:
@stefanprodan Just rebased, can we merge this one?

stefanprodan:
Can you please test this out on AWS with pod identity for controller-level WID.

matheuscscp:
Output from `k get rsip,rset,ocirepo -o yaml`:

```yaml
apiVersion: v1
items:
- apiVersion: fluxcd.controlplane.io/v1
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"fluxcd.controlplane.io/v1","kind":"ResourceSetInputProvider","metadata":{"annotations":{},"name":"test","namespace":"flux-system"},"spec":{"type":"ECRArtifactTag","url":"oci://211125720409.dkr.ecr.us-east-1.amazonaws.com/oci-artifact"}}
      reconcile.fluxcd.io/forceAt: 2025-06-22 18:31:14.621660189 +0100 IST m=+0.018219804
      reconcile.fluxcd.io/requestedAt: 2025-06-22 18:31:14.621660189 +0100 IST m=+0.018219804
    creationTimestamp: "2025-06-22T16:38:18Z"
    finalizers:
    - fluxcd.controlplane.io/finalizer
    generation: 4
    name: test
    namespace: flux-system
    resourceVersion: "29218"
    uid: bc5f488f-0baf-49b1-8dc6-e20a94f89c6c
  spec:
    serviceAccountName: source-controller
    type: ECRArtifactTag
    url: oci://211125720409.dkr.ecr.us-east-1.amazonaws.com/oci-artifact
  status:
    conditions:
    - lastTransitionTime: "2025-06-22T17:41:43Z"
      message: Reconciliation finished in 496ms
      observedGeneration: 4
      status: "True"
      type: Ready
    exportedInputs:
    - digest: sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd
      id: "147587726"
      tag: latest
    lastExportedRevision: sha256:f5c55cf5863d9cb9ab63b8ae59054317d12729e4c438f1d5f13dad7a646d9dde
    lastHandledForceAt: 2025-06-22 18:31:14.621660189 +0100 IST m=+0.018219804
    lastHandledReconcileAt: 2025-06-22 18:31:14.621660189 +0100 IST m=+0.018219804
- apiVersion: fluxcd.controlplane.io/v1
  kind: ResourceSet
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"fluxcd.controlplane.io/v1","kind":"ResourceSet","metadata":{"annotations":{},"name":"test","namespace":"flux-system"},"spec":{"inputsFrom":[{"kind":"ResourceSetInputProvider","name":"test"}],"resources":[{"apiVersion":"v1","data":{"tag":"\u003c\u003c inputs.tag \u003e\u003e"},"kind":"ConfigMap","metadata":{"name":"cm-name","namespace":"flux-system"}}]}}
    creationTimestamp: "2025-06-22T17:38:30Z"
    finalizers:
    - fluxcd.controlplane.io/finalizer
    generation: 3
    name: test
    namespace: flux-system
    resourceVersion: "29375"
    uid: 78b35e5e-5b6d-4469-964f-c33897b251a2
  spec:
    inputsFrom:
      name: test
    resources:
    - apiVersion: source.toolkit.fluxcd.io/v1
      kind: OCIRepository
      metadata:
        annotations:
          event.toolkit.fluxcd.io/tag: << inputs.tag >>
        name: test
        namespace: flux-system
      spec:
        interval: 1m
        provider: aws
        ref:
          digest: << inputs.digest >>
        url: oci://211125720409.dkr.ecr.us-east-1.amazonaws.com/oci-artifact
  status:
    conditions:
    - lastTransitionTime: "2025-06-22T17:42:03Z"
      message: Reconciliation finished in 55ms
      observedGeneration: 3
      status: "True"
      type: Ready
    inventory:
      entries:
      - id: flux-system_test_source.toolkit.fluxcd.io_OCIRepository
        v: v1
    lastAppliedRevision: sha256:1f9792fc091a3d5ed77023ca014aa9a23bbe8aaaa40692ea3a1c213fc355fc16
- apiVersion: source.toolkit.fluxcd.io/v1
  kind: OCIRepository
  metadata:
    annotations:
      event.toolkit.fluxcd.io/tag: latest
    creationTimestamp: "2025-06-22T17:42:03Z"
    finalizers:
    - finalizers.fluxcd.io
    generation: 1
    labels:
      resourceset.fluxcd.controlplane.io/name: test
      resourceset.fluxcd.controlplane.io/namespace: flux-system
    name: test
    namespace: flux-system
    resourceVersion: "29396"
    uid: eea6b4e2-d7dd-4ce7-a9b9-38340384c2d8
  spec:
    interval: 1m
    provider: aws
    ref:
      digest: sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd
    timeout: 60s
    url: oci://211125720409.dkr.ecr.us-east-1.amazonaws.com/oci-artifact
  status:
    artifact:
      digest: sha256:20ac27cf403191a4e1747005122adb240c90dc639df50f1d3354d373b24760aa
      lastUpdateTime: "2025-06-22T17:42:04Z"
      metadata:
        org.opencontainers.image.created: "2025-05-05T14:44:04Z"
        org.opencontainers.image.revision: '@sha1:156f812f4d684126745b4b0197292884973e4bec'
        org.opencontainers.image.source: git@github.com:matheuscscp-test-org/test-flux-gcp-wi.git
      path: ocirepository/flux-system/test/sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd.tar.gz
      revision: sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd
      size: 1217
      url: http://source-controller.flux-system.svc.cluster.local./ocirepository/flux-system/test/sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd.tar.gz
    conditions:
    - lastTransitionTime: "2025-06-22T17:42:04Z"
      message: stored artifact for digest 'sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd'
      observedGeneration: 1
      reason: Succeeded
      status: "True"
      type: Ready
    - lastTransitionTime: "2025-06-22T17:42:04Z"
      message: stored artifact for digest 'sha256:ac771ae694eccd50f3d7a8332c0ba248a23555534e79cb2c8678d897bd1eb1fd'
      observedGeneration: 1
      reason: Succeeded
      status: "True"
      type: ArtifactInStorage
    observedGeneration: 1
    url: http://source-controller.flux-system.svc.cluster.local./ocirepository/flux-system/test/latest.tar.gz
kind: List
metadata:
  resourceVersion: ""
```

matheuscscp:
I also tested with EKS Pod Identity for the operator pod and it also works.

PR #312, opened 2025-06-20
cli: Add `flux-operator reconcile resource` command

stefanprodan:
This PR adds a new command for reconciling Flux custom resources:

```
Usage:
  flux-operator reconcile resource [flags]

Examples:
  # Trigger the reconciliation of a Flux Kustomization
  flux-operator -n apps reconcile resource Kustomization/my-app

  # Force reconcile a Flux HelmRelease
  flux-operator -n apps reconcile resource HelmRelease/my-release --force

  # Trigger the reconciliation of an OCIRepository without waiting for it to become ready
  flux-operator -n apps reconcile resource OCIRepository/my-app --wait=false


Flags:
      --force   Force the reconciliation of the resource, applies only to Flux HelmReleases.
  -h, --help    help for resource
      --wait    Wait for the resource to become ready. (default true)
```

### DIFF
--- a/cmd/cli/completion.go
+++ b/cmd/cli/completion.go
@@ -76,6 +76,33 @@ func resourceNamesCompletionFunc(gvk schema.GroupVersionKind) func(cmd *cobra.Co
 	}
 }
 
+// resourceKindCompletionFunc returns a function that can be used as a completion function for Flux resource kinds.
+func resourceKindCompletionFunc() func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		// These are the Flux kinds that react to the reconcileAt annotation.
+		kinds := []string{
+			"GitRepository/",
+			"OCIRepository/",
+			"Bucket/",
+			"HelmRepository/",
+			"HelmChart/",
+			"HelmRelease/",
+			"Kustomization/",
+			"ImageRepository/",
+			"Receiver/",
+		}
+
+		var comps []string
+		for _, kind := range kinds {
+			if strings.HasPrefix(kind, toComplete) {
+				comps = append(comps, kind)
+			}
+		}
+
+		return comps, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
 // completionError is a helper function to handle errors in completion functions.
 func completionError(err error) ([]string, cobra.ShellCompDirective) {
 	cobra.CompError(err.Error())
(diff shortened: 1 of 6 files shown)


cmd/cli/completion.go
```diff
@@ -76,6 +76,31 @@ func resourceNamesCompletionFunc(gvk schema.GroupVersionKind) func(cmd *cobra.Co
 	}
 }
 
+// resourceKindCompletionFunc returns a function that can be used as a completion function for Flux resource kinds.
+func resourceKindCompletionFunc() func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		kinds := []string{
```
matheuscscp, line 83: Is this list only supposed to have suspendable objects? If yes, I'd add this to the function comment. If not, I'd add the remaining Flux kinds to the list.

cmd/cli/reconcile_inputprovider.go
matheuscscp, line 1: Nice improvement in this file :+1:

cmd/cli/reconcile_resource.go
```diff
@@ -0,0 +1,102 @@
...
+  # Trigger the reconciliation of an OCIRepository without waiting for it to become ready
+  flux-operator -n apps reconcile resource OCIRepository/my-app --wait=false
+`,
+	Args:              cobra.ExactArgs(1),
+	RunE:              reconcileResourceCmdRun,
+	ValidArgsFunction: resourceKindCompletionFunc(),
+}
+
+type reconcileResourceFlags struct {
+	force bool
```
matheuscscp, line 34: `HelmRelease` also supports the `--reset` flag in `flux reconcile hr` (I think it's the only exception missing)

stefanprodan: Force implies reset, I don't think anyone uses reset, we'll add it if someone asks for it.

cmd/cli/reconcile_resource.go
```diff
@@ -0,0 +1,102 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), rootArgs.timeout)
+	defer cancel()
+
+	now := metav1.Now().String()
+	annotations := map[string]string{
+		meta.ReconcileRequestAnnotation: now,
+	}
+
+	if reconcileeResourceArgs.force {
```
matheuscscp, line 73: I'd also put a check here to error out if the kind is not `HelmRelease`

stefanprodan: I thought about this, and in all other cases, what force does for a HelmRelease happens for all others, so no point in erroring out.

PR #315, opened 2025-06-22
Set operator version information

stefanprodan:
This PR adds the Flux Operator version info:

- Set version at build time
- Add version info to `FluxReport.spec.operator`
- Display client and server version in the CLI with `flux-operator version`


matheuscscp (approved):
LGTM

PR #316, opened 2025-06-22
Add ResourceSet stats to `FluxReport.spec.reconcilers`

stefanprodan:
This PR adds `ResourceSet` and `ResourceSetInputProvider` reconcile statistics to the `FluxReport.spec.reconcilers`.


matheuscscp (approved):
LGTM

PR #318, opened 2025-06-22
cli: Add `flux-operator suspend/resume resource` commands

stefanprodan:
This PR adds commands for suspending and resuming Flux resources:

- `flux-operator suspend resource <kind>/<name> -n <namespace>`
- `flux-operator resume resource <kind>/<name> -n <namespace>`

In addition, the `flux-operator resume` commands will wait for a resumed resource to become ready and will print the reconciliation result. Can be disabled with `--wait=false`.


matheuscscp (approved):
LGTM

PR #319, opened 2025-06-22
Introduce regex and latest filters for tags in ResourceSetInputProvider

matheuscscp:
This PR introduces the fields `.spec.filter.value` and `.spec.filter.groupKey`, which can be set alongside `.spec.filter.includeTag`. They are templates that can be expanded using the `.includeTag` regex matching results into a value and a group key from a tag name for usage with other filters. The `.value` can be used with `.semver` and `.latestPolicy` (see next), and `.groupKey` can be used only with `.latestPolicy` for now.

Introduces also `.spec.filter.latestPolicy`, which can be one of `SemVer`, `Alphabetical`, `ReverseAlphabetical`, `Numerical` and `ReverseNumerical`. This filter uses the expanded group keys to group tags together, then applies the specified "latest policy" to each group individually using the expanded values. When `.groupKey` is not specified, all tags are considered to belong to a single group, and only a single latest tag will be exported. When `.value` is not specified, the value used for comparison is the tag itself.

The `.spec.filter.limit` is ignored when `.spec.filter.latestPolicy` is set.

Examples:

```yaml
spec:
  filter:
    includeTag: ^([a-z]+)-([a-z]+)-(v.+)$
    groupKey: $1-$2
    value: $3
    semver: ">=1.0.0 <2.0.0" # uses the expanded template from .value
    latestPolicy: SemVer     # uses both .value and .groupKey
```

```yaml
spec:
  filter:
    includeTag: ^([a-z]+)-([a-z]+)-(.+)$
    groupKey: $1-$2
    value: $3
    latestPolicy: Alphabetical # uses both .value and .groupKey
```

### DIFF
--- a/internal/filtering/filters.go
+++ b/internal/filtering/filters.go
@@ -1,0 +1,310 @@
+// Copyright 2025 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package filtering
+
+import (
+	"cmp"
+	"regexp"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/fluxcd/pkg/version"
+
+	fluxcdv1 "github.com/controlplaneio-fluxcd/flux-operator/api/v1"
+)
+
+// DefaultLimit is the default limit for the number of results returned by the filters.
+const DefaultLimit = 100
+
+// Filters holds the filters for the input provider responses.
+type Filters struct {
+	// Labels is used for a "match labels" filter.
+	Labels []string
+
+	// Include is used for including tags or branches.
+	// When set together with Value or GroupKey, can be used
+	// to expand a value or group key from the branch or tag
+	// name for usage with another filter.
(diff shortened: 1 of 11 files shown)


internal/filtering/filters.go
```diff
@@ -0,0 +1,141 @@
...
+		// Recover the original tags for each extracted value
+		// in the sorted order of semver, but keeping the original
+		// order of the tags with the same extracted value.
+		if mustExtract {
+			filtered = nil
+			for _, extract := range semverList {
+				filtered = append(filtered, tagsForExtract[extract]...)
+			}
+		}
+	}
```
stefanprodan, line 164:
```suggestion
	} else {
	  // sort calver
	  slices.Sort(filtered)
	  slices.Reverse(filtered)
	}
```

When semver is not specified we can assume calver / unix timestamps, tested here https://go.dev/play/p/s-556GNp4HV

matheuscscp:
Closing in favor of #320

PR #320, opened 2025-06-23
Introduce regex filters for tags in ResourceSetInputProvider

matheuscscp:
Closes: #151 

This PR introduces the fields `.spec.filter.includeTag` and `.spec.filter.excludeTag`. The fields are regex patterns that are applied to the listed tags for eliminating the undesired ones from the RSIP exported inputs, similar to `.spec.filter.includeBranch` and `.spec.filter.excludeBranch`.

Additionally, this PR is also now sorting the tags in reverse alphabetical order if `.spec.filter.semver` is not specified.


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #327, opened 2025-06-28
Fix GitHub Enterprise Server authentication in ResourceSetInputProvider

### LINKED ISSUE #324: Unable to use ResourceSetInputProvider with GitHub Enterprise Server
I'm running into some difficulty in getting ResourceSetInputProvider to work on our self-hosted GitHub Enterprise Server instance.

I've set up both a PAT and App secret, using:

```shell
flux -n flux-system create secret git git-token-auth \
  --url=https://git.instance.tld/orgname/reponame \
  --username=ghes-user \
  --password=<token>


flux -n flux-system create secret githubapp git-app-auth \
  --app-id=1234 \
  --app-installation-id=5678 \
  --app-private-key=/path/to/private-key.pem \
  --app-base-url=https://git.instance.tld/api/v3/
```

And defined the ResourceSetInputProvider:

```yaml
---
metadata:
  name: reponame-pull-requests
  namespace: flux-system
  annotations:
spec:
  type: GitHubPullRequest
  url: https://git.instance.tld/orgname/reponame
  secretRef:
    name: git-token-auth // also tried git-app-auth
  filter:
    labels:
      - "deploy/flux-preview"
```

And receive reconciliation errors using either:
`{"level":"error","ts":"2025-06-27T02:49:09.251Z","msg":"Reconciler error","controller":"resourcesetinputprovider","controllerGroup":"fluxcd.controlplane.io","controllerKind":"ResourceSetInputProvider","ResourceSetInputProvider":{"name":"reponame-pull-requests","namespace":"flux-system"},"namespace":"flux-system","name":"reponame-pull-requests","reconcileID":"fa3c7fcf-721c-46e3-94be-9f51cb003734","error":"failed to list requests: could not list pull requests: GET https://git.instance.tld/api/v3/repos/orgname/reponame/pulls?per_page=100&state=open: 401 Must authenticate to access this API. []"}`

Also to note, I tried the same ResourceSetInputProvider with one of our private github.com repos and using a github.com token without an issue.

I can `curl https://git.instance.tld/api/v3/repos/orgname/reponame/pulls --user ghes-user:token` successfully. The token is good. `git-token-auth` has data properties username and password, and `git-app-auth` has `githubAppBaseURL`, `githubAppID`, `githubAppInstallationID`, and `githubAppPrivateKey`. The same git-token-auth and git-app-auth secret are also used for Flux GitRepository resources and are working as intended against `https://git.instance.tld`.

matheuscscp:
Fixes #324 

As reported in the issue, authentication for GitHub Enterprise Server does not work without specifying a CA. This PR fixes the bug.

I was able to spin up a free trial of GitHub Enterprise Server and test a few different combinations:

* I was able to reproduce the issue with a valid Let's Encrypt cert without specifying the CA and a classic PAT. I got the same error message reported in the issue.

After introducing the fix:

* I was able to reconcile an RSIP with a valid Let's Encrypt cert without specifying the CA and a classic PAT. (**Fixed!**)
* I was able to reconcile an RSIP with the built-in self-signed cert specified as the CA and a classic PAT.
* I was able to reconcile an RSIP with a valid Let's Encrypt cert specifying the CA chain and a classic PAT.

### DIFF
--- a/internal/gitprovider/github.go
+++ b/internal/gitprovider/github.go
@@ -50,6 +50,9 @@ func NewGitHubProvider(ctx context.Context, opts Options) (*GitHubProvider, erro
 			}
 			ctxCA := context.WithValue(ctx, oauth2.HTTPClient, &http.Client{Transport: tr})
 			httpClient = oauth2.NewClient(ctxCA, ts)
+		} else {
+			// Create OAuth2 client without custom cert pool
+			httpClient = oauth2.NewClient(ctx, ts)
 		}
 		client, err = github.NewClient(httpClient).WithEnterpriseURLs(host, host)
 		if err != nil {


stefanprodan (approved):
LGTM

Thanks @synack-badamson and @matheuscscp

PR #328, opened 2025-06-28
cli: Add `flux-operator export` commands

stefanprodan:
The `flux-operator export` commands are used to export the Flux Operator resources in YAML format.
The exported resources can be used for backup, migration, or inspection purposes.

The following commands are available:

- `flux-operator export report`: Exports the FluxReport resource containing the distribution status and version information.
- `flux-operator export resource <kind>/<name> -n <namespace>`: Exports a Flux resource from the specified namespace.


matheuscscp (approved):
LGTM (nice comments from the linter!)

PR #341, opened 2025-07-11
Update dependencies to Kubernetes 1.33.2

stefanprodan:
Update deps and adapt the secrets loading to the new `runtime/secrets` version.


internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -267,9 +267,16 @@ func (r *ResourceSetInputProviderReconciler) callExternalProvider(
 	// Get the TLS config.
 	var tlsConfig *tls.Config
 	if obj.Spec.CertSecretRef != nil {
+		key := types.NamespacedName{
+			Name:      obj.Spec.CertSecretRef.Name,
+			Namespace: obj.GetNamespace(),
+		}
+		var secret corev1.Secret
+		if err := r.Client.Get(ctx, key, &secret); err != nil {
```
matheuscscp: Let's wait for Shu to reintroduce the variant that fetches the secret for this one

matheuscscp: `runtime/v0.66.0` is released, let's use `secrets.TLSConfigFromSecretRef`

Issue #342, opened 2025-07-11
Bug: `flux-operator resume instance flux` times out


matheuscscp:
Steps to reproduce:

```
flux-operator resume instance flux
◎ Waiting for reconciliation...
✗ context deadline exceeded
```

However:

```
k get fluxinstance flux -w
NAME   AGE   READY     STATUS                       REVISION
flux   15d   Unknown   Reconciliation in progress   v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d

flux   15d   True      Reconciliation finished in 28s   v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d
flux   15d   True      Reconciliation finished in 28s   v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d
flux   15d   Unknown   Reconciliation in progress       v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d
flux   15d   True      Reconciliation finished in 2s    v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d
flux   15d   True      Reconciliation finished in 2s    v2.6.4@sha256:af215ebdfdf6940eaad29e027c26f7ab81d515aec879f0cad20c3a2f2a02158d
```

stefanprodan:
Hmm strange, that command should result in an error:

```console
$ flux-operator resume instance flux
✗ unable to read FluxInstance/default/flux error: fluxinstances.fluxcd.controlplane.io "flux" not found
```

This should work:

```console
$ flux-operator -n flux-system suspend instance flux
✔ Reconciliation suspended
$ flux-operator -n flux-system resume instance flux
◎ Waiting for reconciliation...
✔ Reconciliation finished in 2s
```

Are you running the latest CLI?

```console
$ flux-operator version
client: 0.24.1
server: v0.24.1
distribution: v2.6.4
```

stefanprodan:
Wait do you have the FluxInstance and Flux Operator in the default namespace?

matheuscscp:
From `k get fluxinstance flux -o yaml | yq .metadata`:

```yaml
annotations:
  fluxcd.controlplane.io/reconcile: enabled
[...]

PR #343, opened 2025-07-12
cli: Fix readiness check

stefanprodan:
fix: #342


matheuscscp (approved):

stefanprodan:
I still can't reproduce this bug, but if this doesn't fix it, then we have an issue in the controller with setting `lastHandledReconcileAt`. The only way it could get stuck after this PR, is when `.status.lastHandledReconcileAt` would not match the annotation value set by the CLI.

PR #351, opened 2025-07-17
cli: Add `--inputs-from-provider` flag to `flux-operator build rset` command 

stefanprodan:
Build a ResourceSet by providing the inputs from a static ResourceSetInputProvider:

```
  flux-operator build resourceset -f my-resourceset.yaml \
    --inputs-from-provider my-resourcesetinputprovider.yaml
```


matheuscscp:
The flag could be just `--inputs-from`

Edit: Ah right, we already have this flag

matheuscscp (approved):

PR #358, opened 2025-07-22
Unify Flux kinds and groups under the API package

stefanprodan:
Define the Flux APIs Kind and Group constants in `/api` and replace all hardcoded values the the project.


api/v1/common_types.go
```diff
@@ -75,3 +76,56 @@ type CommonMetadata struct {
...
+	FluxHelmRepositoryKind        = "HelmRepository"
+	FluxHelmChartKind             = "HelmChart"
+	FluxNotificationGroup         = "notification.toolkit.fluxcd.io"
+	FluxAlertKind                 = "Alert"
+	FluxAlertProviderKind         = "Provider"
+	FluxReceiverKind              = "Receiver"
+	FluxImageGroup                = "image.toolkit.fluxcd.io"
+	FluxImageRepositoryKind       = "ImageRepository"
+	FluxImagePolicyKind           = "ImagePolicy"
+	FluxImageUpdateAutomationKind = "ImageUpdateAutomation"
```
matheuscscp: I'd separate these in either two blocks, groups and kinds, or 5 blocks, one per group

PR #366, opened 2025-07-24
controller: Fix instance health check on subsequent reconciliations

stefanprodan:
Ensure that on subsequent reconciliations of a `FluxInstance`, the health check is performed on all Flux resources. 

In addition, this PR improves the error reporting by adding the `sync` ready message to the `FluxInstance` failure events and status.

### DIFF
--- a/internal/controller/resourceset_controller.go
+++ b/internal/controller/resourceset_controller.go
@@ -556,7 +556,7 @@ func (r *ResourceSetReconciler) apply(ctx context.Context,
 			Timeout:  obj.GetTimeout(),
 			FailFast: true,
 		}); err != nil {
-			readyStatus := r.aggregateNotReadyStatus(ctx, kubeClient, objects)
+			readyStatus := aggregateNotReadyStatus(ctx, kubeClient, objects)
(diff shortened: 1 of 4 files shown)


matheuscscp (approved):
LGTM! 🚀

internal/controller/resourceset_controller.go
```diff
@@ -456,23 +456,21 @@ func (r *ResourceSetReconciler) apply(ctx context.Context,
...
-				OperationType: metav1.ManagedFieldsOperationApply,
-			},
-			{
-				Name:          "kustomize-controller",
+				// to undo changes made with 'kubectl apply --server-side --force-conflicts'
+				Name:          "kubectl",
 				OperationType: metav1.ManagedFieldsOperationApply,
 			},
 			{
-				Name:          "helm",
```
matheuscscp, line 471: We won't undo changes from helm and kustomize-controller anymore?

stefanprodan: This breaks both controllers as their finalizers are removed by this.

PR #367, opened 2025-07-26
Introduce label selector for watching ConfigMaps and Secrets

matheuscscp:
Trigger the reconciliation of a ResourceSet when a ConfigMap or Secret that is copied by the ResourceSet changes.

xref: https://github.com/fluxcd/flux2/issues/5446


cmd/operator/main.go
```diff
@@ -91,6 +93,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.CommandLine.StringVar(&watchOptions.ConfigsLabelSelector, "watch-configs-label-selector", meta.LabelKeyWatch+"="+meta.LabelValueWatchEnabled,
```
matheuscscp, line 96: Binding only the supported flag for now (not calling `watchOptions.BindFlags()`).

matheuscscp:
Fixed, thanks!

PR #370, opened 2025-07-28
controller: Improve resource ownership takeover

stefanprodan:
Avoid conflicts with Flux controllers when taking ownership of managed resources.

### DIFF
--- a/internal/controller/fluxinstance_controller.go
+++ b/internal/controller/fluxinstance_controller.go
@@ -440,28 +440,7 @@ func (r *FluxInstanceReconciler) apply(ctx context.Context,
 		},
 		// Take ownership of the Flux resources if they
 		// were previously managed by other tools.
-		FieldManagers: []ssa.FieldManager{
-			{
-				Name:          "flux",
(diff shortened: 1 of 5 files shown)


internal/controller/fluxinstance_controller.go
```diff
@@ -440,28 +440,7 @@ func (r *FluxInstanceReconciler) apply(ctx context.Context,
...
-			{
-				Name:          "kubectl",
-				OperationType: metav1.ManagedFieldsOperationUpdate,
-			},
-			{
-				Name:          "flux-controller",
-				OperationType: metav1.ManagedFieldsOperationApply,
-			},
-		},
+		FieldManagers: takeOwnershipFrom([]string{"flux"}),
```
matheuscscp, line 443: Curious about this separation. What facts about this list passed in the caller and about the defaults imply that it makes sense to separate them?

matheuscscp: Ah ok, it makes sense for FluxInstance to take over `flux` CLI, but not RSET. :+1:

PR #375, opened 2025-07-31
reporter: Improve metrics collector

stefanprodan:
This PR improves the reporter by exporting new metrics only after collecting all the stats from the cluster. 

In addition, the `flux_resourceset_info` gains a new label called `resources` that tracks the the number of resources generated by the ResourceSet.


matheuscscp (approved):
LGTM

Issue #372, opened 2025-07-29
Add DetailedMergeStatus to exported inputs of GitlabMergeRequest


Eitan1112:
I would be very happy and would gladly contribute a PR to add DetailedMergeStatus to the exported inputs of a ResourceSetInputProvider with type GitlabMergeRequest.

This will answer 2 use cases:
1. Quote [from the docs](https://fluxcd.control-plane.io/operator/resourcesets/github-pull-requests/#delay-updates-if-the-build-takes-too-long):
There's still a chance that Flux Operator will reconcile your ResourceSetInputProvider between the moment when you make a Git push and the moment when the GitHub Workflow adds the pause label, this will cause your ephemeral environment to be updated with the new SHA before the artifacts are built, but this is unlikely to happen if your GitHub Workflows are quickly scheduled on runners.

I know I am attaching the github docs, but the same principal also applied to the gitlab merge requests flow.

This will solve it because it will be possible to skip based on pipeline status, e.g. skip if pipeline is in running state. 

2. Run integration tests using the operator - I would like to achieve this flow:
Dev create MR -> flux operator spin up an environment -> run some tests as helm hooks or somehow else -> Report status check back to gitlab -> delete env. This will allow me to hook to the detailed_merge_status, and skip if the status is not "status_checks_must_pass".

In general, the more info exported from the providers the better IMO, it will allow to make more precise operations using the operator. I saw in #215 that it was discussed if adding exported inputs to one git provider requires adding to all off them, and while feature parity is important, I truly think adding the custom provider metadata will help tremendously.

If you'd like to see this change, as said I'll be happy to contribute a PR, but wanted to make sure it was discussed first.

Thanks a lot!
Eitan

matheuscscp:
If you're asking for a new exported input I assume you plan on using it in the `ResourceSet` templates. Can you share a snippet of how are you planning to do that?

matheuscscp:
I was thinking about this race problem and thought about the following solution.

```yaml
spec:
  skip:
    toggle: deploy/flux-preview
```

[...]

PR #377, opened 2025-08-01
RSIP: Introduce reliable skip mechanism for updating commits in change requests

matheuscscp:
This PR implements what's described [here](https://github.com/controlplaneio-fluxcd/flux-operator/issues/372#issuecomment-3143631023).

I'll wait for a review before adding docs/tests.

After publishing the artifacts, the CI can do this:

```bash
if [ "$(gh pr view --json labels --jq '.labels | any(.name == "deploy/flux-operator-togglel")')" = "true" ]; then
    gh pr edit 23 --remove-label "deploy/flux-operator-toggle"
else
    gh pr edit 23 --add-label "deploy/flux-operator-toggle"
fi
```

This can be wrapped in a GitHub Action in the https://github.com/controlplaneio-fluxcd/distribution repo called `toggle-rsip-label` with the following inputs:

```yaml
  - uses: controlplaneio-fluxcd/distribution/actions/toggle-rsip-label@main
    with:
      toggle: deploy/flux-operator-toggle
```

### DIFF
--- a/internal/controller/resourcesetinputprovider_controller.go
+++ b/internal/controller/resourcesetinputprovider_controller.go
@@ -425,7 +425,7 @@ func (r *ResourceSetInputProviderReconciler) makeFilters(
 
 // restoreSkippedGitProviderResults skip git provider results when matches skip condition.
 func (r *ResourceSetInputProviderReconciler) restoreSkippedGitProviderResults(results []gitprovider.Result, obj *fluxcdv1.ResourceSetInputProvider) ([]gitprovider.Result, error) {
-	if obj.Spec.Skip == nil || len(obj.Spec.Skip.Labels) == 0 {
+	if obj.Spec.Skip == nil || (len(obj.Spec.Skip.Labels) == 0 && obj.Spec.Skip.Toggle == "") {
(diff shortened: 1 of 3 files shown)


internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -449,14 +455,30 @@ func (r *ResourceSetInputProviderReconciler) restoreSkippedGitProviderResults(re
...
 		}
 
-		if isSkipped {
-			var exportedInput map[string]any
-			for _, ei := range exportedInputs {
-				if ei["id"] == result.ID {
-					exportedInput = ei
-					break
+		// Check .spec.skip.toggle.
+		if !isSkipped && obj.Spec.Skip.Toggle != "" {
```
stefanprodan, line 459: Please add tests for this new logic.

stefanprodan (commented):
Can you please rename the PR and commit to fit into the 80 limit.

matheuscscp:
As discussed, this idea can leave the RSIP stuck in several scenarios. It needs more thought.

PR #380, opened 2025-08-03
operator: Introduce support for helm-controller object level workload identity

matheuscscp:
Flux v2.7 introduces object-level workload support for helm-controller for kubeconfig configuration (`.spec.kubeConfig.configMapRef`). This PR includes helm-controller in the list of controller Deployments that need to receive the feature gate patch when the version is `>= 2.7.0`.


matheuscscp:
I added a commit to fix the feature gate not being rendered when disabled. This is important for keeping the `FluxInstance` API consistent in case this feature gate becomes opt-out in CNCF Flux in the future.

internal/builder/templates.go
```diff
@@ -140,14 +140,14 @@ patches:
       path: /spec/template/spec/containers/0/args/-
       value: --watch-label-selector=!{{.ShardingKey}}
 {{- end }}
-{{- if and .SupportsObjectLevelWorkloadIdentity .EnableObjectLevelWorkloadIdentity }}
+{{- if .SupportsObjectLevelWorkloadIdentity }}
```
stefanprodan, line 143: What's the reason for generating this patch if the gate is set to its default value `false`?

matheuscscp: > I added a commit to fix the feature gate not being rendered when disabled. This is important for keeping the `FluxInstance` API consistent in case this feature gate becomes opt-out in CNCF Flux in the future.

PR #381, opened 2025-08-04
Implement vertical scaling based on cluster size

stefanprodan:
This PR adds an optional field called `.spec.cluster.size` to the FluxInstance API.

The size is used to determine the vertical scaling profile for the Flux controllers based on the number of applications and the deployment frequency.

The supported values are `small`, `medium` and `large` which correspond to the following configuration for Flux `kustomize-controller` and `helm-controller`:

| Size       | Concurrency | CPU limit | Memory limit | Requeue deps |
|------------|-------------|-----------|--------------|--------------|
| **small**  | 5           | 1000m     | 512Mi        | 10s          |
| **medium** | 10          | 2000m     | 1Gi          | 5s           |
| **large**  | 20          | 3000m     | 3Gi          | 5s           |

For all sizes, the `source-controller` is configured to cache artifacts and the concurrency is set between 2 and 10.

The `small` size is suitable for edge clusters with limited resources (Raspberry Pi) or for clusters with tens of apps and moderate deployment frequency.

The `medium` size is suitable for clusters with hundreds of apps, while the `large` size is for clusters with up to a thousand apps and high deployment frequency.

### DIFF
--- a/docs/api/v1/fluxinstance.md
+++ b/docs/api/v1/fluxinstance.md
@@ -28,7 +28,7 @@ metadata:
   annotations:
     fluxcd.controlplane.io/reconcile: "enabled"
     fluxcd.controlplane.io/reconcileEvery: "1h"
-    fluxcd.controlplane.io/reconcileTimeout: "3m"
+    fluxcd.controlplane.io/reconcileTimeout: "5m"
 spec:
   distribution:
     version: "2.x"
@@ -42,6 +42,7 @@ spec:
     - image-automation-controller
   cluster:
     type: kubernetes
+    size: medium
     multitenant: false
     networkPolicy: true
     domain: "cluster.local"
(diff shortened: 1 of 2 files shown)


stefanprodan:
PS. Later on, we may want to implement `size: dynamic` where Flux Operator will automatically scale the controllers based on the metrics generated for FluxReport.

docs/api/v1/fluxinstance.md
```diff
@@ -45,6 +45,8 @@ spec:
     multitenant: false
     networkPolicy: true
     domain: "cluster.local"
+  profile:
+    type: "medium"
```
matheuscscp: Is this outdated?

matheuscscp (approved):
LGTM! 🚀

PR #388, opened 2025-08-05
cli: Handling multiple ResourceSetInputProvider in `build rset` command

### LINKED ISSUE #387: inputs-from-provider handling multiple ResourceSetInputProvider
The current implementation of `inputs-from-provider` only reads the first `ResourceSetInputProvider` from a file, instead of handling an array of providers. This makes it impossible to test examples like the *[Decoupled example*](https://fluxcd.control-plane.io/operator/resourcesets/app-definition/)* shown in the documentation using `flux-operator build rset -f examples/decoupled-rset.yaml --inputs-from-provider examples/decoupled-rsip-1.yaml `, or to perform a dry run with labeled resources.

bugbuilder:
this PR closes #387

### DIFF
--- a/cmd/cli/build_resourceset.go
+++ b/cmd/cli/build_resourceset.go
@@ -15,12 +15,14 @@ import (
 	ssautil "github.com/fluxcd/pkg/ssa/utils"
 	"github.com/spf13/cobra"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/json"
 	"sigs.k8s.io/yaml"
 
 	fluxcdv1 "github.com/controlplaneio-fluxcd/flux-operator/api/v1"
 	"github.com/controlplaneio-fluxcd/flux-operator/internal/builder"
 	"github.com/controlplaneio-fluxcd/flux-operator/internal/inputs"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var buildResourceSetCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


cmd/cli/build_resourceset.go
```diff
@@ -114,36 +116,44 @@ func buildResourceSetCmdRun(cmd *cobra.Command, args []string) error {
...
-		if err := yaml.Unmarshal(inputsData, &provider); err != nil {
-			return fmt.Errorf("error parsing inputs provider file: %w", err)
-		}
+		var filteredProviders []*fluxcdv1.ResourceSetInputProvider
 
-		if provider.Spec.Type != fluxcdv1.InputProviderStatic {
-			return fmt.Errorf("unsupported provider type '%s', only '%s' is supported",
-				provider.Spec.Type,
-				fluxcdv1.InputProviderStatic)
-		}
```
stefanprodan: This validation needs to be added back.

cmd/cli/build_resourceset.go
```diff
@@ -208,3 +218,27 @@ func saveReaderToFile(reader io.Reader) (string, error) {
...
+	var providers []*fluxcdv1.ResourceSetInputProvider
+
+	for _, doc := range documents {
+		if len(bytes.TrimSpace(doc)) == 0 {
+			continue
+		}
+
+		var provider fluxcdv1.ResourceSetInputProvider
+		if err := yaml.Unmarshal(doc, &provider); err != nil {
+			continue
```
stefanprodan: We can't just skip errors, the build command role is to validate both the RSET and the RSIP. One way to solve this is to check if the current doc contains the string `fluxcdv1.ResourceSetInputProviderKind` and skip it if not. If there is a match, then we should return the Unmarshal error.

cmd/cli/build_resourceset.go
```diff
@@ -114,36 +116,44 @@ func buildResourceSetCmdRun(cmd *cobra.Command, args []string) error {
...
+				if err != nil {
+					return fmt.Errorf("failed to parse selector: %w", err)
+				}
 
-		// copy defaultValues from the provider to the ResourceSet inputs
-		if len(provider.Spec.DefaultValues) > 0 {
-			if rset.Spec.Inputs == nil {
-				rset.Spec.Inputs = []fluxcdv1.ResourceSetInput{}
+				for _, provider := range providers {
+					if selector.Matches(labels.Set(provider.Labels)) {
```
stefanprodan: Besides labels, we need to handle here refs by name.

cmd/cli/build_resourceset.go
bugbuilder: already added

cmd/cli/build_resourceset.go
bugbuilder: already addedd

cmd/cli/build_resourceset.go
bugbuilder: already added

stefanprodan:
@bugbuilder thanks for this contribution. Can you please squash all commits into one, rename the commit as the PR name, signoff on the commit and force push. To signoff you can do `git commit -s`.

bugbuilder:
> @bugbuilder thanks for this contribution. Can you please squash all commits into one, rename the commit as the PR name, signoff on the commit and force push. To signoff you can do `git commit -s`.

You're welcome! I’ve squashed the commits.

stefanprodan (approved):
LGTM

Thanks @bugbuilder 🏅

PR #391, opened 2025-08-07
Prevent kustomize-controller from pruning the instance namespace

stefanprodan:
Prevent mistakes where users handoff the `flux-system` namespace management to some Flux Kustomization that can result in catastrophic failures with Flux deleting it's own namespace.  

xref: https://github.com/fluxcd/flux2/discussions/5472


matheuscscp (approved):
LGTM! 🚀

PR #392, opened 2025-08-10
Implement reconcile history for ResourceSets

stefanprodan:
With `.status.history` the operator tracks the reconciliation attempts over time, providing insights
into the ResourceSet's behavior which can be used for audit, anomaly detection and debugging purposes.

The history is stored as a list of snapshots, ordered by last reconciliation time. Each snapshot contains:

- `digest`: A SHA256 digest that uniquely identifies the set of generated resources being reconciled
- `firstReconciled`: The timestamp when this particular configuration was first reconciled
- `lastReconciled`: The timestamp of the most recent reconciliation attempt for this configuration
- `lastReconciledDuration`: How long the most recent reconciliation attempt took
- `lastReconciledStatus`: The status of the most recent reconciliation (e.g., `ReconciliationSucceeded`, `BuildFailed`, `ReconciliationFailed`)
- `totalReconciliations`: The total number of reconciliations for this configuration
- `metadata`: Additional information about the reconciliation, including the number of generated resources and inputs processed

The operator deduplicates entries based on the digest and status.
The history is automatically truncated to keep only the 5 most recent entries.

Example:

```yaml
status:
  history:
    - digest: sha256:43ad78c94b2655429d84f21488f29d7cca9cd45b7f54d2b27e16bbec8eff9228
      firstReconciled: "2025-07-15T10:11:00Z"
      lastReconciled: "2025-07-15T14:30:00Z"
      lastReconciledDuration: 2.818583s
      lastReconciledStatus: ReconciliationSucceeded
      totalReconciliations: 5
      metadata:
        inputs: "2"
        resources: "4"
    - digest: sha256:ec8dbfe61777b65001190260cf873ffe454451bd2e464bd6f9a154cffcdcd7e5
      firstReconciled: "2025-07-14T13:10:00Z"
      lastReconciled: "2025-07-15T10:10:00Z"
      lastReconciledDuration: 4.813292s
      lastReconciledStatus: ReconciliationFailed
      totalReconciliations: 120
      metadata:
        inputs: "1"
        resources: "2"
```


api/v1/history_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// +required
+	LastReconciledStatus string `json:"lastReconciledStatus"`
+
+	// TotalReconciliations is the total number of reconciliations that have occurred for this snapshot.
+	// + required
+	TotalReconciliations int64 `json:"totalReconciliations"`
+
+	// Metadata contains additional information about the snapshot.
+	// +optional
+	Metadata map[string]string `json:"metadata,omitempty"`
```
matheuscscp, line 123: Is there a good reason for using `map[string]string` here? Do we expect the content of this field to change over time? From what I can see, the same fields are passed everywhere and both are numbers, we could just have a `SnapshotMetadata` struct for that. I find it really bad design to use strings for representing numbers, unless we have a strong reason for this that I don't have the context for.

stefanprodan: The History is not specific to RSET, the FluxInstance will store here the Flux version, RSIP will store commit sha/digest, etc. This type matches the Kubernetes meta annotations, it's for storing controller-specific info without leaking all these details in the History struct which is meant to be generic.

api/v1/history_types.go
```diff
@@ -0,0 +1,137 @@
...
+// History represents a collection of snapshots that tracks the reconciliation
+// history of a group of resources, automatically sorted by last reconciled timestamp.
+type History []Snapshot
+
+// Len returns the length of the history slice.
+func (h History) Len() int { return len(h) }
+
+// Less reports whether the element with index i should sort before the element with index j.
+// Sorts by LastReconciled in descending order (most recent first).
+func (h History) Less(i, j int) bool {
```
matheuscscp, line 28: Where do we plan to sort the `History`? This PR is not sorting it anywhere

api/v1/history_types.go
```diff
@@ -0,0 +1,137 @@
...
+func (h History) Latest() *Snapshot {
+	if len(h) == 0 {
+		return nil
+	}
+	return &h[0]
+}
+
+// Truncate keeps only the latest snapshots in the history up to MaxHistorySize.
+// Since the history is maintained with most recent first, we simply truncate from the end.
+func (h *History) Truncate() {
```
matheuscscp:
```suggestion
// truncate keeps only the latest snapshots in the history up to MaxHistorySize.
// Since the history is maintained with most recent first, we simply truncate from the end.
func (h *History) truncate() {
```

This seems like an internal concern of `History`, should not be exported

api/v1/history_types.go
```diff
@@ -0,0 +1,137 @@
...
+			(*h)[i].LastReconciled = now
+			(*h)[i].LastReconciledDuration = durationMeta
+			(*h)[i].TotalReconciliations++
+			(*h)[i].Metadata = metadata
+			// Move to front if not already there
+			if i > 0 {
+				snapshot := (*h)[i]
+				copy((*h)[1:i+1], (*h)[0:i])
+				(*h)[0] = snapshot
+			}
```
matheuscscp, line 73: There's no test coverage for this case. I wrote one locally and found out that `copy()` takes care of overlapping memory areas, I was expecting it to wrongly copy the first item to all the other positions, and it doesn't, it works as expected 🤯

docs/api/v1/resourceset.md
```diff
@@ -766,7 +766,50 @@ Misconfigurations can include:
...
 
-### Inventory status
+### History
+
+With `.status.history` the operator tracks the reconciliation attempts over time, providing insights
+into the ResourceSet's behavior which can be used for audit, anomaly detection and debugging purposes.
+
+The history is stored as a list of snapshots, ordered by last reconciliation time. Each snapshot contains:
+
+- `digest`: A SHA256 digest that uniquely identifies the set of generated resources being reconciled
```
matheuscscp, line 776: Maybe we should document that the digest input is not the generated resources when it's not possible to generate them, e.g. error on getting inputs or on building the objects

api/v1/history_types.go
stefanprodan: In the CLI

api/v1/history_types.go
stefanprodan: This is tested in `TestResourceSetReconciler_HistoryErrorTracking` but fair point, it's missing from the api package.

api/v1/history_types_test.go
```diff
@@ -0,0 +1,280 @@
...
+					{
+						Digest:                 "sha256:xyz123",
+						FirstReconciled:        metav1.NewTime(time.Now().Add(-1 * time.Minute)),
+						LastReconciled:         metav1.NewTime(time.Now().Add(-1 * time.Minute)),
+						LastReconciledDuration: metav1.Duration{Duration: time.Second},
+						LastReconciledStatus:   "Success",
+						TotalReconciliations:   1,
+					},
+					{
+						Digest:                 "sha256:abc123",
```
matheuscscp, line 67: Let's add just one more item here so we exercise the situation where a wrong code moving the contents without taking overlapping memory areas into account could produce wrong results here 🙏

stefanprodan: Added, please check

PR #393, opened 2025-08-11
Extend OpenShift compatibility up to v4.19

wombat:


stefanprodan:
@wombat can you please signoff your commit, you can do it with `commit -s` and force push.

PR #395, opened 2025-08-11
Add jitter to ResourceSet requeue interval

stefanprodan:
Add jitter to ResourceSet requeue interval to spread the workqueue rate after a controller restart.

In addition, the dependency requeue interval which was hardcoded to 5sec, can now be changed with the `--requeue-dependency` flag.


matheuscscp (approved):
LGTM

PR #399, opened 2025-08-12
ResourceSet: Allow force apply of individual resources

stefanprodan:
Adresses https://github.com/fluxcd/flux2/discussions/5442


matheuscscp (approved):
LGTM

Issue #409, opened 2025-08-18
ResourceSet: Permute inputs from all sources into a single input set


matheuscscp:
By merging the inputs from all the provider objects, including the RSIP objects from `.spec.inputsFrom` and the RSET object itself via `.spec.inputs`, into a combined input set, we could allow users to compose resource sets from mixed input sources. This is useful for example to select a container image for a sidecar while also selecting a Helm chart OCI artifact, with both bundled together in the same input set.

For this to work, each input provider object must be scoped under a dedicated key in the Go templates, e.g.:

- `inputs.<input provider object scope>.<some exported input key>`
- `inputs.<input provider object scope>.<another exported input key>`

If only `fluxcd.controlplane.io/v1/ResourceSetInputProvider` is supported in RSET `.spec.inputsFrom` and we never introduce the field `.spec.inputsFrom[].namespace`, and also if we use the RSET object name for `.spec.inputs`, this dedicated key could be the name of the object. Because the Go template engine does not allow hyphens and other characters, we will normalize the object names by converting uppercase letters to lowercase, spaces and punctuation to `_` and dropping any characters that are not in the set `a-z0-9_`. In the normalization we will join the sequence of non-empty words resulting from these normalization procedures into the final normalized name separating each word by `_`, this simplifies names like `_a__b` into `a_b`.

This will result in N+1 input provider objects being merged into a single input set, N RSIP objects from `.spec.inputsFrom` + 1 RSET object from `.spec.inputs`.

However, each input provider object exports *a list* of input sets, and not just a single input set. Therefore, the best way to combine all inputs together is by doing a *permutation* of each input set from each input provider object with every possible input set from every other input provider object. Let's say the symbol `A_{i}` represents the amount of input sets exported by the i-th RSIP object, and `R` is the amount of input sets exported by the RSET object. Then the amount of input sets resulting from this permutation is `A_1 x A_2 x ... x A_N x R`. To achieve a single input set in the end, users should use RSIP filtering features. So if `A_{i} = 1` for every `i` and `R = 1`, then `A_1 x A_2 x ... x A_N x R = 1`.

To shift from the default flattening behavior to this one, we could introduce in RSET `.spec.inputStrategy.name: Permute`. The default can be `Flatten`.

Example:

```yaml
apiVersion: fluxcd.controlplane.io/v1
kind: ResourceSet
metadata:
  name: my-rset
  namespace: default
spec:
  inputStrategy: # defaults to {name: Flatten}
[...]

PR #420, opened 2025-09-05
operator: Introduce permutation for ResourceSet input sets

matheuscscp:
Closes: #409


stefanprodan (commented):
Please add this capability to the CLI `build rset` command along with tests.

matheuscscp:
> Please add this capability to the CLI `build rset` command along with tests.

While working on this, I found two bugs that I'm fixing in this PR:

- `build rset` was setting the wrong exported ID for `Static` RSIPs. I'm introducing the `(*RSIP).GetStaticInputID()` API and using it across the codebase to fix this.
- `build rset` is duplicating a large portion of logic implemented in the operator. In particular, `filterProviders()` was inconsistent with the filtering logic applied by the operator, so I'm fixing this.

matheuscscp:
> `build rset` was setting the wrong exported ID for `Static` RSIPs.

Actually this follows from the fact that the RSIP is loaded from a file, so there's no fix.

matheuscscp:
> Please add this capability to the CLI `build rset` command along with tests.

This is done.

I also marked `.spec.inputsFrom.apiVersion` and `.spec.inputsFrom.kind` as optional and deprecated, documenting that only a single value for each is accepted, i.e. `fluxcd.controlplane.io/v1` and `ResourceSetInputProvider` respectively. I also removed these fields everywhere in the code, tests and docs. This is for honoring our decision of not supporting any other types in the future, so the design of this feature remains appropriate forever.

api/v1/resourceset_types.go
```diff
@@ -71,20 +80,54 @@ type ResourceSetSpec struct {
...
 
-	// Kind of the input provider resource.
+	// Deprecated: The only accepted value is 'ResourceSetInputProvider'.
+	// If the field is not set, it defaults to 'ResourceSetInputProvider'.
 	// +kubebuilder:validation:Enum=ResourceSetInputProvider
-	// +required
-	Kind string `json:"kind"`
+	// +optional
+	// +deprecated
+	Kind string `json:"kind,omitempty"`
```
stefanprodan: Undo this please. Since v1 is GA, in a distant future when we'll release v2, users should be able to chose which version they want to use. The validation is enough, you can't specify anything else.

docs/api/v1/resourceset.md
```diff
@@ -179,7 +190,112 @@ spec:
...
+single input set per `ResourceSetInputProvider` object. If the number of permutations
+exceeds `10000`, the operator will fail the reconciliation of the `ResourceSet` with
+a `Stalled` condition.
+
+When merging the input sets from each object, the `Permute` strategy places each input
+set under a key that is the normalized name of the object providing the input set. The
+normalization process is as follows:
+
+1. Uppercase letters are converted to lowercase.
+2. Spaces and punctuation are converted to underscores (`_`).
```
stefanprodan:
```suggestion
2. Spaces, punctuation and `-` are converted to underscores (`_`).
```

api/v1/resourceset_types.go
matheuscscp: Fixed, thanks!

docs/api/v1/resourceset.md
stefanprodan: I think it's worth mentioning `-` as it's very common in Kubernetes names.

matheuscscp: Fixed, thanks!

internal/inputs/permuter.go
```diff
@@ -0,0 +1,221 @@
...
+	expectedPermutations  uint64
+}
+
+// PermuterOption is a functional option for configuring the Permuter.
+type PermuterOption func(*Permuter)
+
+// WithIncludeEmptyProviders configures the Permuter to include input providers
+// that do not export any inputs when generating permutations. This allows
+// generating an empty permutation when at least one provider has no inputs.
+func WithIncludeEmptyProviders() PermuterOption {
```
matheuscscp, line 34: If somebody asks for a flag to include empty providers in the feature, we are already prepared with this option. Just need to add a field in the spec!

Issue #379, opened 2025-08-03
Full multi-tenancy lockdown support


matheuscscp:
With https://github.com/fluxcd/flux2/issues/5465, `FluxInstance` could now offer an API that supports locking down also decryption and kubeconfig ServiceAccount names. For example:

```yaml
spec:
  cluster:
    tenantDefaultDecryptionServiceAccount: ... # when set locks down kustomize-controller with --default-decryption-service-account
    tenantDefaultKubeConfigServiceAccount: ... # when set locks down kustomize-controller and helm-controller with --default-kubeconfig-service-account
```

Furthermore, the field `.spec.cluster.tenantDefaultServiceAccount` must now also apply to all the other four Flux controllers that will implement `--default-service-account`. This will be a breaking change in the `FluxInstance` API, so alternatives are most welcome.

There's another behavior that must either be a breaking change or a misalignment between the `tenantDefault.*ServiceAccount` fields. The field `tenantDefaultServiceAccount` defaults to the `default` ServiceAccount. The two new `.spec.cluster` fields proposed above can either default to "lockdown disabled" or to `default` like the currently existing field. Defaulting to "lockdown disabled" is a misalignment, and defaulting to `default` is a breaking change similar to the one describe above.

Suggestions about alternatives are most welcome.

matheuscscp:
How about:

```yaml
spec:
  cluster:
    multitenant: true
    multitenantWorkloadIdentity: true # implies true for .objectLevelWorkloadIdentity
    objectLevelWorkloadIdentity: false # mistake! cannot set to false when .multitenantWorkloadIdentity is true
    tenantDefaultServiceAccount: ... # optional, defaults to `default` when at least one of .multitenant or .multitenantWorkloadIdentity is true
    tenantDefaultDecryptionServiceAccount: ... # optional, defaults to `default` when .multitenantWorkloadIdentity is true
    tenantDefaultKubeConfigServiceAccount: ... # optional, defaults to `default` when .multitenantWorkloadIdentity is true
```

The flag `--default-service-account` will be set in kustomize-controller and helm-controller if, and only if, `.spec.cluster.multitenant` is set to `true`.

The flag `--default-service-account` will be set in source-controller, notification-controller, image-reflector-controller and image-automation-controller if, and only if, `.spec.cluster.multitenantWorkloadIdentity` is set to `true`.

[...]

PR #422, opened 2025-09-11
[RFC-0010] Introduce workload identity multi-tenancy lockdown

matheuscscp:
Closes: #379


internal/builder/testdata/v2.6.0-golden/sharding.kustomization.yaml
matheuscscp, line 1: I'm reverting the behavior I introduced [here](https://github.com/controlplaneio-fluxcd/flux-operator/pull/380) that renders the feature gate with the value `false` only for Flux 2.6. No point in changing this behavior for a released version. In `>= 2.7.0` this behavior is still applied.

internal/builder/workload_identity.go
matheuscscp, line 1: All of this logic can now officially be called "complicated" at this point, but I don't think this will change drastically again anymore. If, let's say, in 2026 or 2027 we change the feature gate to opt-out, we can simply change the default value of `.spec.cluster.objectLevelWorkloadIdentity` to `true` in the kubebuilder annotations and that's it. Nothing to change in this file anymore.

docs/api/v1/fluxinstance.md
```diff
@@ -357,8 +360,86 @@ The `.spec.cluster.tenantDefaultServiceAccount` is optional and specifies the de
...
+
+**`.spec.cluster.tenantDefaultDecryptionServiceAccount`** (optional): Default service account
+for `kustomize-controller` SOPS decryption operations. Defaults to `"default"`.
+
+**`.spec.cluster.tenantDefaultKubeConfigServiceAccount`** (optional): Default service account
+for `kustomize-controller` and `helm-controller` remote cluster access via
+`spec.kubeConfig.configMapRef`. Defaults to `"default"`.
+
+**Complete example:**
+```yaml
```
stefanprodan:
```suggestion
Complete example:

```yaml
```

Please remove the bold from examples, and add a line before code blocks.

docs/api/v1/fluxinstance.md
```diff
@@ -357,8 +360,86 @@ The `.spec.cluster.tenantDefaultServiceAccount` is optional and specifies the de
...
+  cluster:
+    objectLevelWorkloadIdentity: true
+    multitenantWorkloadIdentity: true
+    tenantDefaultServiceAccount: "flux-tenant"
+    tenantDefaultDecryptionServiceAccount: "flux-decryption"
+    tenantDefaultKubeConfigServiceAccount: "flux-kubeconfig"
+```
+
+**Use cases:**
+- **Cloud workload identity**: Each tenant namespace has service accounts bound to cloud IAM roles
```
stefanprodan:
```suggestion
Use cases:

- **Cloud workload identity**: Each tenant namespace has service accounts bound to cloud IAM roles
```

internal/builder/testdata/v2.6.0-golden/sharding.kustomization.yaml
stefanprodan: We'll need golden files after we release v2.7

docs/api/v1/fluxinstance.md
matheuscscp: Fixed, thanks!

docs/api/v1/fluxinstance.md
matheuscscp: Thanks! I removed bold everywhere except for the use case list, where it seemed like a good place to have it

internal/builder/semver.go
```diff
@@ -123,6 +123,15 @@ func CheckMinimumVersion(fluxVersion, minimumVersion string) error {
 	return fmt.Errorf("constraint '%s' >= '%s' not satisfied", fluxVersion, minimumVersion)
 }
 
+// checkVersionAgainstConstraint checks if a parsed semver matches the given constraint.
+func checkVersionAgainstConstraint(version *semver.Version, constraint string) (bool, error) {
```
stefanprodan, line 127: The `CheckMinimumVersion` above does the same thing, please try to reuse that.

matheuscscp: Looking at what `CheckMinimumVersion` does, makes sense for `CheckMinimumVersion` to reuse `checkVersionAgainstConstraint` internally, please take a look

Issue #429, opened 2025-09-22
[Feature Request]: implement `flux-operator get all` subcommand


gecube:
Right now:

```bash
$ flux-operator get all
List Flux Operator resources

Usage:
  flux-operator get [command]

Available Commands:
  inputprovider List ResourceSetInputProviders
  instance      List Flux instances
  resources     List Flux custom resources and their status
  resourceset   List ResourceSets

Flags:
  -h, --help   help for get
```

but in the same time

```bash
$ kubectl get all
$ flux get all
```

both would be working.

Following the principle of minimizing friction, I would like to request the implementation of a `flux-operator get all` subcommand, which would act as a synonym for `flux-operator get resources`.

PR #433, opened 2025-09-25
cmd: Add `--ready-status` filter to `get resources` command

stefanprodan:
Changes:

- add `all` alias to `resources` e.g. `flux-operator get all` (fix: #429)
- add ready filtering for `True`, `False`, `Unknown`, `Suspended` e.g. `flux-operator get all -A --ready-status Suspended`


matheuscscp (approved):
LGTM

PR #435, opened 2025-09-25
Update `fluxcd/pkg/ssa` to v0.58.0

stefanprodan:
Migrate to `WaitForSetWithContext` for faster rollout as the health checks will be canceled by the shutdown signal.


matheuscscp (approved):
LGTM

Issue #430, opened 2025-09-23
Onboard source-watcher controller


stefanprodan:
We need to add [source-watcher](https://github.com/fluxcd/source-watcher) to the components list and include it in the builder logic for the Flux 2.7 distribution. 

xref: https://github.com/fluxcd/flux2/issues/5514

PR #437, opened 2025-09-26
Onboard source-watcher controller for Flux 2.7

stefanprodan:
Closes: #430

Prerequisite for: https://github.com/controlplaneio-fluxcd/distribution/issues/245


api/v1/common_types.go
```diff
@@ -101,8 +102,29 @@ const (
...
 )
 
+// FluxKinds is a list of all supported Flux kinds.
+var FluxKinds = []string{
+	FluxKustomizationKind,
+	FluxHelmReleaseKind,
+	FluxGitRepositoryKind,
+	FluxBucketKind,
+	FluxOCIRepositoryKind,
+	FluxHelmRepositoryKind,
```
matheuscscp, line 117: `FluxHelmChartKind` is not in this list

stefanprodan: Fixed thanks!

internal/builder/profiles.go
```diff
@@ -266,7 +266,7 @@ func GetProfileMultitenant(defaultSA string) string {
 const profileClusterMultitenant = `
 - target:
     kind: Deployment
-    name: "(kustomize-controller|helm-controller|notification-controller|image-reflector-controller|image-automation-controller)"
+    name: "(kustomize-controller|helm-controller|notification-controller|image-reflector-controller|image-automation-controller|source-watcher)"
```
matheuscscp, line 269: I think this change is ok since `source-watcher` Deployment will just not be present in Flux 2.6 :+1:

internal/builder/profiles.go
```diff
@@ -359,87 +359,12 @@ const profileClusterObjectLevelWorkloadIdentity = `
...
-    - op: add
-      path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/eventSources/items/properties/kind/enum/-
-      value: ResourceSetInputProvider
-    - op: add
-      path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/eventSources/items/properties/kind/enum/-
-      value: ResourceSetInputProvider
-    - op: add
-      path: /spec/versions/2/schema/openAPIV3Schema/properties/spec/properties/eventSources/items/properties/kind/enum/-
-      value: ResourceSetInputProvider
+const profileExternalArtifactFeatureGate = `
```
matheuscscp, line 362: I think we won't need to change this later for source-watcher after ExternalArtifact becomes a supported source in ArtifactGenerator, makes no sense to add this feature gate in source-watcher as ExternalArtifact is already a critical type for source-watcher (the output!) :+1:

internal/controller/fluxinstance_controller.go
```diff
@@ -359,12 +359,12 @@ func (r *FluxInstanceReconciler) build(ctx context.Context,
...
 	}
 
-	if err := options.ValidateAndApplyWorkloadIdentityConfig(obj.GetCluster()); err != nil {
+	if err := options.ValidateAndPatchComponents(); err != nil {
 		return nil, err
 	}
 
-	if options.HasNotificationController() {
-		options.Patches += builder.GetProfileNotification(options.Namespace)
+	if err := options.ValidateAndApplyWorkloadIdentityConfig(obj.GetCluster()); err != nil {
```
matheuscscp, line 366: If we are changing the order of these two `options.ValidateAnd...()` calls here, we also need to change in the CLI

internal/controller/fluxinstance_uninstaller.go
```diff
@@ -90,48 +90,54 @@ func (r *FluxInstanceReconciler) uninstall(ctx context.Context,
...
 			}
+		} else {
+			ctrl.LoggerFrom(ctx).Info("removed finalizers for " + listKind)
 		}
 	}
 
 	return errors.Join(errs...)
 }
 
+// getInstalledGVKs returns a map of installed Flux custom resource kinds to their preferred API versions.
```
matheuscscp, line 139:
```suggestion
// getInstalledGVKs returns a map of installed Flux custom resource list kinds to their preferred API versions.
```

internal/builder/components.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Ensure that source-watcher is not enabled for versions < 2.7.0.
+	if lt270 && o.HasSourceWatcher() {
+		return errors.New("source-watcher is only supported in Flux versions >= 2.7.0")
+	}
+
+	// Enable the ExternalArtifact feature gate if source-watcher is included.
+	if o.HasSourceWatcher() {
+		o.Patches += profileExternalArtifactFeatureGate
```
matheuscscp, line 79: Users should be able to enable `ExternalArtifact` without enabling `source-watcher`, but this is a feature we can ship later

internal/builder/build_test.go
```diff
@@ -965,6 +965,62 @@ func TestBuild_ShardingWithStorage(t *testing.T) {
 	g.Expect(found).To(BeTrue())
 }
 
+func TestBuild_SourceWatcher(t *testing.T) {
```
matheuscscp, line 968: This test should validate that the `ExternalArtifact` feature gate is enabled in `kustomize-controller` and `helm-controller`

internal/builder/components.go
stefanprodan: They can enable any gate with `.spec.kustomize.patches`

internal/builder/build_test.go
stefanprodan: The feature gate is tested in the controller pkg

matheuscscp (approved):
LGTM! :rocket:

PR #440, opened 2025-09-27
builder: Adapt RBAC patching for Flux 2.7

stefanprodan:
The removal of permission for creating ServiceAccount tokens broke for Flux 2.7, this PR handles the patching.


internal/builder/options.go
```diff
@@ -76,3 +83,22 @@ type Sync struct {
...
 	Provider   string
 }
+
+type VersionInfo struct {
+	Major int
+	Minor int
+	Patch int
+}
+
+func (o *Options) buildVersionInfo() error {
```
matheuscscp, line 93: This function could return an error if `ver.Major() < 2`

stefanprodan: preflight checks take care of this

PR #449, opened 2025-10-02
Add option to force the migration of Flux CRDs

stefanprodan:
This PR adds support for the `reconcile.fluxcd.io/forceAt` annotation to the `FluxInstance` and extended the CLI with:

```bash
flux-operator -n flux-system reconcile instance flux --force
```

When using the force option, the operator will run the migration of all Flux custom resources to their latest API version regardless if it previously done so.

ℹ️ This can be used on cluster where after upgrading Flux to 2.7 the following error persists:

```
dry-run failed: request to convert CR to an invalid group/version: notification.toolkit.fluxcd.io/v1beta1
```


matheuscscp (approved):
LGTM! :rocket:

PR #450, opened 2025-10-03
Include the managed fields in the storage migration process

stefanprodan:
This PR improves the storage migration of the Flux custom resources after an upgrade by updating the managed fields to the latest stable API version.

The operator will always run the storage migration after minor upgrades.


matheuscscp (approved):
LGTM! :rocket:

PR #467, opened 2025-10-14
mcp: Add `install_flux_instance` tool to MCP Server

stefanprodan:
Depends on: #473 

This tool enables automated installation of Flux Operator and Flux instances on Kubernetes clusters.

mcp-flux-install

### install_flux_instance

Installs Flux Operator and a Flux instance on the cluster from a manifest URL.

**Parameters:**

- `instance_url` (required): The URL pointing to the Flux Instance manifest file (supports HTTPS and OCI URLs)
- `timeout` (optional): The installation timeout duration (default: 5m)

**Output:**

Returns a detailed installation log including deployed resources with their change status.

**Installation Steps:**

The tool performs the following operations:

1. Downloads the Flux instance manifest from the provided URL
2. Downloads the Flux Operator manifests from the distribution artifact
3. Installs or upgrades the Flux Operator in the `flux-system` namespace
4. Installs or upgrades the Flux instance according to the manifest configuration
5. Waits for the Flux instance to become ready
6. Configures automatic updates for the Flux Operator

**Example URLs:**

- OCI Artifact: `oci://ghcr.io/org/manifests:latest#clusters/dev/flux-system/flux-instance.yaml`
- GitHub Gist: `https://gist.github.com/user/id#file-flux-instance-yaml`
- GitHub Repo: `https://github.com/org/repo/blob/main/clusters/dev/flux-system/flux-instance.yaml`
- GitLab Repo: `https://gitlab.com/org/proj/-/blob/main/clusters/dev/flux-system/flux-instance.yaml`

### DIFF
--- a/cmd/mcp/toolbox/install_instance.go
+++ b/cmd/mcp/toolbox/install_instance.go
@@ -12,7 +12,7 @@ import (
 
 	ssautil "github.com/fluxcd/pkg/ssa/utils"
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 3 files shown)


matheuscscp:
Let's merge #473 first to make it simpler

cmd/mcp/toolbox/install_instance.go
```diff
@@ -0,0 +1,202 @@
...
+	systemTools[ToolInstallFluxInstance] = systemTool{
+		readOnly:  false,
+		inCluster: true,
+	}
+}
+
+// installFluxInstanceInput defines the input parameters for installing Flux instance.
+type installFluxInstanceInput struct {
+	InstanceURL string `json:"instance_url" jsonschema:"The URL pointing to the Flux Instance manifest file."`
+	Timeout     string `json:"timeout,omitempty" jsonschema:"The installation timeout. Default is 5m."`
```
matheuscscp, line 40: I guess you already figured this, in the new SDK we define what is optional via `omitempty`, as fields without it are considered required :+1:

stefanprodan: Yeap! This is great

stefanprodan:
Tested after the SDK migration using an OCI URL:
Screenshot 2025-10-24 at 16 07 13

PR #478, opened 2025-10-27
operator: Wait for owned resources managed by ResourceSets

stefanprodan:
This PR makes the ResourceSet wait logic aware of Flux Operator owned kinds. For `dependsOn`, the `readyExpr` is no longer needed, also when a ResourceSet generates input providers, it knows how to wait for them to reconcile.

### DIFF
--- a/api/v1/common_types.go
+++ b/api/v1/common_types.go
@@ -19,6 +19,8 @@ const (
 
 	ReconciliationDisabledReason  = "ReconciliationDisabled"
 	ReconciliationDisabledMessage = "Reconciliation is disabled"
+
+	HealthCheckExpr = "status.conditions.filter(c, c.type == 'Ready').all(c, c.status == 'True' && c.observedGeneration == metadata.generation)"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


api/v1/common_types.go
```diff
@@ -19,6 +19,8 @@ const (
 
 	ReconciliationDisabledReason  = "ReconciliationDisabled"
 	ReconciliationDisabledMessage = "Reconciliation is disabled"
+
+	HealthCheckExpr = "status.conditions.filter(e, e.type == 'Ready').all(e, e.status == 'True' && e.observedGeneration == metadata.generation)"
```
matheuscscp: What does `e` mean? Isn't `e` a condition? Can it be changed to `c`?

internal/controller/resourceset_controller.go
```diff
@@ -323,7 +326,23 @@ func (r *ResourceSetReconciler) checkDependencies(ctx context.Context,
 				if !isReady {
 					return fmt.Errorf("dependency %s/%s not ready: expression '%s'", dep.APIVersion, ssautil.FmtObjMetadata(depMd), dep.ReadyExpr)
 				}
-			} else {
+			// Built-in CEL ready expression for ResourceSet and ResourceSetInputProvider.
+			case dep.Kind == fluxcdv1.ResourceSetKind || dep.Kind == fluxcdv1.ResourceSetInputProviderKind:
+				expr, err := cel.NewExpression(fluxcdv1.HealthCheckExpr)
+				if err != nil {
+					return err
+				}
```
matheuscscp, line 334: Since this expression is a constant, it can be built only once during the whole life of the process, on initialization.

stefanprodan: I think it's fine like this, this operation is instant and the expression is still compiled at wait time anyway.

PR #501, opened 2025-11-26
Add support for GitLab Environments to ResourceSetInputProvider

### LINKED ISSUE #414: Add support for GitLab environments to ResourceSetInputProvider
While the `GitLabMergeRequest` type for ResourceSetInputProvider works nicely for creating resources for merge requests, it requires some added complexity in both the GitLab pipeline and the ResourceSetInputProvider to only update when a new image is ready (cf. #215). Due to GitLab's permission model, managing MR labels from a pipeline also requires an access token with API write permissions and at least the Developer role, which seems a bit much "just" to adjust labels, it's a secret in need of rotation etc.

GitLab also has a concept of [environments](https://docs.gitlab.com/ci/environments/) that are well integrated with merge requests for review deployments. Starting, updating, and stopping a review environment from within a pipeline also is natively supported without any extra API access and allows for better control of when the environment needs to be updated, e.g. after an image was pushed.

Supporting environments as a data source for ResourceSetInputProviders would be a very welcome feature for enabling such integrations. The RSIP would use the [List Environments](https://docs.gitlab.com/ci/environments/) API to retrieve all currently `available` environments, with some additional filtering on the environment name, e.g. starting with `review/`.

The list response is a little sparse, but the response for [specific environments](https://docs.gitlab.com/api/environments/#get-a-specific-environment) is quite rich. To at the very least match the `GitLabMergeRequest` type, `last_deployment.deployable.commit` data should be included in the generated input. The name (and maybe also slug) of the environment should also be included. The `external_url` might also be helpful in cases where a user wants to have the environment configure some Ingress resource or similar.

I have not looked deeply into whether an analogous feature could be added for other source control platforms. GitHub also has the concept of environment and deployments, though I do not know how rich that is.

marvin-roesch:
This PR introduces support for [GitLab Environments](https://docs.gitlab.com/ci/environments/) as a new source for the ResourceSetInputProvider. The API has been expanded with a new `GitLabEnvironment` type and generic `includeEnvironment`/`excludeEnvironment` filters, with the option to add support for both GitHub and Azure DevOps deployment environments later down the line.

We additionally introduce a new `slug` key on the input since the environment name may not be valid for use in Kubernetes resource names. Its use is limited to the new environment support right now, but it could potentially be extended to branches and tags.

## Features
 * Get environment name, slug, and "deployable" commit SHA and branch name from GitLab Environments as input
 * Filter environments based on their name (e.g. only environments starting with `review/`)
 * Respects environment stopped status and only considers the last successful or currently running deployment
 * Supports reconciling RSIP while environment is currently being deployed

Especially that last point is very helpful for deploying MR/PR previews as a reconcile can be triggered from within an environment deploy job and the job (and thus deployment) is only successful if the reconcile went through. This basically enables synchronous feedback on the status of a preview deployment from within a GitLab pipeline.

As a small note: This PR currently does not include tests as I wanted some input from @stefanprodan on how to proceed on that front. Currently, the Git provider tests all rely on an actually existing project with the required properties. I could look into mocking some of these interactions instead.

This closes #414.


internal/gitprovider/result.go
```diff
@@ -43,6 +44,10 @@ func (r *Result) ToMap() map[string]any {
 		m["title"] = r.Title
 	}
 
+	if r.Slug != "" {
+		m["slug"] = r.Slug
+	}
```
stefanprodan, line 49: If we are adding slug, maybe it should be populated for GitLab MR also?

stefanprodan:
> I could look into mocking some of these interactions instead.

Please do

internal/gitprovider/result.go
matheuscscp: I haven't evaluated the whole context here but wanted to remind that in other places of the codebase we used a normalization algorithm to avoid introducing a slug field, e.g. when permuting inputs for RSET. Just want to bring this to consideration even if we conclude that introducing a slug field is better.

stefanprodan: We do have a `slugify` function in the ResourceSet templates, so maybe we can avoid adding this to the exported outputs https://fluxcd.control-plane.io/operator/resourceset/#templating-functions

marvin-roesch: For environments, at least, GitLab's slugifying algorithm isn't as straightforward, unfortunately. They limit the length to 24 characters and replace part of the result with a hash to compensate for the length limitation. See [their implementation](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/slug/environment.rb).

So simply deriving the same slug at templating time is a little difficult. Preserving the slug is important, however, as it may be used for an ingress hostname that in turn gets referenced in the environment URL (based on the `$CI_ENVIRONMENT_SLUG` CI/CD variable, for example).

As for including a slug for MRs as well: That's certainly something I can add, assuming GitLab exposes that in the relevant APIs. I mainly wanted to limit the scope of this PR to environments for now.

marvin-roesch:
I have converted all the GitLab provider tests to use the client library's provided mocks. I have also added tests for the environment support.

internal/gitprovider/result.go
stefanprodan: If we are adding `slug` as an exported input, I think we should do it consistently for all providers that expose such a prop. This could be done in a followup PR, before we release Flux Operator next.

marvin-roesch: I can certainly create a followup PR for this purpose, if you agree to keep the slug as an explicit input.

stefanprodan:
@marvin-roesch can you please squash, rebase with upstream and push. If you can do the `slug` next, I would merge this PR tomorrow and include it in the next release.

marvin-roesch:
Not sure if I can find the time today, but I'll create a follow-up PR for the `slug` field before end-of-week for sure, @stefanprodan. Squash and rebase is done.

stefanprodan (approved):
LGTM

Thanks @marvin-roesch 🏅

stefanprodan:
@marvin-roesch one more thing, can you please add a signoff to the commit with `git commit --signoff` and force push again.

marvin-roesch:
Done :ok_hand:

stefanprodan:
Thanks @marvin-roesch This is a great addition for GitLab users. Do you think we should rewrite the GitLab guide which is about MRs or add a new one for GitLab Environments? Is Environments a feature available in the community edition, can anyone use it or it's for GL EE only?

marvin-roesch:
Environments are available with all features required by this integration in all editions of GitLab, as far as I know.

As for the guide, I think adding a new one is the way to go. Using open Merge Requests as inputs is a bit simpler and probably sufficient for a lot of use cases. An environment-based solution grants more control but also requires a little more work in GitLab. For instance, an ideal solution also regularly cleans up review environments from GitLab via the intended [API endpoint](https://docs.gitlab.com/api/environments/#delete-multiple-stopped-review-apps), as the integration has to list *all* environments and would be slowed down by too many stale environments.

Furthermore, this integration could also conceivably be used for use cases outside of review environments. I could see some users having a GitLab environment e.g. per tenant or similar and creating ResourceSets based on that.

stefanprodan:
@marvin-roesch can you please help write the guide for GitLab Envs, the doc should go here https://github.com/controlplaneio-fluxcd/flux-operator/tree/main/docs/guides/resourcesets

marvin-roesch:
@stefanprodan Done: #522 :smile:

PR #513, opened 2025-12-10
feat(rsip): populate slugs for GitLab-provided inputs

marvin-roesch:
This is a follow-up to #501 as requested by @stefanprodan. I have so far only added slugs for GitLab branches, tags, and merge requests since I've realized that none of the providers actually expose a "slug" for these particular types in their APIs.

GitLab at least has CI/CD variables like `CI_COMMIT_REF_SLUG` where they provide these to users. AFAICT, neither Azure DevOps nor GitHub have a "standard" definition of how to slugify their data.

I have introduced a dedicated slugification function that matches GitLab's behavior, as `slug.Make` is configured globally and changing the defaults would mean a breaking change.

Let me know if you'd rather use a different approach or if this should also be expanded to GitHub and Azure DevOps (maybe with `slug.Make` there).


marvin-roesch:
Having trouble with KinD on my local machine right now, so sorry for playing whack-a-mole with the tests :sweat_smile:

PR #539, opened 2025-12-18
olm: add web port

matheuscscp:


erikgb (approved):
LGTM!

We should probably also update https://fluxoperator.dev/docs/web-ui/standalone-install/#access-the-web-ui, but in a separate PR.

stefanprodan:
Hmm why would we expose it for OpenShift? Is not like you can configure SSO for it. You need to deploy the Web UI in standalone mode for that.

stefanprodan:
I'm Ok with doing this once we allow OpenShift users to configure SSO via OLM. Right now this is not possible, as you can not mount secrets inside pods via OLM. We need to implement a flag (and an env var) for OLM users to set the Kubernetes secret name with the SSO config. We also have to watch this secret and reload the web server when it changes. This is not trivial work.

stefanprodan:
Closing this for now. The port should be added to the service when we implement #540

Issue #540, opened 2025-12-18
Load the web config from a Kubernetes Secret


stefanprodan:
Right now the the web server loads the SSO config from a local file. In the Flux Operator Helm chart we mount the config as a volume from a Secret generated by the chart from its values. For OLM users and those that don't want to use Helm, Flux Operator should expose a flag `--web-config-secret-name` and an environment variable `WEB_CONFIG_SECRET_NAME` that point to a Kubernetes Secret in the same namespace as the operator. The operator should watch and reload the server if the content of the Secret changes.

PR #543, opened 2025-12-20
web: introduce flags for watching config

matheuscscp:
Closes: #540


matheuscscp:
I'm not sure why and I tried various different fixes, but when the "Config is not initialized" screen starts getting 401s instead of 503s it does not go to the LoginPage :shrug:

stefanprodan:
>I'm not sure why and I tried various different fixes, but when the "Config is not initialized" screen starts getting 401s instead of 503s it does not go to the LoginPage 

This commit should fix it: https://github.com/controlplaneio-fluxcd/flux-operator/pull/554/changes/6de2aaf9d0b85171f42987cddb40b91f9ab1b4b1

Issue #546, opened 2025-12-21
v0.37 crashes with "fatal error: concurrent map writes"


atzix:
After updating to 0.37.x, flux-operator pod enters into infinite CrashLoopBackOff after some runtime

- The error seem to just appear randomly during runtime
- Once the error appears for the first time, it consistently happens on every subsequent container auto restart
- Restarting the node temporarily fixes the issue for ~6-12 hours before it appears again

```
{"level":"info","ts":"2025-12-21T21:30:35.298Z","logger":"setup","msg":"Operating in FIPS 140-3 mode, integrity check passed"}
{"level":"info","ts":"2025-12-21T21:30:35.327Z","logger":"web-server","msg":"authentication initialized successfully","authProvider":"None"}
{"level":"info","ts":"2025-12-21T21:30:35.327Z","logger":"setup","msg":"starting manager"}
{"level":"info","ts":"2025-12-21T21:30:35.327Z","logger":"controller-runtime.metrics","msg":"Starting metrics server"}
{"level":"info","ts":"2025-12-21T21:30:35.327Z","msg":"starting server","name":"health probe","addr":"[::]:8081"}
{"level":"info","ts":"2025-12-21T21:30:35.328Z","logger":"controller-runtime.metrics","msg":"Serving metrics server","bindAddress":":8080","secure":false}
{"level":"info","ts":"2025-12-21T21:30:35.328Z","logger":"runtime","msg":"attempting to acquire leader lease flux-system/flux-operator..."}
{"level":"info","ts":"2025-12-21T21:30:35.336Z","logger":"runtime","msg":"successfully acquired lease flux-system/flux-operator"}
{"level":"info","ts":"2025-12-21T21:30:35.337Z","msg":"Starting EventSource","controller":"namespace","controllerGroup":"","controllerKind":"Namespace","source":"kind source: *v1.Namespace"}
{"level":"info","ts":"2025-12-21T21:30:35.337Z","msg":"Starting EventSource","controller":"fluxinstance_artifact","source":"kind source: *v1.FluxInstance"}
{"level":"info","ts":"2025-12-21T21:30:35.338Z","msg":"Starting EventSource","controller":"fluxinstance","controllerGroup":"fluxcd.controlplane.io","controllerKind":"FluxInstance","source":"kind source: *v1.FluxInstance"}
[...]

PR #549, opened 2025-12-22
web: fix report cache concurrent writes

matheuscscp:
Fixes: #546

### DIFF
--- a/internal/web/router.go
+++ b/internal/web/router.go
@@ -5,6 +5,7 @@ package web
 
 import (
 	"context"
+	"encoding/json"
 	"io/fs"
 	"net/http"
 	"sync"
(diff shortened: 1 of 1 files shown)


internal/web/router.go
```diff
@@ -105,6 +106,11 @@ func (r *Router) refreshReportCache(ctx context.Context) {
 // getCachedReport returns the cached report if available.
 func (r *Router) getCachedReport() *unstructured.Unstructured {
 	r.reportCacheMu.RLock()
-	defer r.reportCacheMu.RUnlock()
-	return r.reportCache.DeepCopy()
+	report := r.reportCache
+	r.reportCacheMu.RUnlock()
+
+	b, _ := json.Marshal(report)
```
stefanprodan: This should be in inside the lock block

matheuscscp: Fixes, thanks!

PR #558, opened 2025-12-28
ci: Run unit and e2e tests on dedicated runners

stefanprodan:
Changes:
- Split workflow into unit and e2e tests (reduces test suite time in half)
- Add web API e2e test


matheuscscp (approved):
LGTM

Issue #557, opened 2025-12-26
Reduce the mean time to recovery (MTTR)


stefanprodan:
To reduce the mean time to recovery (MTTR), we should cancel the ongoing health checks when a new reconciliation is scheduled for the FluxInstance and ResourceSets. Coupled with Flux 2.8 MTTR (https://github.com/fluxcd/flux2/issues/5584), this feature will not only reduce the recovery time but also improve the time to deploy for GitOps with high frequency commits.

Users will be able to trigger a reconciliation that will cancel ongoing health checks, either by changing the ResourceSet spec in the desired state (Git, OCI, etc), or manually with `flux-operator reconcile rset` or using the Flux Web UI (reconcile action). Instead of waiting for the ResourceSet timeout, Flux Operator will mark the ResourceSet as` Ready=False` with `Reason=HealthCheckCanceled` and will schedule a new reconciliation immediately.

Issue #531, opened 2025-12-17
Add Suspend/Resume/Reconcile actions to resources in web UI


marvin-roesch:
I checked out the new status web UI and am really liking it so far, definitely going in a great direction! Potentially you already have this on your roadmap, but it'd be really interesting to expand it further in scope and add write operations for Flux resources as well rather than being a read-only view of the status.

It would be great if the `flux[-operator]` CLI operations of `suspend`, `resume`, and `reconcile` were doable from within the web UI on a resource's detail page and maybe also from the overview. Suspend/Resume could be a simple button that toggles the `suspend` flag. For reconciliation, I think a button that brings up a modal dialog where you can pick whether to include the source (analogously to the `--with-source` flag) and whether to force the reconciliation for Helm Releases would be ideal. The reconciliation timeout could also be specified there. The dialog could also stay open to show a progress report, akin to the current CLI output.

These buttons should only show if the user has patch permissions on the corresponding resource.

Finally, but definitely as more of a nice-to-have, the ability to perform these operations in bulk on a selection of resources would be great.

stefanprodan:
For authenticated users (when the UI is using [SSO](https://fluxoperator.dev/docs/web-ui/user-management#single-sign-on) or Dex with basic auth) we are considering implementing "Flux Actions" which would be a mirror of the operator CLI commands. Given that Kubernetes RBAC is very limited (you can grant `edit` or not), we will allow cluster admins to specify which actions they want to enable in the [web config spec](https://fluxoperator.dev/docs/web-ui/web-config-api).

matheuscscp:
Another interesting action would be the ability to delete an object that descends from a Flux/Flux Operator reconciler. Why only objects with this property? Well, because it ensures they were not applied manually and can therefore be eventually brought back by the reconciliation. Sometimes you can only unstuck something by deleting an object. This could be a last-resort alternative e.g. instead of making Flux forcefully recreate objects, which can be more dangerous as something that has a continuous effect (vs. a once-off action in a web UI).

kmgrime:
This would be really nice to have when replacing Weave Gitops, as the users are very used to this interaction to handle their workload.

stefanprodan:
We've shipped Suspend/Resume/Reconcile actions in Flux Operator v0.39. To enable the actions, the cluster admins can use the operator pre-defined Kubernetes Cluster Roles, docs here: https://fluxoperator.dev/docs/web-ui/user-management/#predefined-roles

[...]

PR #560, opened 2025-12-29
web: Implement action bar in the resource dashboard

stefanprodan:
Implement RBAC-aware action bar with real-time feedback in the resource dashboard:

- **Reconcile** triggers a reconciliation (with force for HelmRelease and RSIP). While the reconciliation is progressing the button stays disabled in loading state.
- **Pull** triggers a reconciliation of the HelmRelease or Kustomization source. While the source is pulling changes from upstream, the button stays disabled in loading state. If the source revision changes, the dashboard transitions in a progressing state and the Reconcile button switches to loading state.
- **Suspend/Resume** toggles the suspend field. While a resource is suspended, the other actions are disabled, only the Resume button is actionable. When the Resume button is clicked, the dashboard transitions in a progressing state and the Reconcile button switches to loading state.

Security features: 
- The action bar is rendered only if the user has `patch` permissions for the resource in-scope. Cluster admins must setup Kubernetes RBAC for the user's group to allow patching of Flux resources.
- The server impersonates the user's group and/or username when patching resources. If RBAC changes while a user clicks a button, the permission errors are reported in a panel that disappears after 5 seconds.  
- By default, actions are not allowed when the authentication is set to `Anonymous`. Cluster admins can opt-in to enable actions for anonymous access in the web-config and setup proper Kubernetes RBAC.
- When the authentication is set to `OAuth2`, all actions are allowed if RBAC permits. Cluster admins can opt-in to disable certain actions globally in the web-config. 
- Audit logs can be enabled by cluster admins in the web-config. When enabled, Kubernetes events are issued for each action taken by an user, these events can be forwarded to an alert provider by notification-controller.

_Currently the reconcile button is disabled in a loading state while a resource is progressing. We may want to reconsider this behaviour after we implement the MTTR features (#557 and https://github.com/fluxcd/flux2/issues/5584), maybe the button could change to "Restart"._

_Currently the pull action is not available for ExternalArtifact sources. To support ExternalArtifact we need to discover the ArtifactGenerator and trigger a reconciling for all its sources._

---
flux-action-bar

Part of: #531

### DIFF
--- a/docs/web/web-user-management.md
+++ b/docs/web/web-user-management.md
@@ -7,8 +7,7 @@ description: Flux Status Web UI user management guide.
 
 By default, the Flux Operator exposes the Web UI without authentication,
 providing read-only access to all Flux resources in the cluster.
-
-The Web UI is strictly read-only. Users can view the current state of Flux reconcilers
+Users can view the current state of Flux reconcilers
 and their managed resources, but cannot access sensitive data such as Kubernetes Secrets and ConfigMaps.
 
 We recommend restricting access to the Web UI using [Single Sign-On](#single-sign-on)
@@ -56,6 +55,60 @@ web:
       type: Anonymous
       anonymous:
         username: flux-web
+```
+
+### Anonymous User Actions
+
+To allow anonymous users to perform specific actions,
+such as triggering reconciliations or suspending/resuming resources,
+cluster admins have to enable these actions in the Web UI configuration and assign
+the necessary permissions to the anonymous user.
+
+To enable all actions for anonymous users, set the following values in the Flux Operator
+[Helm chart](https://github.com/controlplaneio-fluxcd/charts/tree/main/charts/flux-operator):
+
+```yaml
+web:
+  config:
+    authentication:
+      type: Anonymous
+      anonymous:
+        username: flux-web
+    userActions:
+      authType: Anonymous
+```
+
+Then, create the necessary RBAC resources to grant the required permissions to the `flux-web` user:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flux-web-actions
+rules:
+  - apiGroups:
+     - "fluxcd.controlplane.io"
+     - "source.toolkit.fluxcd.io"
+     - "source.extensions.fluxcd.io"
+     - "kustomize.toolkit.fluxcd.io"
+     - "helm.toolkit.fluxcd.io"
+     - "image.toolkit.fluxcd.io"
+     - "notification.toolkit.fluxcd.io"
+    resources: ["*"]
+    verbs: ["patch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: flux-web-global-actions
+subjects:
+  - kind: User
+    name: flux-web
+    apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
+  name: flux-web-actions
+  apiGroup: rbac.authorization.k8s.io
 ```
 
 For more information about configuring the Web UI, see the [Web Config API](web-config-api.md) documentation.


internal/web/kubeclient/client.go
```diff
@@ -6,19 +6,23 @@ package kubeclient
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
 	"time"
 
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/cache"
```
matheuscscp: Nit: We usually keep these separated after external deps

internal/web/kubeclient/client.go
```diff
@@ -287,3 +300,121 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
...
+		return true, nil
+	}
+
+	ssar := &authzv1.SelfSubjectAccessReview{
+		Spec: authzv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authzv1.ResourceAttributes{
+				Verb:      "patch",
+				Group:     group,
+				Resource:  resource,
+				Namespace: namespace,
```
matheuscscp, line 314: We can be more specific here, maybe you have RBAC to patch only a single specific resource of that kind in the namespace, but not all (RBAC allows resource names). This can done with the field `Name`.

internal/web/kubeclient/client.go
```diff
@@ -287,3 +300,121 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
...
+
+	if err := kubeClient.Create(ctx, ssar); err != nil {
+		return false, fmt.Errorf("failed to create SelfSubjectAccessReview: %w", err)
+	}
+
+	return ssar.Status.Allowed, nil
+}
+
+// AnnotateResource annotates a resource with the provided map of annotations.
+func (c *Client) AnnotateResource(ctx context.Context, gvk schema.GroupVersionKind, name, namespace string, annotations map[string]string) error {
```
matheuscscp: In the Flux CLI we do this inside RetryOnConflict :thinking: (I just noticed we don't do this in the Flux Operator CLI :thinking:)

matheuscscp: I'd move AnnotateResource and ToggleSuspension to the `internal/web/` package, I'd like to keep `kubeclient` minimal and only with operations that really need to look at the internal fields of the `Client` struct, like `CanPatchResource` that needs to look at the `c.client` field. It's important to keep the high level operations outside this package so that they don't accidentally use a field that they should not, e.g. the privileged client.

internal/web/kubeclient/client.go
```diff
@@ -287,3 +300,121 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
...
+		return fmt.Errorf("unable to annotate %s/%s/%s: %w", gvk.Kind, namespace, name, err)
+	}
+
+	return nil
+}
+
+// ToggleSuspension toggles the suspension of a Flux resource.
+// For Flux Operator resources, it uses the reconcile annotation.
+// For Flux resources, it patches the spec.suspend field.
+func (c *Client) ToggleSuspension(ctx context.Context, gvk schema.GroupVersionKind, name, namespace, requestTime string, suspend bool) error {
```
matheuscscp: Toggling is the action of inverting a boolean against its current state. I'd rename this function to `SetSuspension`.

internal/web/action.go
```diff
@@ -0,0 +1,140 @@
...
+
+// ActionResponse represents the response body for POST /api/v1/action.
+type ActionResponse struct {
+	Success bool   `json:"success"`
+	Message string `json:"message"`
+}
+
+// ActionHandler handles POST /api/v1/action requests to perform actions on Flux resources.
+func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
```
matheuscscp, line 52: These checks are a bit too defensive since the mux already takes care of matching the method.

stefanprodan: I prefer this for readability, you don't have to go on a search to figure out what methods are accepted.

internal/web/kubeclient/client.go
```diff
@@ -245,13 +250,22 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
 		return namespaces, true, nil
 	}
 
+	// Look up the plural for ResourceSet from FluxOperatorKinds.
+	var resourceSetPlural string
+	for _, kind := range fluxcdv1.FluxOperatorKinds {
```
matheuscscp: nit: I see `findFluxKindInfo` in `internal/web/resource.go`. This function could be extracted to the `api` package and used here (and everywhere else in the codebase where we look up a flux kind by name)

internal/web/resource.go
```diff
@@ -190,23 +191,46 @@ func (h *Handler) GetResource(ctx context.Context, kind, name, namespace string)
...
+	if err != nil {
+		log.FromContext(ctx).Error(err, "failed to check patch permission")
+	} else {
+		actionable = canPatch && h.conf.UserActionsEnabled()
+	}
+	if err := unstructured.SetNestedField(obj.Object, actionable, "status", "actionable"); err != nil {
+		return nil, fmt.Errorf("unable to set actionable in status: %w", err)
+	}
+
+	// Inject the available actions if actionable
```
matheuscscp:
```suggestion
```

Could delete this snippet and use only `.status.actions` in the frontend code (and rename it to `.status.userActions`)

internal/web/server.go
```diff
@@ -171,13 +171,14 @@ func RunServer(ctx context.Context, c cluster.Cluster,
...
 		// Successfully created all components with the new configuration.
 		confVersion = conf.Version
-		conf = nil // Clear conf to receive a new one in the next iteration.
 
 		// Create new handler.
 		newHandlerCtx, cancelNewHandlerCtx := context.WithCancel(context.Background())
 		newHandler, newHandlerStopped := NewHandler(newHandlerCtx, conf, spaHandler, kubeClient,
 			version, statusManager, namespace, reportInterval, eventRecorder, authMiddleware, serverLog)
 
+		conf = nil // Clear conf to receive a new one in the next iteration.
```
matheuscscp, line 180: Ouch! Sorry for that, hope it was not hard to debug

stefanprodan: I couldn't believe `system:masters` can not call the actions, then I realised what's going on, took some minutes :)))

internal/web/kubeclient/client.go
```diff
@@ -245,13 +250,22 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
...
+		if kind.Name == fluxcdv1.ResourceSetKind {
+			resourceSetPlural = kind.Plural
+			break
+		}
+	}
```
matheuscscp:
```suggestion
	rsetKind, _ := fluxcdv1.FindFluxKindInfo(fluxcdv1.ResourceSetKind)
```

internal/web/kubeclient/client.go
```diff
@@ -245,13 +250,22 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
...
+
 	// Check for cluster-wide access first in case the user has a ClusterRoleBinding.
 	clusterSSAR := &authzv1.SelfSubjectAccessReview{
 		Spec: authzv1.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &authzv1.ResourceAttributes{
 				Verb:     "get",
 				Group:    fluxcdv1.GroupVersion.Group,
-				Resource: "resourcesets",
+				Resource: resourceSetPlural,
```
matheuscscp:
```suggestion
				Resource: rsetKind.Plural,
```

internal/web/kubeclient/client.go
```diff
@@ -270,7 +284,7 @@ func (c *Client) filterNamespacesByAccess(ctx context.Context, namespaces []stri
 				ResourceAttributes: &authzv1.ResourceAttributes{
 					Verb:      "get",
 					Group:     fluxcdv1.GroupVersion.Group,
-					Resource:  "resourcesets",
+					Resource:  resourceSetPlural,
```
matheuscscp:
```suggestion
					Resource:  rsetKind.Plural,
```

stefanprodan:
@matheuscscp can you please add the `userActions` section to the web-config API docs.

docs/web/web-user-management.md
```diff
@@ -58,6 +57,60 @@ web:
...
+  - apiGroups:
+     - "fluxcd.controlplane.io"
+     - "source.toolkit.fluxcd.io"
+     - "source.extensions.fluxcd.io"
+     - "kustomize.toolkit.fluxcd.io"
+     - "helm.toolkit.fluxcd.io"
+     - "image.toolkit.fluxcd.io"
+     - "notification.toolkit.fluxcd.io"
+    resources: ["*"]
+    verbs: ["patch"]
```
matheuscscp, line 98: Maybe make this RBAC example complete with regards to read permissions?

```suggestion
    verbs: ["patch", "get", "list", "watch"]
```

matheuscscp (approved):
LGTM! :rocket:

docs/web/web-user-management.md
stefanprodan: There is a dedicated RBAC for read access above.

PR #561, opened 2025-12-29
web: add audit events to actions

matheuscscp:
Screenshot from 2025-12-29 14-04-33

### DIFF
--- a/internal/web/action.go
+++ b/internal/web/action.go
@@ -55,10 +55,8 @@ func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// Check if actions are enabled.
-	if !h.actionsEnabled() {
-		http.Error(w,
-			"User actions are only available when authentication is configured with the OAuth2 type",
-			http.StatusMethodNotAllowed)
+	if !h.conf.UserActionsEnabled() {
+		http.Error(w, "User actions are disabled", http.StatusMethodNotAllowed)
 		return
 	}
 
@@ -76,7 +74,7 @@ func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// Check if the specified action is enabled in the configuration.
-	if ua := h.conf.UserActions; ua != nil && len(ua.Enabled) > 0 && !slices.Contains(ua.Enabled, actionReq.Action) {
+	if !h.conf.UserActions.IsEnabled(actionReq.Action) {
 		http.Error(w, fmt.Sprintf("Action '%s' is not enabled", actionReq.Action), http.StatusMethodNotAllowed)
 		return
 	}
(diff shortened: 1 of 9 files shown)


internal/web/action.go
```diff
@@ -133,6 +141,32 @@ func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// Send audit event.
+	if h.eventRecorder != nil {
```
stefanprodan: Please make this opt-in via a Web Config field

internal/web/config/user_actions_types.go
```diff
@@ -0,0 +1,62 @@
...
+var (
+	// AllUserActions lists all possible user actions.
+	AllUserActions = []string{
+		UserActionReconcile,
+		UserActionSuspend,
+		UserActionResume,
+	}
+)
+
+// UserActionsSpec holds the actions configuration.
```
stefanprodan, line 31: Let's add here auth type with the default set to `OAuth2`. We should let admins opt-in for actions under `Anonymous` if they expose the UI internally or they use some Ingress with mTLS in front.

matheuscscp: Fixed, thanks!

internal/web/config/config_types.go
```diff
@@ -67,13 +73,34 @@ func (c ConfigSpec) Validate() error {
...
+
 	c.Authentication.ApplyDefaults()
 }
+
+// UserActionsEnabled checks if user actions are enabled.
+func (c *ConfigSpec) UserActionsEnabled() bool {
+	return c != nil &&
+		(c.UserActions.Enabled == nil || len(c.UserActions.Enabled) > 0) &&
+		(c.Authentication != nil && c.Authentication.Type == c.UserActions.AuthType)
+}
```
stefanprodan, line 106: Hmm so I can not enable actions when running as Flux Operator? Would this work:

```yaml
apiVersion: web.fluxcd.controlplane.io/v1
kind: Config
spec:
  userActions:
    authType: Anonymous
```

matheuscscp: Exactly. This would not work. Anonymous auth is not the same as no auth.

matheuscscp: I prefer it like this. If you want to enable actions you need at least Anonymous auth. You're forced to think what RBAC will the UI have.

matheuscscp: Minimal config (+ RBAC for `flux-web` `User`):

```yaml
apiVersion: web.fluxcd.controlplane.io/v1
kind: Config
spec:
  userActions:
    authType: Anonymous
  authentication:
    type: Anonymous
    anonymous:
      username: flux-web
```

stefanprodan (approved):
LGTM

PR #563, opened 2025-12-31
Ensure the `set_kubeconfig_context` tool persists changes

smuth4:
This also uses kubectl's internal loading logic across the board, as opposed to requiring $KUBECONFIG to be set.

It's intended to be backwards-compatible, as $KUBECONFIG will still be respected if set. I tested it with both out and in cluster installations. I saw $KUBECONFIG being referenced directly in `cmd/cli/main.go` but decided this PR was already changing a lot of files. It would also be feasible to return the merged `contexts` from the configs as-is, but decided to keep that backwards-compatible as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,24 +124,19 @@ To run the MCP Server using stdio, add the following configuration to your AI as
   "mcpServers": {
     "flux-operator-mcp": {
       "command": "/path/to/bin/flux-operator-mcp",
-      "args": ["serve"],
-      "env": {
-        "KUBECONFIG": "/path/to/.kube/config"
-      }
+      "args": ["serve"]
     }
   }
 }
 ```
 
-Replace `/path/to/bin/flux-operator-mcp` with the absolute path to the binary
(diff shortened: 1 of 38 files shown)


stefanprodan:
Where are the changes persisted? Are we altering the kubeconfig file on disk?

matheuscscp:
What's the issue with the current implementation? If I understand it correctly, we set the context in the shared `flags` and use it to construct a new client on every tool call. Is that not working?

smuth4:
My apologies, chalk this one up to PEBKAC, turns out I was doing something odd and incidentally fixed it while testing this change. I might split things out into a separate PR that just uses the kubeconfig loading logic to avoid the need for explicitly setting the KUBECONFIG env var, but this change is definitely too invasive, not to mention trying to fix what ain't broken.

> Where are the changes persisted? Are we altering the kubeconfig file on disk?

Minor note, but this line seems to imply to the LLM that it does indeed alter the kubeconfig file, but I assumed that the current behavior changing it only in-memory was desired: https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/cmd/mcp/toolbox/manager.go#L203

PR #584, opened 2026-01-10
web: document standard roles for UI access

stefanprodan:
Depends on: https://github.com/controlplaneio-fluxcd/charts/pull/132


docs/web/web-user-management.md
```diff
@@ -193,14 +127,104 @@ subjects:
...
 
 When a user from the `dev-team` group logs in, they can only search and view
 Flux resources in the `apps` namespace. Attempting to access resources in other
-namespaces will result in a "403 Forbidden" error.
+namespaces will result in a "403 Forbidden" error. To allow the dev team to
+perform actions on their resources, you can bind them to the `flux-web-admin` role instead.
 
 Users who are not part of any group with assigned permissions will only see the main dashboard,
 without access to any specific resources. The dashboard displays cluster-wide statistics
 such as the number of deployed apps and the readiness status of the Flux controllers.
```
matheuscscp: This line needs an update, users without any RBAC will now see zeros.

stefanprodan: Well spotted! I wander if we should show some kind of warning in main dashboard for users with no namespace access.

matheuscscp: That seems sensible 👍

matheuscscp: That can probably be checked in the frontend code when the report returns an empty list of namespaces

Issue #585, opened 2026-01-10
Scope groups => Cookies > 4 KB => failed to get auth storage cookie


powdermouse:
Hello,

I have the following problem:

- The web config works without any problems

```
web:
  config:
    authentication:
    :
        provider: OIDC
        scopes:
          - openid
          - email
          - profile
```		  
		  

- If you now want to process the groups as well, you configure

```
web:
  config:
    authentication:
    :
        provider: OIDC
        scopes:
          - openid
          - email
          - profile          
          - groups
```
		  
_You get the following error:_
**“authProvider”:“OAuth2/OIDC”,‘error’:“failed to get auth storage cookie: http: named cookie not present”**

_The following is displayed in the browser:_
**“... set cookie header was blocked because the cookie header was too large.”**

With scope groups, Keycloak presumably delivers the claim groups with all of the user's groups in the token.
Perhaps the WebUI stores the complete identity (claims + token) in a cookie after a successful OIDC login?
If there are a lot of/long groups, the cookie quickly exceeds 4096 bytes.
→ Browsers do not store cookies > 4 KB – the set cookie is ignored.
→ The auth cookie is missing in the next request ⇒ “failed to get auth storage cookie: http: named cookie not present.”

Without groups, the content remains small enough → cookie is stored → no error.

What must/can be configured, how and where, so that access to the groups works?

Thank you very much for your help and advice.

PR #586, opened 2026-01-10
web: implement cookie chunking for large OIDC tokens

stefanprodan:
Fix: #585

This PR adds automatic cookie chunking that splits large tokens across multiple cookies and reassembles them transparently on read. The implementation is backward compatible with existing single-cookie sessions. To prevent abuse, we limit the JWT payload to 35Kb, max 10 chunks.

PS. Assuming LDAP DN which is highly verbose, with an avg length of 80 chars per group, we now support ~300 groups.

### DIFF
--- a/internal/web/auth/cookies.go
+++ b/internal/web/auth/cookies.go
@@ -24,6 +24,18 @@ const (
 	cookiePathOAuth2LoginState = "/oauth2/"
 
 	cookieDurationShortLived = 5 * time.Minute
+
+	// Cookie chunking constants for handling large OIDC tokens.
+	// Maximum safe size per cookie chunk (leaving room for cookie attributes).
+	// Browser limit is 4KB but we use 3.5KB to leave room for cookie name, path,
+	// domain, and other attributes that count toward the limit.
+	cookieChunkMaxSize = 3584
+
+	// Maximum number of chunks allowed (prevents abuse, covers tokens up to ~35KB).
+	cookieChunkMaxCount = 10
+
+	// Separator for chunk cookies: auth-storage-1, auth-storage-2, etc.
+	cookieChunkSeparator = "-"
 )
 
 // setCookie sets a cookie in the response.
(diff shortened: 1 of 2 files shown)


internal/web/auth/cookies.go
```diff
@@ -135,3 +157,107 @@ func clearCookieFromResponse(w http.ResponseWriter, name string) {
...
+}
+
+// getChunkedCookieValue retrieves and reassembles a potentially chunked cookie value.
+// It first tries to read a single cookie (backward compatibility), then looks for chunks.
+func getChunkedCookieValue(r *http.Request, baseName string) (string, error) {
+	// First, try to get the base cookie.
+	baseCookie, err := r.Cookie(baseName)
+	if err != nil {
+		return "", fmt.Errorf("failed to get cookie %s: %w", baseName, err)
+	}
```
matheuscscp, line 197: The only error returned by r.Cookie() is when the cookie is not present, so here we should not error out. It's possible that a chunked cookie is still present.

stefanprodan: First chunk is the base cookie, I implemented it like this for backwards compatibility.

internal/web/auth/cookies.go
```diff
@@ -24,6 +24,18 @@ const (
...
+	// Maximum safe size per cookie chunk (leaving room for cookie attributes).
+	// Browser limit is 4KB but we use 3.5KB to leave room for cookie name, path,
+	// domain, and other attributes that count toward the limit.
+	cookieChunkMaxSize = 3584
+
+	// Maximum number of chunks allowed (prevents abuse, covers tokens up to ~35KB).
+	cookieChunkMaxCount = 10
+
+	// Suffix format for chunk cookies: auth-storage-1, auth-storage-2, etc.
+	cookieChunkSuffix = "-"
```
matheuscscp: nit

```suggestion
	cookieChunkSeparator = "-"
```

PR #590, opened 2026-01-12
cli: Add JSON support to `flux-operator export` commands

stefanprodan:
Add `flux-operator export -o, --output`: Specifies the output format (yaml, json). Default is yaml.

### DIFF
--- a/cmd/cli/export_report.go
+++ b/cmd/cli/export_report.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -16,9 +17,12 @@ import (
 
 var exportReportCmd = &cobra.Command{
 	Use:   "report",
-	Short: "Export the FluxReport resource in YAML format",
+	Short: "Export the FluxReport resource in YAML or JSON format",
 	Example: `  # Export the FluxReport to standard output
   flux-operator export report
+
+  # Export the FluxReport in JSON format
+  flux-operator export report -o json
 
   # Export the FluxReport to a YAML file
   flux-operator export report > flux-report.yaml`,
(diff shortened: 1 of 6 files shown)


cmd/cli/export_report.go
```diff
@@ -53,9 +65,20 @@ func exportReportCmdRun(_ *cobra.Command, args []string) error {
 	report := &reportList.Items[0]
 	cleanObjectForExport(report)
 
-	output, err := yaml.Marshal(report.Object)
-	if err != nil {
-		return fmt.Errorf("unable to marshal output to YAML: %w", err)
+	var output []byte
+	switch exportReportArgs.output {
+	case "json":
+		output, err = json.Marshal(report.Object)
```
matheuscscp: Is there indentation here?

cmd/cli/export_report_test.go
```diff
@@ -167,3 +168,90 @@ func TestExportReportCmd(t *testing.T) {
...
+				return
+			}
+
+			g.Expect(err).ToNot(HaveOccurred())
+
+			// Parse output based on format
+			obj := &unstructured.Unstructured{}
+			switch tt.outputFormat {
+			case "json":
+				err = json.Unmarshal([]byte(output), &obj.Object)
```
matheuscscp, line 240: I think `json.Unmarshal` can also parse YAML, so maybe a stronger check would be looking for `{` when expecting JSON, and making sure there's no occurrences of `{` when expecting YAML

cmd/cli/export_report.go
stefanprodan: Good catch, I always pipe to `jq`, but we should do pretty print. Fixed now.

cmd/cli/export_report_test.go
```diff
@@ -167,3 +168,91 @@ func TestExportReportCmd(t *testing.T) {
...
+			g.Expect(err).ToNot(HaveOccurred())
+
+			// Parse output based on format
+			obj := &unstructured.Unstructured{}
+			switch tt.outputFormat {
+			case "json":
+				g.Expect(output).To(HavePrefix("{"))
+				err = json.Unmarshal([]byte(output), &obj.Object)
+			case "yaml":
+				err = yaml.Unmarshal([]byte(output), &obj.Object)
```
matheuscscp, line 243:
```suggestion
				g.Expect(output).NotTo(HavePrefix("{"))
				err = yaml.Unmarshal([]byte(output), &obj.Object)
```

Issue #607, opened 2026-01-19
Debugging UX Improvment: Display RBAC information for the authenticated user


matheuscscp:
#531 shows that we need a better way to display the `username` and `groups` fields under the `impersonation` field from the report. This way users can debug RBAC issues more easily.

PR #611, opened 2026-01-20
web: return provider info for auth debugging

matheuscscp:
Part of: #607

### DIFF
--- a/internal/web/user/user.go
+++ b/internal/web/user/user.go
@@ -15,6 +15,7 @@ import (
 type Details struct {
 	Profile
 	Impersonation
+	Provider map[string]any
 }
 
 // Profile holds the user profile information for display purposes.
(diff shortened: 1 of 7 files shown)


internal/web/user/user.go
```diff
@@ -15,6 +15,7 @@ import (
 type Details struct {
 	Profile
 	Impersonation
+	Provider any
```
stefanprodan: Can we have this as `map[string]any` or is there any chance that provider would be a `string`?

matheuscscp: Fixed, thanks!

internal/web/user/user.go
```diff
@@ -122,30 +128,43 @@ func Permissions(ctx context.Context) Impersonation {
...
 		// We are using an identity provider that does not provide
 		// a name, or we are using Anonymous authentication.
-		// Return username and role (groups).
-		return s.Username, strings.Join(s.Groups, ", ")
+
+		// Do we have at least a username?
+		if s.Username != "" {
+			return s.Username
+		}
+		return "(unknown)"
```
stefanprodan:
```suggestion
		return "unknown"
```

matheuscscp: Fixed, thanks!

Issue #612, opened 2026-01-20
[web-ui] Ability to see who suspended a resource


erikgb:
And eventually "why" (could be a follow-up).

Context: Big multi-tenant clusters with multiple teams with many members.

As suspend/resume is now supported in the UI, ref. https://github.com/controlplaneio-fluxcd/flux-operator/issues/531, we would like to be able to see who (username) suspended a resource using the UI.

This could be visible as an annotation, which was a feature in the Weave GitOps:

````yaml
kind: Kustomization
apiVersion: kustomize.toolkit.fluxcd.io/v1
metadata:
  annotations:
    metadata.weave.works/suspended-by: jonasl
````

Or even "hidden" inside the managed fields, if the suspend/resume request performed by the UI was an Apply request performed on behalf of the user (setting field manager).

lind:
It could be something like this in the UI: 

<img width="1247" height="154" alt="Image" src="https://github.com/user-attachments/assets/95fbcd0a-b18b-4fb8-b9d5-41639f0c008b" />

PR #617, opened 2026-01-21
web: track the user who suspended a resource

stefanprodan:
Closes: #612

Track who suspended a resource by adding the `fluxcd.controlplane.io/suspendedBy: <username>` annotation. On resume, the  annotation is removed. The Web UI displays the `Suspended by: <username>` in the reconciler section of the resource dashboard.

---

Screenshot 2026-01-21 at 10 50 37

### DIFF
--- a/api/v1/common_types.go
+++ b/api/v1/common_types.go
@@ -33,6 +33,7 @@ var (
 	ForceAnnotation                  = fmt.Sprintf("%s/force", GroupVersion.Group)
 	RevisionAnnotation               = fmt.Sprintf("%s/revision", GroupVersion.Group)
 	CopyFromAnnotation               = fmt.Sprintf("%s/copyFrom", GroupVersion.Group)
+	SuspendedByAnnotation            = fmt.Sprintf("%s/suspendedBy", GroupVersion.Group)
 )
 
 // FluxObject is the interface that all Flux objects must implement.
--- a/internal/web/action.go
+++ b/internal/web/action.go
@@ -244,25 +244,54 @@ func (h *Handler) annotateResource(ctx context.Context, gvk schema.GroupVersionK
 // setSuspension sets the suspension state of a Flux resource.
 // For Flux Operator resources, it uses the reconcile annotation.
 // For Flux resources, it patches the spec.suspend field.
+// When suspending, it sets the SuspendedBy annotation to track the user who performed the action.
+// When resuming, it removes the SuspendedBy annotation if present.
 func (h *Handler) setSuspension(ctx context.Context, gvk schema.GroupVersionKind,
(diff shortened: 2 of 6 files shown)


sebastiangaiser:
Hey, thanks for implementing this awesome feature 🚀 
Would it be possible to add this to the flux cli, too? In order that a `flux suspend` would add the same annotation? Thanks in advance :wave:

stefanprodan:
There is no OIDC in the CLI, so anyone can act as anyone else e.g. `suspend --username "someone else"` which makes this feature a security nightmare for the CLI.

matheuscscp:
Unfortunately no. The identity here is not an input: it's extracted from the authentication. The same is not possible in CLIs: you could blame someone else. But we can introduce `--reason`

matheuscscp:
See https://github.com/controlplaneio-fluxcd/flux-operator/issues/618

marvin-roesch:
I like the idea of being able to provide an additional reason :+1: 

Very much spitballing here, but potentially this could be handled in a mutating webhook instead of at the client level? Authentication information is available there. That would have the benefit that whichever method is employed for triggering a reconciliation correctly associates an acting user with it. Right now, there is nothing stopping a user from running a manual patch against a resource that suspends it and mentions somebody else as the `fluxcd.controlplane.io/suspendedBy` value. Alternatively, somebody could remove the annotation after having suspended it.

A webhook could both automatically attach the correct user information and also perform some validation that only itself is enacting changes on that annotation, although I haven't put too much thought into the latter. My current thinking is that if a resource is suspended, the webhook would simply refuse any modifications to the annotation.

Finally, I think @sebastiangaiser was also thinking of making this a more general thing for Flux as a whole, not only the operator and its CLI. So using the "regular" flux CLI could have the same user attribution.

stefanprodan:
> A webhook could both automatically attach the correct user information and also perform some validation

We are not going to introduce webhooks in CNCF Flux nor in Flux Operator as this would complicate bootstrap with certs and introduce latency and possible SPOFs for the Kube API. But this can be a community project in a different org not endorsed or maintained by us.

> Right now, there is nothing stopping a user from running a manual patch against a resource

There is, if you do GitOps no end-user should have write access to the cluster. Especially now with the Web UI, there is really no reason what so ever for allowing end-users to alter the cluster state in production.

sebastiangaiser:
> There is, if you do GitOps no end-user should have write access to the cluster. Especially now with the Web UI, there is really no reason what so ever for allowing end-users to alter the cluster state in production.

I probably disagree a bit on that because as a user I would like to have the same possibility via the CLI and the UI. E.g. during an incident it would be beneficial to suspend/resume a `Kustomization`. We already implemented our own webhook which allows users to suspend/resume only. Probably it's worth extending that.

PR #623, opened 2026-01-21
web: track web session start

matheuscscp:


web/src/components/user/ProfilePage.jsx
```diff
@@ -83,11 +83,11 @@ export function ProfilePage() {
...
               </h1>
             </div>
-            {/* Session Started - only show when provider has iat claim */}
-            {userInfo?.provider?.iat && (
+            {/* Session Started - only show when present */}
+            {userInfo?.sessionStart && (
               <div class="hidden md:block text-right flex-shrink-0">
                 <div class="text-sm text-gray-600 dark:text-gray-400">Session Started</div>
-                <div class="text-lg font-semibold text-gray-900 dark:text-white">{formatTime(new Date(userInfo.provider.iat * 1000))}</div>
+                <div class="text-lg font-semibold text-gray-900 dark:text-white">{formatTimestamp(userInfo.sessionStart)}</div>
```
stefanprodan, line 90: Please add `sessionStart` to the report.js mock, you can copy the `iat` value.

matheuscscp: Fixed, thanks!

PR #654, opened 2026-02-02
web: extend workload views with CronJob support

stefanprodan:
Changes:
- Add CronJob support to the Workload API, Graph view and Workloads list view
- Display CronJob-specific statuses: `Idle`, `Suspended`, `Progressing`
- Refactor the pod status reporting to use the pod phase which is more accurate than kstatus
- List pods from CronJob's owned Jobs using field indexer for efficient lookups

---

Screenshot 2026-02-02 at 18 12 56

### DIFF
--- a/internal/web/workload.go
+++ b/internal/web/workload.go
@@ -13,20 +13,29 @@ import (
 	"time"
 
 	"github.com/fluxcd/cli-utils/pkg/kstatus/status"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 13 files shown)


internal/web/workload.go
```diff
@@ -202,46 +234,77 @@ func (h *Handler) GetWorkloadPods(ctx context.Context, obj *unstructured.Unstruc
...
+
+	return podsStatus, nil
+}
+
+// getCronJobPods returns the pods managed by a CronJob.
+// CronJob ownership is cascading: CronJob -> Job -> Pod.
+func (h *Handler) getCronJobPods(ctx context.Context, cronJob *unstructured.Unstructured) ([]WorkloadPodStatus, error) {
+	// Query Jobs owned by this CronJob using the field index on the cluster's cached client.
+	// We use WithPrivileges() because the user already has access to the CronJob,
+	// and the index is only available on the cluster's cache.
```
matheuscscp:
```suggestion
	// and the index is only available on the privileged cache. This effectively results
	// in escalating read access on CronJobs to status-only read access on Pods created
	// by the CronJob, which we consider safe.
```

Issue #572, opened 2026-01-06
Trigger a rollout restart from the Web UI


stefanprodan:
If a workload hits the progress deadline, the only way to restart it via Flux is by committing a change to the deployment's pod spec. We could introduce an action in the UI that triggers a rollout restart for workloads directly in the cluster, without tripping Flux's drift detection. For the `restart` button to be visible and actionable, users will need RBAC that allows `patch` for Deployments, StatefulSets and DaemonSets.

matheuscscp:
Can't RBAC be evaluated for the specific workload kind instead of requiring patch on all three kinds?

stefanprodan:
We will assert access by kind/name/namespace, like for resources, same code.

PR #659, opened 2026-02-04
web: implement `Rollout Restart` action for workloads

stefanprodan:
This PR adds an action bar to the workload view with the `Rollout Restart` action. The action is guarded by the `restart` RBAC verb.

The UI highlights the newly created pods for a short period of time and disables the `Rollout Restart` button while the rollout is in progress.

Closes: #572 

Screenshot 2026-02-04 at 11 50 22


api/v1/webconfig_types.go
```diff
@@ -31,6 +31,9 @@ const (
 
 	// UserActionDownload is the download user action.
```
matheuscscp:
```suggestion
	// UserActionDownload is the download user action for artifacts.
```

internal/web/workload.go
```diff
@@ -107,6 +107,13 @@ type WorkloadStatus struct {
 	// about the workload observed state.
 	StatusMessage string `json:"statusMessage,omitempty"`
 
+	// CreatedAt is the creation timestamp of the workload in RFC3339 format.
+	CreatedAt string `json:"createdAt,omitempty"`
```
matheuscscp:
```suggestion
	CreatedAt *time.Time `json:"createdAt,omitempty"`
```

matheuscscp: RFC3339 is used by `time.Time`

internal/web/workload.go
```diff
@@ -107,6 +107,13 @@ type WorkloadStatus struct {
...
+
+	// RestartedAt is the timestamp of the last rollout restart in RFC3339 format.
+	// Extracted from the kubectl.kubernetes.io/restartedAt annotation.
+	RestartedAt string `json:"restartedAt,omitempty"`
```
matheuscscp, line 115:
```suggestion
	RestartedAt *time.Time `json:"restartedAt,omitempty"`
```

internal/web/workload.go
```diff
@@ -360,6 +376,25 @@ func extractContainerImages(obj *unstructured.Unstructured) []string {
 	return containerImages
 }
 
+// extractRestartedAt extracts the kubectl.kubernetes.io/restartedAt annotation from the pod template.
+// For CronJob, the path is spec.jobTemplate.spec.template.metadata.annotations.
+// For Deployment/StatefulSet/DaemonSet, the path is spec.template.metadata.annotations.
+func extractRestartedAt(obj *unstructured.Unstructured) string {
```
matheuscscp, line 382:
```suggestion
func extractRestartedAt(obj *unstructured.Unstructured) time.Time {
```

Should parse to time.Time

internal/web/workload_action.go
```diff
@@ -0,0 +1,244 @@
...
+			Kind:    actionReq.Kind,
+		})
+
+		// Send the event.
+		notifier.
+			New(req.Context(), h.eventRecorder,
+				h.kubeClient.GetScheme(), notifier.WithClient(kubeClient)).
+			AnnotatedEventf(obj, annotations, corev1.EventTypeNormal, reason,
+				"User '%s' performed action '%s' on the web UI", perms.Username, actionReq.Action)
+	}
```
matheuscscp: Should eliminate this duplication with resource action

internal/web/audit.go
```diff
@@ -0,0 +1,116 @@
...
+			return
+		}
+
+		// Swap the object with the Flux resource managing it,
+		// so the event is associated with the Flux resource instead of the workload.
+		obj = fluxObj
+	}
+
+	token := fmt.Sprintf("%s/%s", obj.GetObjectKind().GroupVersionKind().Group, eventv1.MetaTokenKey)
+	annotations := map[string]string{
```
matheuscscp, line 74: The involved object reference is sent internally by `AnnotatedEventf` as structured data, retrieved from `obj`. When `obj` changes above due to sending an event for a non-Flux resource, we should send the reference of the non-Flux resource here in this map: `actionTargetAPIVersion`, `actionTargetKind`, `actionTargetName`, `actionTargetNamespace`

internal/web/workload_action.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	// Send audit event.
+	// Only fetch workload for reconciler ref if audit is enabled.
+	reconcilerRef := ""
+	if h.isAuditEnabled(actionReq.Action) {
+		workload := &unstructured.Unstructured{}
+		workload.SetGroupVersionKind(schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: actionReq.Kind})
+		if err := h.kubeClient.GetClient(ctx).Get(ctx, client.ObjectKey{
```
matheuscscp, line 148: You could have `patch` and `restart` but not `get`. But we can't let the audit fail at this point, so we need privileges here

internal/web/workload_action.go
```diff
@@ -0,0 +1,227 @@
...
+			Namespace: actionReq.Namespace, Name: actionReq.Name,
+		}, workload); err == nil {
```
matheuscscp, line 150: If an error is returned here, we need to print an error log and skip the audit because we already know it will fail

internal/web/workload_action.go
```diff
@@ -0,0 +1,227 @@
...
+			reconcilerRef = getReconcilerRef(workload)
```
matheuscscp, line 151: The behavior of `getReconcilerRef` is not changing in this PR, but I looked at it and the comment needs an update (it also looks at `.spec` in case of ExternalArtifact):

```go
// getReconcilerRef retrieves the Flux reconciler information
// from the labels of the provided unstructured object.
```

->

```go
// getReconcilerRef retrieves the Flux reconciler information
// from the provided unstructured object.
```

internal/web/audit.go
```diff
@@ -0,0 +1,116 @@
...
+	kind, namespace, name := parts[0], parts[1], parts[2]
+
+	gvk, err := h.preferredFluxGVK(ctx, kind)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get GVK for kind %s: %w", kind, err)
+	}
+
+	obj := &metav1.PartialObjectMetadata{}
+	obj.SetGroupVersionKind(*gvk)
+	if err := h.kubeClient.GetClient(ctx).Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, obj); err != nil {
```
matheuscscp, line 111: We need privileges here too

internal/web/workload_action.go
stefanprodan: How would you get to the page to click the button if you can not see the Flux resource?

stefanprodan:
@matheuscscp I'll do the audit refactoring in a dedicated PR

matheuscscp (approved):
LGTM! :rocket:

PR #663, opened 2026-02-05
web: enable audit for download actions

stefanprodan:
Move all the audit related code to `audit.go` and enable auditing for the artifact download action. Now all actions are properly audited, for actions like "Rollout Restart" the audit contains in metadata the subject e.g. `Deployment/apps/podinfo`.


matheuscscp (approved):
LGTM! :rocket:

PR #664, opened 2026-02-06
operator: consolidate RSIP api docs regarding change requests

matheuscscp:
In https://github.com/fluxcd/pkg/pull/1086 we introduced the `change_request` metadata field in the Event API to refer to change requests in notification-controller. We should use the same generic wording where applicable in the RSIP docs.


stefanprodan (approved):
LGTM

Thanks @matheuscscp

Issue #658, opened 2026-02-04
[web-ui] Run Job from CronJobs action


stefanprodan:
One feature request we got from Flux users migrating from Kubernetes Dashboard to the Flux Web UI, is the possibility to trigger a Job from an existing CronJob. In the workloads view, we [list the CronJobs](#654) managed by Flux. There we should add an action to run a job on-demand in the same way `kubectl create job test-job --from=cronjob/a-cronjob` does it. The UI should show the new job and it's pod status.

PR #665, opened 2026-02-06
web: implement `Run Job` action for CronJobs

stefanprodan:
This PR adds the "Run Job" to the workloads action bar. This action is guarded by the `restart` RBAC verb on CronJobs and requires `create` for Jobs.

The UI highlights the newly created job pods for a short period of time and disables the `Run Job` button while the rollout is initiated. The job pods created via the UI display the user that triggered them. The jobs created in the UI have their owner set to the CronJob, so they are subject to garbage collection in the same way as the jobs atomically created by the CronJob controller.

Closes: #658

https://github.com/user-attachments/assets/e70de4b3-e579-4ac1-8916-6e56ec47f9d5

### DIFF
--- a/internal/web/workload_action.go
+++ b/internal/web/workload_action.go
@@ -11,12 +11,14 @@ import (
 	"slices"
 	"time"
 
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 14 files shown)


internal/web/workload_action.go
```diff
@@ -47,9 +49,10 @@ type workloadKindInfo struct {
 
 // supportedWorkloadKinds maps workload kinds to their API group and supported actions.
 var supportedWorkloadKinds = map[string]workloadKindInfo{
-	"Deployment":  {group: "apps", plural: "deployments", actions: []string{fluxcdv1.UserActionRestart}},
-	"StatefulSet": {group: "apps", plural: "statefulsets", actions: []string{fluxcdv1.UserActionRestart}},
-	"DaemonSet":   {group: "apps", plural: "daemonsets", actions: []string{fluxcdv1.UserActionRestart}},
+	"Deployment":        {group: "apps", plural: "deployments", actions: []string{fluxcdv1.UserActionRestart}},
+	"StatefulSet":       {group: "apps", plural: "statefulsets", actions: []string{fluxcdv1.UserActionRestart}},
+	"DaemonSet":         {group: "apps", plural: "daemonsets", actions: []string{fluxcdv1.UserActionRestart}},
```
matheuscscp: Please add constants for the other kinds here since we're typing them by hand in other places

Issue #666, opened 2026-02-07
[web-ui] Delete pods owned by workloads managed by Flux


stefanprodan:
One feature request we got from Flux users migrating from Kubernetes Dashboard to the Flux Web UI, is the possibility to delete pods owned by Deployments, StatefulSets, DaemonSets and CronJobs managed by Flux. At times, performing a [rollout restart](https://github.com/controlplaneio-fluxcd/flux-operator/pull/659) is not desired if only a specific pod is stuck due to node DNS/CNI issues. Deleting just the stuck pods can be less disruptive then performing a rollout restart of workloads with hundreds of replicas.

We should implement a `delete` action for pods guarded by RBAC and allow users who have this permission to perform the action in the workload list view.

PR #667, opened 2026-02-07
web: implement `Delete` action for Pods

stefanprodan:
This PR adds a `delete` action for Pods owned by workloads managed by Flux. This action is guarded by RBAC and is traced by the Flux Operator audit.

Closes: #666 

---

https://github.com/user-attachments/assets/711ce4b1-da67-4074-bb5a-857a820c1efb

---

Screenshot 2026-02-07 at 12 06 02

### DIFF
--- a/docs/web/web-user-management.md
+++ b/docs/web/web-user-management.md
@@ -247,6 +247,12 @@ rules:
     verbs:
       - create
       - restart
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - delete
 ```
 
 Note that the `patch` verb is not enough to allow a user to perform actions in the Web UI.
@@ -259,3 +265,6 @@ The `download` verb allows users to download artifacts from Flux source resource
 
 The `restart` verb allows users to trigger a rollout restart on workloads
 (Deployment, StatefulSet, and DaemonSet) and to create Jobs from CronJobs.
+
+The `delete` verb on pods allows users to delete individual pods,
+which will be recreated by their controller.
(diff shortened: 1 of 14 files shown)


docs/web/web-user-management.md
```diff
@@ -259,3 +265,6 @@ The `download` verb allows users to download artifacts from Flux source resource
 
 The `restart` verb allows users to trigger a rollout restart on workloads
 (Deployment, StatefulSet, and DaemonSet) and to create Jobs from CronJobs.
+
+The `delete` verb on pods allows users to delete individual pods,
```
matheuscscp, line 269: We need to start discriminating which verbs are custom and which ones are not in this piece of doc

internal/web/audit.go
```diff
@@ -120,3 +121,64 @@ func (h *Handler) fetchReconcilerRef(ctx context.Context, kubeClient client.Clie
 
 	return obj, nil
 }
+
+// getPodOwnerWorkload resolves a pod's owner workload using the privileged client for audit purposes.
+// It walks the owner chain: Pod -> ReplicaSet/Job -> Deployment/StatefulSet/DaemonSet/CronJob.
```
matheuscscp, line 126: > Pod -> ReplicaSet/Job -> Deployment/StatefulSet/DaemonSet/CronJob

A bit overloaded 😅

Luckily, I get it 🤗

matheuscscp (approved):
LGTM! 🚀 

Code looks awesome!

Just a nit in the docs

docs/web/web-user-management.md
stefanprodan: This needs its own PR, audit docs are also very light and in a different doc, very hard to connect the two. I think we need a dedicated doc for actions, with sections for RBAC and audit.

matheuscscp: Cool, let's open an issue for that and resolve this thread :+1:

internal/web/audit.go
stefanprodan: Every Kube user should get this comment haha

PR #669, opened 2026-02-07
web: treat `DependencyNotReady` as progressing

stefanprodan:
When a Flux resource has `Ready: False` with reason `DependencyNotReady`, it means its dependencies haven't reconciled yet. This is a **transient condition**, the controller will automatically requeue and retry. Currently, the code treats all `Ready: False` conditions as `Failed`, which is misleading since dependency-not-ready is not a failure, it's normal in-progress behavior. 

This PR corrects the status reporting logic, the Web UI will display the `Progressing` status for the resources that are waiting on dependencies.


matheuscscp (approved):
LGTM! 🚀 

Turned out to be simple

PR #671, opened 2026-02-09
operator: fix  forced CRD migration on Flux minor upgrade

Aman-Cool:
This PR fixes an inverted condition in FluxInstance reconciliation that
prevented forced CRD migration during Flux minor version upgrades.

In `FluxInstanceReconciler.reconcile()` (internal/controller/fluxinstance_controller.go:585),
the code intended to force migration on minor upgrades checked
`err != nil && minor`. However, `builder.IsMinorUpgrade()` never returns
`(true, error)`, making this branch unreachable dead code.

As a result, forced migration was silently skipped on every Flux minor
upgrade, contradicting the comment and expected upgrade semantics.

The condition is corrected to `err == nil && minor`, restoring the
intended behavior: if a minor upgrade is successfully detected, CRD
migration is forced.

This is a minimal, one-character fix (`!=` → `==`) that restores
correctness without changing behavior in error cases.

### DIFF
--- a/internal/builder/semver_test.go
+++ b/internal/builder/semver_test.go
@@ -38,6 +38,80 @@ func TestMatchVersion(t *testing.T) {
 	}
 }
 
+func TestIsMinorUpgrade(t *testing.T) {
+	tests := []struct {
+		name        string
+		fromVer     string
+		toVer       string
+		expected    bool
+		expectError bool
+	}{
+		{
+			name:     "minor upgrade",
+			fromVer:  "v2.3.0",
+			toVer:    "v2.4.0",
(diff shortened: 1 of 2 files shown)


Aman-Cool:
There were no existing tests covering this code path, which is consistent with the branch being unreachable prior to this fix.

Aman-Cool:
@stefanprodan @matheuscscp ,Would appreciate a quick look when you have a chance.
Happy to adjust or extend this if you think any changes are needed.

stefanprodan (approved):
LGTM

Thanks @Aman-Cool great catch!

PR #690, opened 2026-02-13
cli: introduce `flux-operator diff yaml`

matheuscscp:
Output from:

```
./cli diff yaml \
  https://github.com/fluxcd/helm-controller/blob/v1.4.5/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml \
  https://github.com/fluxcd/helm-controller/blob/main/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml
```

```yaml
- op: remove
  path: /spec/versions/1
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/healthCheckExprs
  value:
    description: |-
      HealthCheckExprs is a list of healthcheck expressions for evaluating the
      health of custom resources using Common Expression Language (CEL).
      The expressions are evaluated only when the specific Helm action
      taking place has wait enabled, i.e. DisableWait is false, and the
      'poller' WaitStrategy is used.
    items:
      description: CustomHealthCheck defines the health check for custom resources.
      properties:
        apiVersion:
          description: APIVersion of the custom resource under evaluation.
          type: string
        current:
          description: |-
            Current is the CEL expression that determines if the status
            of the custom resource has reached the desired state.
          type: string
        failed:
          description: |-
            Failed is the CEL expression that determines if the status
            of the custom resource has failed to reach the desired state.
          type: string
        inProgress:
          description: |-
            InProgress is the CEL expression that determines if the status
            of the custom resource has not yet reached the desired state.
          type: string
        kind:
          description: Kind of the custom resource under evaluation.
          type: string
      required:
      - apiVersion
      - current
      - kind
      type: object
    type: array
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/install/properties/serverSideApply
  value:
    description: |-
      ServerSideApply enables server-side apply for resources during install.
      Defaults to true (or false when UseHelm3Defaults feature gate is enabled).
    type: boolean
- op: replace
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/rollback/properties/recreate/description
  value: |-
    Recreate performs pod restarts for any managed workloads.

    Deprecated: This behavior was deprecated in Helm 3:
      - Deprecation: https://github.com/helm/helm/pull/6463
      - Removal: https://github.com/helm/helm/pull/31023
    After helm-controller was upgraded to the Helm 4 SDK,
    this field is no longer functional and will print a
    warning if set to true. It will also be removed in a
    future release.
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/rollback/properties/serverSideApply
  value:
    description: |-
      ServerSideApply enables server-side apply for resources during rollback.
      Can be "enabled", "disabled", or "auto".
      When "auto", server-side apply usage will be based on the release's previous usage.
      Defaults to "auto".
    enum:
    - enabled
    - disabled
    - auto
    type: string
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/upgrade/properties/serverSideApply
  value:
    description: |-
      ServerSideApply enables server-side apply for resources during upgrade.
      Can be "enabled", "disabled", or "auto".
      When "auto", server-side apply usage will be based on the release's previous usage.
      Defaults to "auto".
    enum:
    - enabled
    - disabled
    - auto
    type: string
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/waitStrategy
  value:
    description: |-
      WaitStrategy defines Helm's wait strategy for waiting for applied
      resources to become ready.
    properties:
      name:
        description: |-
          Name is Helm's wait strategy for waiting for applied resources to
          become ready. One of 'poller' or 'legacy'. The 'poller' strategy uses
          kstatus to poll resource statuses, while the 'legacy' strategy uses
          Helm v3's waiting logic.
          Defaults to 'poller', or to 'legacy' when UseHelm3Defaults feature
          gate is enabled.
        enum:
        - poller
        - legacy
        type: string
    required:
    - name
    type: object
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/status/properties/history/items/properties/action
  value:
    description: Action is the action that resulted in this snapshot being created.
    type: string
- op: replace
  path: /spec/versions/0/schema/openAPIV3Schema/properties/status/properties/history/items/properties/apiVersion/description
  value: |-
    APIVersion is the API version of the Snapshot.
    When the calculation method of the Digest field is changed, this
    field will be used to distinguish between the old and new methods.
- op: add
  path: /spec/versions/0/schema/openAPIV3Schema/properties/status/properties/inventory
  value:
    description: |-
      Inventory contains the list of Kubernetes resource object references
      that have been applied for this release.
    properties:
      entries:
        description: Entries of Kubernetes resource object references.
        items:
          description: ResourceRef contains the information necessary to locate a
            resource within a cluster.
          properties:
            id:
              description: |-
                ID is the string representation of the Kubernetes resource object's metadata,
                in the format '<namespace>_<name>_<group>_<kind>'.
              type: string
            v:
              description: Version is the API version of the Kubernetes resource object's
                kind.
              type: string
          required:
          - id
          - v
          type: object
        type: array
    required:
    - entries
    type: object
```


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #693, opened 2026-02-16
web: add cached search index for Flux resources

stefanprodan:
Replace live Kubernetes API calls in quick search with an in-memory index built during the existing report refresh cycle. This makes the quick search functionally blazing fast, even on cluster with tens of thousands of Flux resources. The main tradeoff is that newly added resources will show up after `20s` from creation time.

The Web Config gains a new boolean option called `search.cached`. When enabled, the `/api/v1/resources` endpoint returns data from the periodically refreshed in-memory cache instead of querying the Kubernetes API in realtime. This reduces API server load and improves response latency for large clusters.

⚠️  For tenants with restricted access, the quick search will display all Flux resources present in the namespaces where the user has access to. We assume that a tenant has read-only access to all Flux kinds, even if we perform a `SelfSubjectAccessReview` only for `ResourceSet` kind in each namespace.

---

Screenshot 2026-02-16 at 14 47 52

### DIFF
--- a/internal/reporter/components_test.go
+++ b/internal/reporter/components_test.go
@@ -1,0 +1,214 @@
+// Copyright 2026 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package reporter
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestGetComponentsStatus_Success(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+
+	scheme := runtime.NewScheme()
+	g.Expect(appsv1.AddToScheme(scheme)).To(Succeed())
+
+	deploy1 := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "source-controller",
+			Namespace: "flux-system",
+			Labels: map[string]string{
+				"app.kubernetes.io/part-of": "flux",
+			},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": "source-controller"},
+			},
+			Template: corev1.PodTemplateSpec{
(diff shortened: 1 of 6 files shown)


api/v1/webconfig_types.go
```diff
@@ -91,6 +91,14 @@ type WebConfigSpec struct {
 	// +optional
 	Insecure bool `json:"insecure"`
 
+	// CachedSearch enables serving resource listings from the cached search index
+	// instead of querying the Kubernetes API in realtime. When enabled, the
+	// /api/v1/resources endpoint returns data from the periodically refreshed
+	// in-memory index, reducing API server load and improving response latency.
+	// The cache is updated on the same interval as the report cache.
+	// +optional
+	CachedSearch bool `json:"cachedSearch"`
```
matheuscscp: Maybe make it `.spec.search.cached`

internal/reporter/resource_status.go
```diff
@@ -0,0 +1,151 @@
...
+
+				break // Found Ready condition, no need to check others
+			}
+		}
+	}
+
+	// If kind is Alert or Provider set status to Ready as they don't have conditions
+	if (obj.GetKind() == fluxcdv1.FluxAlertKind ||
+		obj.GetKind() == fluxcdv1.FluxAlertProviderKind) &&
+		status == StatusUnknown {
```
matheuscscp, line 115:
```suggestion
		obj.GetKind() == fluxcdv1.FluxAlertProviderKind) {
```

internal/reporter/resource_status.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// if kind is HelmRepository and has .spec.type of 'oci', set status to Ready
+	if obj.GetKind() == fluxcdv1.FluxHelmRepositoryKind {
+		if specType, found, err := unstructured.NestedString(obj.Object, "spec", "type"); found && err == nil {
+			if specType == "oci" && status == StatusUnknown {
+				status = StatusReady
+				message = "Valid configuration"
+			}
+		}
+	}
```
matheuscscp, line 128: I'd remove this whole block

internal/web/search_index.go
```diff
@@ -0,0 +1,130 @@
...
+// SearchIndex holds cached search data built from the reporter's Compute() results.
+type SearchIndex struct {
+	mu        sync.RWMutex
+	resources []reporter.ResourceStatus
+	updatedAt time.Time
+}
+
+// Update replaces the indexed resources and sets the updated timestamp.
+func (idx *SearchIndex) Update(resources []reporter.ResourceStatus) {
+	idx.mu.Lock()
```
matheuscscp, line 25: The caller of `.Update()` should not be calling `buildSearchIndex()`

```suggestion
func (idx *SearchIndex) Update(resources []reporter.ResourceStatus) {
	resources = buildSearchIndex(resources)
	idx.mu.Lock()
```

PR #713, opened 2026-02-24
operator: Add ExternalService ResourceSetInputProvider

### LINKED ISSUE #150: `ResourceSet` dynamic inputs from 3rd-party orchestrators
With the introduction of [ResourceSet input providers](https://github.com/controlplaneio-fluxcd/flux-operator/pull/143), Flux Operator can dynamically deploy/uninstall applications based on data provided by Git services e.g. open/close Pull Request or push/delete Git branch.

While this feature is akin to GitOps, it limits the decision-making to the operations performed by developers on the Git service (e.g. GitHub, GitLab) and to the data provided by these services (e.g. Pull Request ID, labels, branch name, commit SHA).

For complex scenarios where the decision-making is based on a mix of data from Git services and from other systems (e.g. testing, monitoring, ticketing), the Flux Operator could delegate the decision-making to a 3rd-party orchestrator that could provide the necessary data for the `ResourceSet` controller to perform the desired actions (e.g. deploy app version X to the canary production cluster when the release manager has signed off on it).

In the case of a 3rd-party orchestrator, the desired state is not fully defined in Git, instead it is dynamically generated from multiple sources:

- Git Repository - where the `ResourceSet` and `ResourceSetInputProvider` manifests are defined.
- Container Registry - where the application images, Helm charts and Flux artifacts are stored.
- Orchestrator API - where the decision-making logic is implemented and from where the Flux Operator
  fetches the input data for the `ResourceSet` templates.

While we could consolidate the desired state in the Container Registry by packaging the Flux manifests from Git and pushing them to the registry, the inputs for the `ResourceSet` templates would still need to be fetched from the orchestrator API at runtime.

To close the feedback loop, the orchestrator could expose an API endpoint where Flux will send the status of the deployments managed by a `ResourceSet`. This is already implemented in Flux, where events are sent to external services via notification-controller. The Flux events contain the information a 3rd-party orchestrator would need: the app version deployed, the target cluster, the deployment status, etc.

## Proposal

To enable the Flux Operator to dynamically deploy applications based on data provided by 3rd-party orchestrators, the `ResourceSetInputProvider` should be extended with a new type `ExternalService` that would allow the operator to fetch arbitrary data from an external service via HTTP/S.

### Example

Example of a `ResourceSetInputProvider` that fetches input values from an orchestrator API:

```yaml
metadata:
  name: app1-feature-flags
  namespace: apps
  annotations:
spec:
    type: ExternalService
    url: https://my-orchestrator.com/api/v1/flux/inputs?env=prod&app=app1
    secretRef:
        name: my-orchestrator-basic-auth
    certSecretRef:
        name: my-orchestrator-client-certs
```

The orchestrator API should be able to authorise the Flux Operator API calls using basic auth and/or mTLS.

The orchestrator API should return a JSON object that contains an array of input values for the `ResourceSet` templates.

Example of a response that contains input values for two tenants of app1 with different feature flags:

```json
{
  "inputs": [
    {
      "tenant": "tenant1",
      "version": "2.0.0-rc.1",
      "featureFlags": [
        "feature1",
        "feature2",
        "feature3"
      ]
    },
    {
      "tenant": "tenant2",
      "version": "1.9.0",
      "featureFlags": [
        "feature1",
        "feature2"
      ]
    }
  ]
}
```

Example of a `ResourceSet` that uses the inputs fetched from the orchestrator API:

```yaml
kind: ResourceSet
metadata:
  name: app1
  namespace: apps
spec:
  inputsFrom:
      name: app1-feature-flags
  resources:
    - apiVersion: source.toolkit.fluxcd.io/v1beta2
      kind: OCIRepository
      metadata:
        name: app1-<< inputs.tenant >>
        namespace: apps
      spec:
        interval: 1h
        url: oci://ghcr.io/org/charts/app1
        ref:
          semver: << inputs.version >>
    - apiVersion: helm.toolkit.fluxcd.io/v2
      kind: HelmRelease
      metadata:
        name: app1-<< inputs.tenant >>
        namespace: apps
      spec:
        interval: 1h
        chartRef:
          kind: OCIRepository
          name: app1-<< inputs.tenant >>
        values:
          features: << inputs.featureFlags | join "," >>
          ingress:
            hosts:
              - host: app1-<< inputs.tenant >>.example.com
```

Given the above configuration, the orchestrator can instruct Flux Operator to perform the following actions:

- Onboard a new tenant by adding a new input set to the response. The Flux Operator will deploy an instance of app1 for the new tenant and expose it outside the cluster using a dedicated hostname.
- Offboard a tenant by removing the input set from the response. The Flux Operator will uninstall the app1 instance dedicated to the tenant.
- Toggle features for a tenant by updating the input set in the response. The Flux Operator will update the `HelmRelease` values and trigger a Helm upgrade.
- Change the app version for a tenant by updating the input set in the response. The Flux Operator will update the `OCIRepository` ref and trigger a Helm upgrade for the new Chart version.

dgunzy:
This PR adds support for ExternalService as a new ResourceSetInputProvider.spec.type, so Flux Operator can fetch dynamic ResourceSet inputs from third-party orchestrator APIs over HTTP/S.

Closes #150  

#### Implemented with these constraints in mind: 

-  reject payloads that don't fit into etcd (900Ki max)
- protect against OOM by not loading the whole response into memory if it's over the limit
- the expected format is an array, each element with a unique id
- trim the array based on .spec.filter.limit that has 100 as the default

I am attaching an e2e kind testing script, with its log outputs. 

[e2d-logs.txt](https://github.com/user-attachments/files/25524513/e2d-logs.txt)
[externalservice-e2e.sh](https://github.com/user-attachments/files/25524518/externalservice-e2e.sh)

### DIFF
--- a/api/v1/resourcesetinputprovider_types.go
+++ b/api/v1/resourcesetinputprovider_types.go
@@ -35,9 +35,11 @@ const (
 	InputProviderACRArtifactTag         = "ACRArtifactTag"
 	InputProviderECRArtifactTag         = "ECRArtifactTag"
 	InputProviderGARArtifactTag         = "GARArtifactTag"
+	InputProviderExternalService        = "ExternalService"
 
 	ReasonInvalidDefaultValues  = "InvalidDefaultValues"
 	ReasonInvalidExportedInputs = "InvalidExportedInputs"
+	ReasonInvalidSpec           = "InvalidSpec"
 
 	DefaultResourceSetInputProviderFilterLimit = 100
 )
@@ -48,12 +50,15 @@ const (
 // +kubebuilder:validation:XValidation:rule="!self.type.startsWith('Git') || self.url.startsWith('http')", message="spec.url must start with 'http://' or 'https://' when spec.type is a Git provider"
 // +kubebuilder:validation:XValidation:rule="!self.type.startsWith('AzureDevOps') || self.url.startsWith('http')", message="spec.url must start with 'http://' or 'https://' when spec.type is a Git provider"
 // +kubebuilder:validation:XValidation:rule="!self.type.endsWith('ArtifactTag') || self.url.startsWith('oci')", message="spec.url must start with 'oci://' when spec.type is an OCI provider"
+// +kubebuilder:validation:XValidation:rule="self.type != 'ExternalService' || self.url.startsWith('http')", message="spec.url must start with 'http://' or 'https://' when spec.type is 'ExternalService'"
+// +kubebuilder:validation:XValidation:rule="!has(self.insecure) || !self.insecure || self.type == 'ExternalService'", message="spec.insecure can only be set when spec.type is 'ExternalService'"
+// +kubebuilder:validation:XValidation:rule="self.type != 'ExternalService' || !self.url.startsWith('http://') || (has(self.insecure) && self.insecure)", message="spec.url must use 'https://' unless spec.insecure is true"
 // +kubebuilder:validation:XValidation:rule="!has(self.serviceAccountName) || self.type.startsWith('AzureDevOps') || self.type.endsWith('ArtifactTag')", message="cannot specify spec.serviceAccountName when spec.type is not one of AzureDevOps* or *ArtifactTag"
 // +kubebuilder:validation:XValidation:rule="!has(self.certSecretRef) || !(self.url == 'Static' || self.type.startsWith('AzureDevOps') || (self.type.endsWith('ArtifactTag') && self.type != 'OCIArtifactTag'))", message="cannot specify spec.certSecretRef when spec.type is one of Static, AzureDevOps*, ACRArtifactTag, ECRArtifactTag or GARArtifactTag"
 // +kubebuilder:validation:XValidation:rule="!has(self.secretRef) || !(self.url == 'Static' || (self.type.endsWith('ArtifactTag') && self.type != 'OCIArtifactTag'))", message="cannot specify spec.secretRef when spec.type is one of Static, ACRArtifactTag, ECRArtifactTag or GARArtifactTag"
 type ResourceSetInputProviderSpec struct {
 	// Type specifies the type of the input provider.
-	// +kubebuilder:validation:Enum=Static;GitHubBranch;GitHubTag;GitHubPullRequest;GitLabBranch;GitLabTag;GitLabMergeRequest;GitLabEnvironment;AzureDevOpsBranch;AzureDevOpsTag;AzureDevOpsPullRequest;GiteaBranch;GiteaTag;GiteaPullRequest;OCIArtifactTag;ACRArtifactTag;ECRArtifactTag;GARArtifactTag
+	// +kubebuilder:validation:Enum=Static;GitHubBranch;GitHubTag;GitHubPullRequest;GitLabBranch;GitLabTag;GitLabMergeRequest;GitLabEnvironment;AzureDevOpsBranch;AzureDevOpsTag;AzureDevOpsPullRequest;GiteaBranch;GiteaTag;GiteaPullRequest;OCIArtifactTag;ACRArtifactTag;ECRArtifactTag;GARArtifactTag;ExternalService
 	// +required
 	Type string `json:"type"`
 
@@ -72,13 +77,16 @@ type ResourceSetInputProviderSpec struct {
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
-	// SecretRef specifies the Kubernetes Secret containing the basic-auth credentials
+	// SecretRef specifies the Kubernetes Secret containing the credentials
 	// to access the input provider.
 	// When connecting to a Git provider, the secret must contain the keys
 	// 'username' and 'password', and the password should be a personal access token
 	// that grants read-only access to the repository.
 	// When connecting to an OCI provider, the secret must contain a Kubernetes
 	// Image Pull Secret, as if created by `kubectl create secret docker-registry`.
+	// When connecting to an ExternalService provider, the secret must contain either
+	// a 'token' key for bearer token authentication, or 'username' and 'password'
+	// keys for basic authentication.
 	// +optional
 	SecretRef *meta.LocalObjectReference `json:"secretRef,omitempty"`
 
@@ -87,12 +95,17 @@ type ResourceSetInputProviderSpec struct {
 	// - a PEM-encoded CA certificate (`ca.crt`)
 	// - a PEM-encoded client certificate (`tls.crt`) and private key (`tls.key`)
 	//
-	// When connecting to a Git or OCI provider that uses self-signed certificates, the CA certificate
-	// must be set in the Secret under the 'ca.crt' key to establish the trust relationship.
-	// When connecting to an OCI provider that supports client certificates (mTLS), the client certificate
+	// When connecting to a Git, OCI, or ExternalService provider that uses self-signed certificates,
+	// the CA certificate must be set in the Secret under the 'ca.crt' key to establish the trust relationship.
+	// When connecting to a provider that supports client certificates (mTLS), the client certificate
 	// and private key must be set in the Secret under the 'tls.crt' and 'tls.key' keys, respectively.
 	// +optional
 	CertSecretRef *meta.LocalObjectReference `json:"certSecretRef,omitempty"`
+
+	// Insecure allows connecting to an ExternalService provider over
+	// plain HTTP without TLS. When not set, the URL must use HTTPS.
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
 
 	// DefaultValues contains the default values for the inputs.
 	// These values are used to populate the inputs when the provider
--- a/docs/api/v1/resourcesetinputprovider.md
+++ b/docs/api/v1/resourcesetinputprovider.md
@@ -109,6 +109,7 @@ The following types are supported:
 - `ACRArtifactTag`: fetches input values from Azure Container Registry OCI artifact tags.
 - `ECRArtifactTag`: fetches input values from Elastic Container Registry OCI artifact tags.
 - `GARArtifactTag`: fetches input values from Google Artifact Registry OCI artifact tags.
+- `ExternalService`: fetches input values from a 3rd-party orchestrator API via HTTP/S.
 
 Throughout this document, the term **change request** is used to refer collectively to
 GitHub Pull Requests, GitLab Merge Requests, Azure DevOps Pull Requests, and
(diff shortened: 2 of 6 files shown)


dgunzy:
@stefanprodan @matheuscscp - let me know if this is the right approach. The e2e tests failed a few times, but didnt seem related to this PR. I will keep looking and iterating as well.

stefanprodan:
Tests are failing due to GHCR having an outage, nothing we can do about it.

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -723,6 +732,127 @@ func (r *ResourceSetInputProviderReconciler) callOCIProvider(ctx context.Context
...
+			username, password, err := r.getBasicAuth(obj, authData)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get auth credentials: %w", err)
+			}
+			req.SetBasicAuth(username, password)
+		}
+	}
+
+	// Execute request.
+	resp, err := httpClient.Do(req)
```
stefanprodan: I think we could use the same retryable client we use for GitLab: 

https://github.com/controlplaneio-fluxcd/flux-operator/blob/e6c5f8b0ef1ccb4bd9067a7bea5bce7340743665/internal/gitprovider/gitlab.go#L34

dgunzy:
Hi Stefan, I updated to use the same retryable client, reran the tests and everything seemed good. 

[e2d-logs.txt](https://github.com/user-attachments/files/25558627/e2d-logs.txt)

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -723,6 +733,136 @@ func (r *ResourceSetInputProviderReconciler) callOCIProvider(ctx context.Context
...
+	}
+
+	// Create request.
+	req, err := retryablehttp.NewRequest(http.MethodGet, obj.Spec.URL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+	req = req.WithContext(ctx)
+	req.Header.Set("Accept", "application/json")
+
```
stefanprodan, line 814: Let's set the user agent so that the orchestrator knows exactly which version of Flux Operator is calling it:

```go
req.Header.Set("User-Agent", "flux-operator/"+r.Version)
```

We can pass the Version to the reconciler in the same way we do for `FluxReportReconciler`.

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -723,6 +733,136 @@ func (r *ResourceSetInputProviderReconciler) callOCIProvider(ctx context.Context
...
+			username, password, err := r.getBasicAuth(obj, authData)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get auth credentials: %w", err)
+			}
+			req.SetBasicAuth(username, password)
+		}
+	}
+
+	// Execute request.
+	resp, err := rtClient.Do(req)
```
stefanprodan, line 829: Here we allow plain HTTP without TLS. Flux APIs have a dedicated field `insecure: true` for users to opt-in and for cluster admins to be able to block this usage at admission.

I suggest we do the same here. cc @matheuscscp

api/v1/resourcesetinputprovider_types.go
```diff
@@ -35,6 +35,7 @@ const (
 	InputProviderACRArtifactTag         = "ACRArtifactTag"
 	InputProviderECRArtifactTag         = "ECRArtifactTag"
 	InputProviderGARArtifactTag         = "GARArtifactTag"
+	InputProviderExternalService        = "ExternalService"
```
matheuscscp, line 38: Maybe we should call this `ExternalHTTPService`? In the future we could have maybe `ExternalGRPCService`?

api/v1/resourcesetinputprovider_types.go
```diff
@@ -72,13 +74,16 @@ type ResourceSetInputProviderSpec struct {
...
-	// SecretRef specifies the Kubernetes Secret containing the basic-auth credentials
+	// SecretRef specifies the Kubernetes Secret containing the credentials
 	// to access the input provider.
 	// When connecting to a Git provider, the secret must contain the keys
 	// 'username' and 'password', and the password should be a personal access token
 	// that grants read-only access to the repository.
 	// When connecting to an OCI provider, the secret must contain a Kubernetes
 	// Image Pull Secret, as if created by `kubectl create secret docker-registry`.
+	// When connecting to an ExternalService provider, the secret must contain either
+	// a 'token' key for bearer token authentication, or 'username' and 'password'
```
matheuscscp, line 88: I think we should design something here that will, in the future, be able to support short-lived credentials a-la https://github.com/fluxcd/flux2/issues/5681

api/v1/resourcesetinputprovider_types.go
stefanprodan: The URL proto would be enough to distinguish e.g. `grpc://`, `wss://`

api/v1/resourcesetinputprovider_types.go
stefanprodan: That would be an extension of what's in this PR in a distant future no? short-lived credentials would not need breaking changes in the current API.

api/v1/resourcesetinputprovider_types.go
dgunzy: Hmm so if we want to base the communication off of `https://` `grps://` and `wss://`,  I could: 
- Change the CEL expression to allow all of these
- do a match on the prefix
- call and parse the data based on the communication. 

Let me know, wss in particular would be an interesting change if we were going to hold the connection (no polling?)

stefanprodan: For now we should have a CEL expression for http and https if the provider is `ExternalService`. I don't see us using gRCP or WS any time soon.

matheuscscp: Yeah I just wanted the API to support other things in the future, we don't need to implement any of that now. The URL being used for distinction is good enough for me :+1: Right now it can only accept HTTP/S like Stefan said, if  we support other things we'll relax the validation when the implementation comes :+1:

api/v1/resourcesetinputprovider_types.go
matheuscscp: This comment is not requesting a specific change in the PR, it's just a vague reminder for the author to have this in mind and not make choices that would make it harder to support these other things in the future

internal/controller/resourcesetinputprovider_controller.go
dgunzy: Working on this - `insecure: true` should just apply to `ExternalService`  correct?

stefanprodan: In this PR yes.

 But we should extend it to `oci://` providers in the same way we do in OCIRepository.

dgunzy: Ok - want me to open an issue to add this to `oci://`? I can do that in another PR.

stefanprodan: Please do

matheuscscp: Agree with everything, thanks for working on this @dgunzy!

dgunzy:
@stefanprodan - I added `insecure: true`, and the flux operator version to the header. 

@matheuscscp  - if there is tweak to be more compatible for future short-lived credentials let me know, but I think I am following the established patterns for now. 

[e2e-logs.txt](https://github.com/user-attachments/files/25577296/e2e-logs.txt)
[externalservice-e2e.sh](https://github.com/user-attachments/files/25577302/externalservice-e2e.sh)

api/v1/resourcesetinputprovider_types.go
```diff
@@ -48,12 +49,15 @@ const (
...
 // +kubebuilder:validation:XValidation:rule="!self.type.startsWith('AzureDevOps') || self.url.startsWith('http')", message="spec.url must start with 'http://' or 'https://' when spec.type is a Git provider"
 // +kubebuilder:validation:XValidation:rule="!self.type.endsWith('ArtifactTag') || self.url.startsWith('oci')", message="spec.url must start with 'oci://' when spec.type is an OCI provider"
+// +kubebuilder:validation:XValidation:rule="self.type != 'ExternalService' || self.url.startsWith('http')", message="spec.url must start with 'http://' or 'https://' when spec.type is 'ExternalService'"
+// +kubebuilder:validation:XValidation:rule="!has(self.insecure) || !self.insecure || self.type == 'ExternalService'", message="spec.insecure can only be set when spec.type is 'ExternalService'"
+// +kubebuilder:validation:XValidation:rule="self.type != 'ExternalService' || !self.url.startsWith('http://') || (has(self.insecure) && self.insecure)", message="spec.url must use 'https://' unless spec.insecure is true"
```
stefanprodan, line 55: Please let's also enforce this in the code, some clusters like old OpenShift versions ignore CEL, and this will result in a CVE for us.

dgunzy: https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/internal/controller/resourcesetinputprovider_controller.go#L160-L169 - Can i validate just below this block in a similar way @stefanprodan ?  
Mark it as not ready, stall, error and notify?

stefanprodan: Yes

dgunzy:
@stefanprodan  - I added some validation to match the CEL expressions in the reconcile loop. 

[externalservice-e2e.sh](https://github.com/user-attachments/files/25580288/externalservice-e2e.sh)
[e2e-logs.txt](https://github.com/user-attachments/files/25580293/e2e-logs.txt)

ebourgeois:
Awesome, looking forward to this @stefanprodan !!

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -163,13 +164,46 @@ The ACR, ECR and GAR Artifact Tag providers export the same inputs as the OCI Ar
...
+      "featureFlags": ["feature1", "feature2"]
+    }
+  ]
+}
+```
+
+The response body must not exceed 900Ki in size and each `id` value must be unique within the array.
+The number of inputs is trimmed based on `.spec.filter.limit` (default 100).
+
+When connecting to an ExternalService provider, the secret referenced by `.spec.secretRef` must
```
matheuscscp:
```suggestion
When connecting to an ExternalService provider, the secret referenced by `.spec.secretRef` may
```

Otherwise it sounds like credentials are mandatory, which is not the case according to the code?

dgunzy: Good catch - not mandatory. Updated the docs 😄

matheuscscp:
@dgunzy I saw that in the  last push  a few more things apparently unrelated were updated in the markdown, was it intentional?

dgunzy:
> @dgunzy I saw that in the last push a few more things apparently unrelated were updated in the markdown, was it intentional?

No - some aggressive formatting from vs-code I think. Let me revert that. Sorry I did not notice.

PR #744, opened 2026-03-06
cli: Expose license key verification functions

matheuscscp:


cmd/cli/distro_verify_license_key.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/controlplaneio-fluxcd/flux-operator/internal/lkm"
+	"github.com/controlplaneio-fluxcd/flux-operator/cmd/cli/lkm"
```
matheuscscp, line 14: This is to make sure the CLI is always consistent with the exposed functions

PR #747, opened 2026-03-06
cmd: Add cosign verification to install command 

stefanprodan:
Add flags for verifying the cosign signature of the distribution artifact before installing.

⚠️ The verification is only compatible with Cosign v3 signatures which require Flux Operator v0.44+


matheuscscp (approved):
LGTM! 🚀

PR #762, opened 2026-03-13
Build preview image on push to tags

matheuscscp:
~~The `workflow_dispatch` new input `tag` is needed because we need to build this for `v0.44.0` and the tag push event was already missed.~~

Edit: This new input is no longer needed, I already used the workflow to build the tag we missed. The tag push trigger will do the job.

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*'
     paths-ignore:
       - 'docs/**'
       - 'config/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/preview.yaml
```diff
@@ -3,10 +3,17 @@ on:
...
+      - 'v*'
     paths-ignore:
       - 'docs/**'
       - 'config/**'
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'preview image tag; when set, builds as a tag release preview'
+        required: false
+        type: string
```
matheuscscp: If this new manual input is not specified, the behavior is the same as before

matheuscscp: No longer needed, I already built the tag we missed (`v0.44.0`)

PR #792, opened 2026-03-30
Use least privilege file permissions in CLI Dockerfile

archy-rock3t-cloud:
Change chmod=777 to chmod=755 for copied binaries since they only need read and execute permissions.


stefanprodan:
Did you test the image locally, are the binaries running Ok?

stefanprodan:
Superseded by https://github.com/controlplaneio-fluxcd/flux-operator/pull/841

PR #814, opened 2026-04-11
Migrate Terraform to the Flux Operator bootstrap module

### LINKED ISSUE #799: Migrate Terraform to the Flux Operator bootstrap module
We should migrate to https://github.com/controlplaneio-fluxcd/terraform-kubernetes-flux-operator-bootstrap

Tasks:
- [ ] Rewrite the Terraform example and docs in https://github.com/controlplaneio-fluxcd/flux-operator/tree/main/config/terraform
- [ ] Update the install guide in https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/docs/guides/operator/operator-install.md
- [x] Rewrite the Terraform bootstrap in https://github.com/controlplaneio-fluxcd/d2-fleet/tree/main/terraform

matheuscscp:
Closes: #799


config/terraform/clusters/staging/flux-system/flux-instance.yaml
```diff
@@ -0,0 +1,47 @@
...
+    path: config/terraform/clusters/staging
+    # pullSecret: flux-system # Uncomment if using GitHub App authentication with a private repository.
+  kustomize:
+    patches:
+      - target:
+          kind: Deployment
+        patch: |
+          - op: add
+            path: /metadata/annotations/cluster-region
+            value: ${CLUSTER_REGION}
```
stefanprodan: Let's use `commonMetadata` instead of patching the deployments.

```
spec:
  commonMetadata:
    labels:
      app.kubernetes.io/name: flux
    annotations:
      cluster-region: ${CLUSTER_REGION}
```

config/terraform/clusters/staging/flux-system/flux-operator.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: fluxcd.controlplane.io/v1
```
stefanprodan, line 1: Let's make this an exact match of the RSET used in the upgrade guide from https://github.com/controlplaneio-fluxcd/distribution/pull/379

PS. I realised now that we are missing the `dependsOn` in the distro docs...

config/terraform/variables.tf
```diff
@@ -1,67 +1,37 @@
...
-variable "cluster_size" {
-  description = "Cluster size, e.g. small, medium, large"
+variable "cluster_name" {
+  description = "Name of the cluster directory under clusters/ (e.g. staging)."
   type        = string
-  default     = "medium"
+  nullable    = false
 }
 
-variable "git_token" {
```
stefanprodan, line 25: We should keep both options to demo a Git token also, for GL users.

config/terraform/clusters/staging/flux-system/flux-instance.yaml
matheuscscp: Fixed, thanks!

config/terraform/clusters/staging/flux-system/flux-operator.yaml
matheuscscp: Fixed, thanks!

config/terraform/variables.tf
matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #824, opened 2026-04-16
cli: Add migrate owner command to clean up stale managed fields

### LINKED ISSUE #740: New CLI command to assist in migration of managed fields
I posted this question on Slack:

> I want to migrate resources sourced from a huge public Helm chart. The resources are currently managed by a Flux Kustomization pointing to a directory containing inflated/templated resources from the Helm chart. Now I want to replace this with a Flux HelmRelease, which should make the HelmRelease the owner of the resources in cluster. Any advice for this migration? I have done the opposite before, and it works like a breeze, but I know from experience that Helm is a lot more conservative in it's prosessing....

Which led to an interesting discussion with @stefanprodan and @matheuscscp. Read the full story here: https://cloud-native.slack.com/archives/CLAJ40HV3/p1772523525745189

To summarize, it seems like there is a consensus about the need for better tooling to migrate managed fields when trying to migrate from having a Flux `Kustomization` to `HelmRelease`  (or vice versa) as the controller for resources in the cluster. An initial migration will probably work well, but because of leftover managed fields from the old controller, a user could end up in a situation where the actual state of the resource will be different from the desired state. Example: If a user wants to unset an optional field after migration, the old value will be retained because the old controller still owns the field.

Tentative spec from Stefan:

> How I think it should be is have a command that you point to a HelmRelease, Kustomization, or ResourceSet. The command should detect Flux managers and clean up the managed fields of old Flux managers that don’t match the current owner. This command should enable the migration of managed fields between Flux appliers, no matter what you decide to do. It may well be that you migrate from HelmRelease to ResourceSet or any other combination.

Example command: `flux-operator migrate owner hr kube-prom-stack -n monitoring`.
A --dry-run could print all manages and the fields that would get stripped

stefanprodan:
The migrate owner command takes a Flux applier resource (`HelmRelease`, `Kustomization`, `ResourceSet`, or `FluxInstance`) and removes stale managed fields entries left behind by *other* Flux field managers on every resource listed in its `status.inventory`. This is intended to resolve post-migration issues where a field set by a previous Flux applier is retained even after the resource is reassigned to a new applier.

Example for migrating the managed fields of podinfo resources after helm-controller takes ownership from kustomize-controller:

```bash
flux-operator -n apps migrate owner hr/podinfo --dry-run
```

Transferring ownership workflow:

1. A Flux Kustomization named `podinfo` reconciles the  `podinfo` Deployment, Service, etc
2. We suspend the `podinfo` Kustomization and we delete it from source
3. We wait for Flux to sync it, the `podinfo` Deployment, Service, etc are left in the cluster
4. We add a HelmRelease for `podinfo`, helm-controller becomes the co-owner of the Deployment, Service, etc
5. We run `flux-operator -n apps migrate owner hr/podinfo `
6. From this moment on, helm-controller is the sole owner of podinfo and can remove fields from the Deployment, Service, etc

Note that this command requires Flux 2.8+

Closes: #740

### DIFF
--- a/api/v1/common_types.go
+++ b/api/v1/common_types.go
@@ -114,6 +114,7 @@ const (
 
 // Flux controller names.
 const (
+	FluxOperator                  = "flux-operator"
 	FluxSourceController          = "source-controller"
 	FluxKustomizeController       = "kustomize-controller"
 	FluxHelmController            = "helm-controller"
--- a/cmd/cli/README.md
+++ b/cmd/cli/README.md
@@ -181,33 +181,6 @@ Arguments:
 - `-n, --namespace`: Specifies the namespace scope of the command.
 - `--wait`: On resume, waits for the reconciliation to complete before returning.
 
-### Migrate Commands
-
-The `flux-operator migrate` commands are used to migrate the managed fields of
-Kubernetes resources to a new API version. This is useful when a CRD introduces
-a new storage version that adds defaulted fields, which can otherwise cause
-server-side applies to fail with `field not declared in schema` errors for
-field managers whose entries are still pinned to the old API version.
-
-The migration rewrites the `apiVersion` of every `metadata.managedFields` entry
-on every listed resource, across all field managers.
-
-The following command is available:
-
-- `flux-operator migrate resources`: Migrates the managed fields of the
-  specified Kubernetes resources to the given API version.
-
-Arguments:
-
-- `--api-version`: The target API version in the format `group/version` (required).
-- `--kind`: The kind of resources to migrate (required).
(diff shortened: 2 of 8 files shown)


matheuscscp (approved):
LGTM! 🚀 

It would be nice to document the full procedure for migrating applied resources from one Flux applier to another using this command.

matheuscscp:
> Note that this command requires Flux 2.8+

Just 2.8+ is not enough, the HelmRelease must have upgraded since the Flux upgrade to 2.8+, which is how `.status.inventory` gets populated.

PR #825, opened 2026-04-17
rset: Introduce `checksumFrom` for triggering rollout restarts

### LINKED ISSUE #271: Kustomize secretGenerator vs ResourceSet
Hello, absolutely grateful for this project, ResourceSet is a great tool.

When I add e.g. a Deployment to a ResourceSet, and attempt to inject secrets via envFrom, the regular kustomize suffixing (to trigger updates) doesn't work, it tries to find the secret without the suffix.

stefanprodan:
## Computing checksums from ConfigMaps and Secrets

To trigger a rolling update of a workload when the data of a referenced
ConfigMap or Secret changes, the `fluxcd.controlplane.io/checksumFrom` annotation
can be placed on any `metadata.annotations` map in the ResourceSet template.

The annotation value must be a comma-separated list of references in the format
`Kind/namespace/name`, where `Kind` is either `ConfigMap` or `Secret`. 

During reconciliation, the operator resolves each reference, computes a SHA256
digest over the combined data, and writes the result into a sibling
`fluxcd.controlplane.io/checksum` annotation as `sha256:<hex>`. The original
`checksumFrom` annotation is preserved.

Example of using `checksumFrom` to trigger a Pod rollout on a Deployment
whenever the referenced ConfigMap or Secret changes:

```yaml
kind: ResourceSet
metadata:
  name: frontend
  namespace: apps
spec:
  inputs:
    - name: podinfo
  resources:
    - apiVersion: apps/v1
      kind: Deployment
      metadata:
        name: << inputs.name >>
        namespace: << inputs.provider.namespace >>
      spec:
        selector:
          matchLabels:
            app.kubernetes.io/name: << inputs.name >>
        template:
          metadata:
            annotations:
              fluxcd.controlplane.io/checksumFrom: |
                ConfigMap/<< inputs.provider.namespace >>/<< inputs.name >>-config,
                Secret/<< inputs.provider.namespace >>/<< inputs.name >>-secret,
            labels:
              app.kubernetes.io/name: << inputs.name >>
          spec:
            containers:
              - name: podinfo
                image: ghcr.io/stefanprodan/podinfo:6.0.0
                envFrom:
                  - configMapRef:
                      name: << inputs.name >>-config
                  - secretRef:
                      name: << inputs.name >>-secret
```

After reconciliation, the Deployment's pod template will carry both annotations:

```yaml
template:
  metadata:
    annotations:
      fluxcd.controlplane.io/checksumFrom: "ConfigMap/apps/podinfo-config,Secret/apps/podinfo-secret"
      fluxcd.controlplane.io/checksum: "sha256:abc123def456..."
```

When the data of any referenced ConfigMap or Secret changes, the computed
checksum changes, the pod template hash changes, and Kubernetes performs a
rolling update of the Deployment.

References to ConfigMaps or Secrets generated by the same ResourceSet are
resolved from the pending apply. For references to resources outside the ResourceSet,
the checksum refreshes on the next reconciliation of the ResourceSet. To trigger
an immediate refresh when the referenced resource changes, you can set the
`reconcile.fluxcd.io/watch: Enabled` label on the referenced ConfigMap or Secret.

Note that on multi-tenant clusters, the service account
used by the ResourceSet must have the necessary permissions to read the ConfigMaps
and Secrets referenced by `checksumFrom`, if those are outside the ResourceSet itself.

Closes: #271


matheuscscp:
In other YAML multi-line string APIs that we have we don't require commas for separation, e.g. this one:

https://github.com/controlplaneio-fluxcd/distribution/tree/main/actions/push

stefanprodan:
GHA is not a good example of how Kubernetes APIs should look.

internal/controller/resourceset_controller_configs.go
```diff
@@ -0,0 +1,428 @@
...
+	return nil
+}
+
+// checksumResolver resolves checksumFrom references during a single
+// reconciliation pass. It indexes in-set ConfigMaps/Secrets so references
+// to objects rendered by the same ResourceSet are resolved from pending
+// data (not from an absent or stale cluster copy), and caches canonical
+// data per reference so repeated references fetch at most once.
+type checksumResolver struct {
+	ctx       context.Context
```
matheuscscp: Storing `ctx` in structs is a well-known Go anti-pattern and can be easily avoided in this case. It can still be passed to constructors for operations the constructor itself will call, but storing in the struct is a bad practice

```suggestion
```

matheuscscp:
> GHA is not a good example of how Kubernetes APIs should look.

Commas are very polemic and lead to discussions like "in a multi-line list of items, should we have a comma after the last item or not". Go has opted for a comma after the last item, JSON has opted for making a comma after the last item illegal but some JSON parsers allow it (e.g. JavaScript/TypeScript). I'm not aware of any standards for this in Kubernetes APIs, so I think commas are completely surperfluous here and we don't need them. Let's not catch ourselves in endless discussions about comma after the last item or not.

stefanprodan:
We can support comma ending and without. Would that be Ok? Or just drop multi-line and force one-line comma separated.

matheuscscp:
> Or just drop multi-line and force one-line comma separated.

This is the worst option possible

> We can support comma ending and without. Would that be Ok?

In that case, when using comma I think we should enforce comma after the last line

stefanprodan:
> In that case, when using comma I think we should enforce comma after the last line

Why? we trim and don't care if there is an ending comma or not.

matheuscscp:
Boom, we are trapped. See? Commas are evil.

stefanprodan:
Added support for trailing commas without making it a requirement

matheuscscp (approved):
LGTM! 🚀

PR #826, opened 2026-04-17
rset: Watch external references from `checksumFrom`

matheuscscp:
Also removing `ctx` from the `checksumResolver` struct.


stefanprodan (approved):
LGTM

Thanks @matheuscscp  🥇

PR #827, opened 2026-04-18
rset: Watch `convertKubeConfigFrom` Secrets

matheuscscp:
React immediately for kubeconfig conversions. Same label as always for ConfigMaps and Secrets: `reconcile.fluxcd.io/watch: Enabled`.


stefanprodan (approved):
LGTM

PR #828, opened 2026-04-19
rset: Add `includeEmptyProviders` option to `Permute` `inputStrategy`

jtyr:
## Summary

This adds an optional `.spec.inputStrategy.includeEmptyProviders` field to the `ResourceSet`, which changes how the `Permute` strategy handles providers that export zero inputs.

By default, empty providers are silently skipped and the remaining providers still permute among themselves. When `includeEmptyProviders: true`, the strategy behaves as a strict Cartesian product - if any provider exports zero inputs, the result is zero permutations.

## Motivation

I ran into this while experimenting with the `GitHubPullRequest` providers. The setup was a `ResourceSet` combining a dynamic PR provider with a static one holding shared defaults (`driftDetection`, `install`, `upgrade`, etc.) via `Permute`:

```yaml
inputStrategy:
  name: Permute
inputsFrom:
  - name: myapp-pull-requests      # dynamic, may have 0 entries
  - name: helmrelease-defaults     # static, always 1 entry
```

When there were no labeled PRs, the dynamic provider returned zero inputs. The remaining (static) provider then permuted by itself, which produced one render set that didn't have the PR-specific fields the template referenced. The result was that the `ResourceSet` ends up in an error state (`map has no entry for key "hello_pull_requests"`) until the first PR appears.

Mathematically, anything x empty set = empty set, so the expected behavior for "no PRs -> nothing to render" should be zero renders and no error. That's what `includeEmptyProviders: true` gives you.

## API

- New field: `.spec.inputStrategy.includeEmptyProviders` (`*bool`, optional, default `false`)
- CEL rule rejects the field unless `name == 'Permute'` (error: `includeEmptyProviders only applies when name is Permute`)
- Default behavior is unchanged - opt-in only

## Implementation

The underlying `WithIncludeEmptyProviders()` permuter option already existed and was tested in `internal/inputs/permuter_test.go`; this PR just wires it through from the CRD via `Combine()` in `internal/inputs/combine.go`.

## Tests

- Added three subtests to `TestCombine` covering the new flag path:
  - empty provider is skipped by default
  - empty provider causes zero results when `includeEmptyProviders: true`
  - flag is a no-op when all providers have inputs
- Added `TestResourceSetInputStrategyValidation` exercising the CEL rule via `testEnv.Create`
- All existing tests still pass, no changes to default behavior

### DIFF
--- a/api/v1/resourceset_types.go
+++ b/api/v1/resourceset_types.go
@@ -83,6 +83,7 @@ type ResourceSetSpec struct {
 // InputStrategySpec defines how the inputs are combined when multiple
 // input provider objects are used. Defaults to flattening all inputs
 // from all providers into a single list of input sets.
+// +kubebuilder:validation:XValidation:rule="!has(self.includeEmptyProviders) || self.name == 'Permute'",message="includeEmptyProviders only applies when name is Permute"
 type InputStrategySpec struct {
 	// Name defines how the inputs are combined when multiple
 	// input provider objects are used. Supported values are:
@@ -100,6 +101,15 @@ type InputStrategySpec struct {
 	// +kubebuilder:validation:Enum=Flatten;Permute
 	// +required
 	Name string `json:"name,omitempty"`
+
+	// IncludeEmptyProviders controls how input providers that export no
+	// inputs are treated. Only applies when Name is Permute. When true, if
+	// any provider has zero inputs the resulting permutation set is empty
+	// (mathematically correct Cartesian product behavior). When false or
+	// unset (default), providers with zero inputs are silently skipped and
+	// the remaining providers still permute among themselves.
+	// +optional
+	IncludeEmptyProviders bool `json:"includeEmptyProviders,omitempty"`
 }
 
 // GetInputStrategy returns the input strategy name.
@@ -108,6 +118,15 @@ func (in *ResourceSet) GetInputStrategy() string {
 		return InputStrategyFlatten
 	}
 	return in.Spec.InputStrategy.Name
+}
+
+// GetIncludeEmptyProviders returns whether providers with zero exported inputs
+// should be kept when combining inputs. Defaults to false.
+func (in *ResourceSet) GetIncludeEmptyProviders() bool {
+	if in.Spec.InputStrategy == nil {
+		return false
+	}
+	return in.Spec.InputStrategy.Name == InputStrategyPermute && in.Spec.InputStrategy.IncludeEmptyProviders
 }
 
 // InputProviderReference defines a reference to an input provider resource
--- a/config/crd/bases/fluxcd.controlplane.io_resourcesets.yaml
+++ b/config/crd/bases/fluxcd.controlplane.io_resourcesets.yaml
@@ -110,6 +110,15 @@ spec:
                   input provider objects are used. Defaults to flattening all inputs
                   from all providers into a single list of input sets.
                 properties:
+                  includeEmptyProviders:
+                    description: |-
+                      IncludeEmptyProviders controls how input providers that export no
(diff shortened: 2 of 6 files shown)


api/v1/resourceset_types.go
```diff
@@ -100,6 +101,15 @@ type InputStrategySpec struct {
...
 	Name string `json:"name,omitempty"`
+
+	// IncludeEmptyProviders controls how input providers that export no
+	// inputs are treated. Only applies when Name is Permute. When true, if
+	// any provider has zero inputs the resulting permutation set is empty
+	// (mathematically correct Cartesian product behavior). When false or
+	// unset (default), providers with zero inputs are silently skipped and
+	// the remaining providers still permute among themselves.
+	// +optional
+	IncludeEmptyProviders *bool `json:"includeEmptyProviders,omitempty"`
```
stefanprodan:
```suggestion
	IncludeEmptyProviders bool `json:"includeEmptyProviders,omitempty"`
```
No need to use pointer for boolean fields that default to false.

api/v1/resourceset_types.go
```diff
@@ -100,6 +101,15 @@ type InputStrategySpec struct {
 	// +kubebuilder:validation:Enum=Flatten;Permute
 	// +required
 	Name string `json:"name,omitempty"`
+
+	// IncludeEmptyProviders controls how input providers that export no
+	// inputs are treated. Only applies when Name is Permute. When true, if
+	// any provider has zero inputs the resulting permutation set is empty
+	// (mathematically correct Cartesian product behavior). When false or
```
stefanprodan: We can remove `(mathematically correct Cartesian product behavior)` from here.

stefanprodan (commented):
@jtyr please signoff your commit with `git commit -s` and force push. See: https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/CONTRIBUTING.md#certificate-of-origin

api/v1/resourceset_types.go
matheuscscp: I think it's a good call to leave this here.

api/v1/resourceset_types.go
```diff
@@ -110,6 +120,15 @@ func (in *ResourceSet) GetInputStrategy() string {
 	return in.Spec.InputStrategy.Name
 }
 
+// GetIncludeEmptyProviders returns whether providers with zero exported inputs
+// should be kept when combining inputs. Defaults to false.
+func (in *ResourceSet) GetIncludeEmptyProviders() bool {
+	if in.Spec.InputStrategy == nil || in.Spec.InputStrategy.IncludeEmptyProviders == nil {
+		return false
+	}
+	return *in.Spec.InputStrategy.IncludeEmptyProviders
```
matheuscscp:
```suggestion
	if in.Spec.InputStrategy == nil || in.Spec.InputStrategy.Name != InputStrategyPermute {
		return false
	}
	return in.Spec.InputStrategy.IncludeEmptyProviders
```

matheuscscp (commented):
Nice! I always wondered if anyone would have a use case for this option when I added it 🤗

api/v1/resourceset_types.go
jtyr: Fixed.

api/v1/resourceset_types.go
jtyr: Fixed.

api/v1/resourceset_types.go
jtyr: I also think we should keep it. But it's up to you @stefanprodan to decide.

matheuscscp (approved):
LGTM! 🚀 

Thanks very much for the PR (and for validating that this behavior can be useful)!

PR #834, opened 2026-04-24
operator: introduce support for `aws` provider with `GitRepository` sync

matheuscscp:
Flux 2.9 comes with support for the `aws` provider in the `GitRepository` API: https://github.com/fluxcd/flux2/issues/5866

I'm also enforcing the `.provider` versus `.kind` compatibility matrix, which was not being enforced before. Users can lock themselves out if they apply an invalid source for the sync.


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #839, opened 2026-04-29
mcp: ensure tool schemas include "properties" for OpenAI compatibility

yesid-lopez:
## Summary

OpenAI's function calling API requires the `"properties"` field to be present in tool input schemas, even when empty. The MCP Go SDK's `jsonschema.ForType` omits it for `struct{}` input types, producing `{"type":"object"}` without a `"properties"` key.

This causes MCP clients using OpenAI models (e.g., GPT-4o, GPT-5.5) to reject the tool schemas with:

```
Invalid schema for function 'mcp_flux_get_flux_instance': In context=(), object schema missing properties
```

### Affected tools

The three tools with `struct{}` input (no parameters):
- `get_flux_instance`
- `get_kubernetes_api_versions`
- `get_kubeconfig_contexts`

### Fix

Normalize tool input schemas in the `addTool` wrapper so that all tools always serialize with an explicit `"properties"` field. This covers both:
- Tools with no parameters (`struct{}` input → `InputSchema` is nil at registration time)
- Tools whose `*jsonschema.Schema` has a nil `Properties` map

Before:
```json
{"type":"object"}
```

After:
```json
{"type":"object","properties":{}}
```

### Test

Added `TestManager_ToolSchemasIncludeProperties` that:
1. Registers all tools on an in-memory MCP server
2. Connects a client and lists tools
3. Verifies every tool's `inputSchema` JSON contains the `"properties"` key

### Related

- Root cause is in the upstream `jsonschema-go` library which uses `omitempty` on `Properties`, but this fix is self-contained and doesn't require SDK changes.
- Affects any MCP client that forwards tool schemas to OpenAI's API (PicoClaw, custom integrations, etc.)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/mcp/toolbox/manager.go
+++ b/cmd/mcp/toolbox/manager.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
 	"github.com/controlplaneio-fluxcd/flux-operator/cmd/mcp/k8s"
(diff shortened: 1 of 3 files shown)


stefanprodan:
Hey @yesid-lopez is there an issue tracking this upstream in the [MCP Go SDK](https://github.com/modelcontextprotocol/go-sdk)? I think this is the responsibility of the SDK to produce tools schema compatible with all MCP clients.

yesid-lopez:
Hi @stefanprodan, thanks for the review!

Yes, this was tracked upstream in the MCP Go SDK as [#693](https://github.com/modelcontextprotocol/go-sdk/issues/693) and [#747](https://github.com/modelcontextprotocol/go-sdk/issues/747). Both are now closed.

The SDK maintainers' position is that `{"type":"object"}` without `"properties"` is valid per the JSON Schema spec, and **OpenAI is the non-conforming party**. They decided not to change the SDK's behavior since it follows the MCP spec correctly. Their recommendation for tool authors is to manually set `InputSchema` on parameter-less tools to satisfy OpenAI's stricter requirements.

The underlying `jsonschema-go` library was fixed in v0.4.2 to preserve empty `properties` maps, but the issue remains for tools using `struct{}` input because `ForType(struct{})` never creates a `Properties` map in the first place.

Given the SDK won't change, this fix handles it at the `flux-operator` level — which is exactly what the SDK maintainers suggest. Happy to adjust the approach if you have a different preference.

stefanprodan:
Please run `make tidy` and force push.

stefanprodan (approved):
LGTM

Thanks @yesid-lopez

matheuscscp (approved):
LGTM

PR #841, opened 2026-04-29
Bundle Flux and Flux Schema in the CLI container image

stefanprodan:
Besides `kubectl`, include `flux` and `flux-schema` binaries in the flux-operator-cli container image. This will enable users to use a single container image for CI and in Kubernetes Jobs to automate all things Flux related.

Also switch the base image to `gcr.io/distroless/static:debug-nonroot` for being able to run shell scripts with busybox.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -8,4 +8,12 @@ updates:
         patterns:
           - "*"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+  - package-ecosystem: "docker"
+    directory: "/cmd/cli"
+    groups:
+      cli-tools:
+        patterns:
+          - "*"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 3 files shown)


matheuscscp:
Why not add `helm` and `yq` as well? https://github.com/controlplaneio-fluxcd/terraform-kubernetes-flux-operator-bootstrap/blob/639c3229f4b124bb7bd4368d16decc5e3ba99f56/Dockerfile#L3-L4

Then the TF module can depend on a single image

stefanprodan:
Hmm we could actually switch the CLI image to `static-debian12:debug-nonroot` then the TF module will not need its own image at all.

matheuscscp:
It still adds the `bootstrap.sh`

stefanprodan:
> It still adds the bootstrap.sh

It doesn't need to be in the image, it can be mounted from a ConfigMap in the chart.

matheuscscp (approved):
LGTM! 🚀

PR #852, opened 2026-05-04
Include exported inputs into drift event for ResourceSetInputProvider

### LINKED ISSUE #850: Include exported inputs into drift event for ResourceSetInputProvider
#234 is now implemented and I was able to get a hand onto this and was able to test it. While the changes work for me so far I'm missing a possibility to see the actual exported inputs in the event. Can we add them always or atleast have a way to do so?

Currently the event looks like this (in Microsoft Teams)
Image

With this simple change (https://github.com/Mik4sa/flux-operator/commit/f14d94a3b55ef4e047711b58c99c4ea670bc8e85) the result looks like this
Single value:
Image

Multiple values:
Image

Is this a change we can add like this or would a need a different solution?

Mik4sa:
Fixes #850

Adds the new annotation `fluxcd.controlplane.io/eventMetadataExportedInputKeys`. With it the user can control which fields from exported inputs are included in the event metadata for the Flux event (alert). Optionally the name of event metadata fields can be customized.

What would it look like (for example for ResourceSetInputProvider `type: OCIArtifactTag` and Microsoft Teams provider):

### No annotation
No behaviour change.
image

### Annotation value: "tag,digest"
image

**What I just realized. Here is the "digest" field missing. Is this a reserved name or so?**

### Annotation value: "tag=image-tag,digest=manifest-digest"
image

### Multiple exported inputs
When there are multiple exported inputs (more than a single) then the index of the exported input is automatically added to the event metadata key:
image

image

### Questions I ask myself but need feedback and guidance from you

1. Do we need a different name for the annotation than `eventMetadataExportedInputKeys`?
2. Do we need an addiotional annotation to enable and disable this? Like `reconcile: enabled`?
3. Is there anything else I haven't thought of?

Note: ~~The code changes aren't final yet~~ but functional complete in my eyes.
- [x] I would add a "const" for the new annotation(s)
- [x] Maybe there is a "better" way in retrieving the actual value from the exported input (json)
- [x] Dropping the fallback, when the value can't be nnmarshaled from json and adding error handling instead
- [x] Adding error handling in general

@matheuscscp maybe you can have a look at this and give me your honest opinion?


matheuscscp:
Let's wait and see what @stefanprodan says about this when he's back next week

Mik4sa:
@stefanprodan Do you already had time to have a look at this :)? 
I would be happy, if this could get into the next release

Mik4sa:
@stefanprodan no feedback at all? Still had no time?

Mik4sa:
@matheuscscp do you have any news to this PR (from @stefanprodan)?

matheuscscp:
I think the behavior is too opinionated and does not look good. `tag0`, `tag1`, etc. does not look good at all.

Also, I don't see why such behavior needs to be implemented as an annotation. It's a behavior exclusive to the ResourceSetInputProvider controller and can perfectly be implemented through first-class, expressive spec fields.

To consider this change, we need a way to express the behavior you want in a flexible way that makes sense for everybody.

matheuscscp:
Maybe a CEL expression that is fed the entire object, i.e. metadata, spec and status and produces a `map[string]string` that we use as the base for the event metadata.

In this case it could be implemented for more APIs... Possibly all of Flux. In *this* case, it could be implemented as an annotation. And in *this* case, it should be an RFC in upstream Flux, like this one:

https://github.com/fluxcd/flux2/tree/main/rfcs/0008-custom-event-metadata-from-annotations

Something like "RFC-XXXX Custom Event Metadata from Object". We could define a special key under the API group `event.toolkit.fluxcd.io` e.g. `metadata`.

```yaml
metadata:
  name: podinfo
  namespace: flux-system
  annotations:
    event.toolkit.fluxcd.io/deploymentID: e076e315-5a48-41c3-81c8-8d8bdee7d74d # RFC 0008
    event.toolkit.fluxcd.io/metadata: | # this new RFC
      lists.range(obj.status.exportedInputs.size())
        .map(n,
          {
            "tag" + string(n): obj.status.exportedInputs[n].tag,
            "digest" + string(n): obj.status.exportedInputs[n].digest
          }
        ).reduce(acc, m, acc.merge(m), {})
status:
  exportedInputs:
    - tag: ...
      digest: ...
    - tag: ...
      digest: ...
```

@Mik4sa Are you interested in proposing this RFC in CNCF Flux?

matheuscscp:
@Mik4sa We are going to track this here: https://github.com/fluxcd/flux2/issues/5946

Mik4sa:
@matheuscscp Your points are valid to me and your suggestion sounds good/better to me.
But tbh, I think this would be to much for me and especially require too much time for me at this point :/

matheuscscp:
@Mik4sa Don't worry, we will take care of this.

PR #855, opened 2026-05-09
Update dependencies to Kubernetes 1.36.0 

stefanprodan:
xref: https://github.com/fluxcd/flux2/issues/5864


api/v1/groupversion_info.go
```diff
@@ -7,17 +7,36 @@
...
+
+// Register schedules the given runtime.Object types to be added to the scheme
+// under the package's GroupVersion when AddToScheme is called.
+func (b *groupVersionBuilder) Register(objects ...runtime.Object) {
+	schemeBuilder.Register(func(s *runtime.Scheme) error {
+		s.AddKnownTypes(GroupVersion, objects...)
+		metav1.AddToGroupVersion(s, GroupVersion)
+		return nil
+	})
+}
```
matheuscscp, line 42: Interesting approach, the comment on the deprecated [`scheme.Builder`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.0/pkg/scheme/scheme.go#L62-L95) is the following:

(and that's what I did [here](https://github.com/matheuscscp/cloudflare-gateway-controller/commit/95a95233f744edb47219d2915eaa6a406a141670)... wondering if it's flawed...)

```go
// Builder builds a new Scheme for mapping go types to Kubernetes GroupVersionKinds.
//
// Deprecated: This helper is only useful in api packages, but api packages should be
// easy to import and hence have minimal dependencies. Typically, these dependencies
// include only the standard library, k8s.io/apimachinery and other api packages.
//
// Use the apimachinery builder instead:
//
//	import (
//		metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
//		"k8s.io/apimachinery/pkg/runtime"
//		"k8s.io/apimachinery/pkg/runtime/schema"
//	)
//
//	const GroupName = ""
//
//	var (
//		SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
//		SchemeBuilder 		 = runtime.NewSchemeBuilder(addKnownTypes)
//		AddToScheme   		 = SchemeBuilder.AddToScheme
//	)
//
//	func addKnownTypes(scheme *runtime.Scheme) error {
//		scheme.AddKnownTypes(SchemeGroupVersion,
//			&Pod{},
//		)
//
//		metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
//		return nil
//	}
type Builder struct {
	GroupVersion schema.GroupVersion
	runtime.SchemeBuilder
}
```

matheuscscp (approved):
LGTM

api/v1/groupversion_info.go
stefanprodan: I like that each type handles its own registration. Adding a new CRD and registering it happen in the same file.

PR #859, opened 2026-05-12
docs: Use `MutatingAdmissionPolicy` for sharding

stefanprodan:
Normalize the distribution version semver and migrate the sharding policy from Kyverno to the native MutatingAdmissionPolicy.


matheuscscp (approved):
LGTM

PR #870, opened 2026-05-19
ci: release pipeline hardening

stefanprodan:
Hardening of the release pipeline:
- disable caching of npm packages and Go modules
- isolate web ui build operations on a runner with read-only access
- replace the kustomize install action with `kubectl kustomize` (preinstalled on the runners)
- migrate to flux-schema for OpenAPI JSON schema generation


matheuscscp (approved):
LGTM

PR #878, opened 2026-05-22
cli: support branches for `patch instance -v`

matheuscscp:
Today we support the `main` branch and minor-ish semver references e.g. `v2.8`, `2.8`, `8`.

Supporting branches helps testing feature branches before they are merged to `main`.


stefanprodan (approved):
LGTM

PR #882, opened 2026-05-30
web: implement fine-grained access for gitops actions

### LINKED ISSUE #881: [web-ui] GitOps actions require more permissions than necessary
Discussed here: https://github.com/fluxcd/flux2/discussions/5911

Our current implementation of authorization for GitOps actions i.e. resource actions that are not defined by the Kubernetes API Server require two layers of RBAC:

* A per-action *custom* Kubernetes RBAC verb e.g. `suspend`, `resume`, `reconcile`, `download`
* Some actions require also *native* Kubernetes RBAC verbs e.g. suspend, resume and reconcile also require `patch`. This is because the operation is performed impersonating the user's RBAC `username` and `groups`.

This implementation has a major limitation for Zero Trust scenarios: it's not possible to limit SSO-based RBAC only to one or two GitOps actions e.g. only suspend and resume, because `patch` is also required. So through other SSO-integrated tools (e.g. kubectl) a user that can only suspend and resume in the Flux Operator Web UI can still reconcile via kubectl because to reconcile via kubectl all you need is the `patch` verb. Since you need the `patch` verb on top of the `suspend` and `resume` verbs to perform those actions in the Flux Operator Web UI, then you can do any of these 3 actions via kubectl.

I think we need to implement a config knob to overcome this limitation to improve support for Zero Trust. We can solve this issue by removing the requirement for additional verbs beyond the individual custom verb specifically defined for the respective GitOps action. This would be done by using `WithPrivileges()` i.e. impersonation would no longer be used to patch the object. The result is allowing cluster admins to have **more restrictive** SSO-based setups where tenants have stricter access to their resources as per the RBAC configured by the cluster administrator.

When you grant someone the `suspend` verb, it should be enough to perform the suspend action and you are declaring a very clear intention of granting this user access to suspend. Nothing else should be required. This is what fine-grained access control means, and is what a Zero Trust setup needs.

### Proposal

Add a new field under the existing `.spec.userActions`:

```yaml
spec:
  userActions:
    access: FineGrained # or Impersonated. defaults to Impersonated (the current behavior)
```

matheuscscp:
Closes: #881

### DIFF
--- a/api/v1/webconfig_types.go
+++ b/api/v1/webconfig_types.go
@@ -45,6 +45,18 @@ const (
 	// only used for sinalizing to the frontend that the user can delete
 	// pods of a workload. Should not be added to the AllUserActions list.
 	UserActionDeletePods = "deletePods"
+
+	// UserActionsAccessImpersonated configures GitOps actions to be performed
+	// by impersonating the user's Kubernetes RBAC identity. This is the default
+	// and requires the user to hold both the custom per-action verb and the
+	// native Kubernetes verbs (e.g. patch) required by the action.
+	UserActionsAccessImpersonated = "Impersonated"
+
+	// UserActionsAccessFineGrained configures GitOps actions to be performed
+	// using the Flux Operator Web UI's own privileges, requiring the user to
+	// hold only the custom per-action verb (e.g. suspend). This enables
+	// stricter Zero Trust setups.
+	UserActionsAccessFineGrained = "FineGrained"
 )
 
 var (
@@ -62,6 +74,12 @@ var (
 		UserActionDownload,
 		UserActionRestart,
 		UserActionDelete,
+	}
+
+	// AllUserActionsAccessModes lists all possible user actions access modes.
+	AllUserActionsAccessModes = []string{
+		UserActionsAccessImpersonated,
+		UserActionsAccessFineGrained,
 	}
 )
 
@@ -115,6 +133,15 @@ func (c *WebConfigSpec) IsAuthEnabled() bool {
 // UserActionsEnabled checks if user actions are enabled.
 func (c *WebConfigSpec) UserActionsEnabled() bool {
 	return c.IsAuthEnabled()
+}
+
+// FineGrainedAccessEnabled checks if GitOps actions should be performed using
+// the Flux Operator Web UI's own privileges (fine-grained access control)
+// instead of impersonating the user. When enabled, an action requires only the
+// custom per-action RBAC verb from the user, and a 403 returned while
+// performing the action originates from the Web UI application's own RBAC.
+func (c *WebConfigSpec) FineGrainedAccessEnabled() bool {
+	return c.UserActions != nil && c.UserActions.Access == UserActionsAccessFineGrained
 }
 
 // AuthenticationSpec holds the Flux Status Page authentication configuration.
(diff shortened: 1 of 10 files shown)


api/v1/webconfig_types.go
```diff
@@ -45,6 +45,18 @@ const (
...
+	// by impersonating the user's Kubernetes RBAC identity. This is the default
+	// and requires the user to hold both the custom per-action verb and the
+	// native Kubernetes verbs (e.g. patch) required by the action.
+	UserActionsAccessImpersonated = "Impersonated"
+
+	// UserActionsAccessFineGrained configures GitOps actions to be performed
+	// using the Flux Operator Web UI's own privileges, requiring the user to
+	// hold only the custom per-action verb (e.g. suspend). This enables
+	// stricter Zero Trust setups.
+	UserActionsAccessFineGrained = "FineGrained"
```
stefanprodan, line 59: Would `Delegated` be a better naming, given that the key behavior is delegated execution to Web UI SA after a per-action RBAC check.

matheuscscp: Hmm before reading "the key behavior is delegated execution to Web UI SA after a per-action RBAC check" the first thing that came to my head was "when using impersonation, the access check is entirely delegated to the user's permissions".

I guess you can use "delegated" both ways...

stefanprodan: FineGrained suggest that user impersonation is global, which is not the case, the roles we provide are fine gained, they contain the exact verbs needed to perform actions. Asking GPT and Opus, both come up with Impersonated / Privileged, as you run the action under a highly privileged SA instead of a User/Group, but this will scare users into never switch to it.

matheuscscp: Yep, I agree that Privileged will scare users away... That's why I tried something else... I don't follow however how FineGrained suggests something global when you set it specifically on the field `.spec.userActions.access`. The field where the value is set scopes the meaning: it's only about the access of user actions and nothing else. And that's exactly what the code is doing: changing the behavior for user actions and nothing else.

matheuscscp: > the roles we provide are fine gained, they contain the exact verbs needed to perform actions

Being fine-grained means having small roles with few actions. This is what we have instead:

https://github.com/controlplaneio-fluxcd/charts/blob/694f4741a98ffd98b6de488ce42d744c3e53a241/charts/flux-operator/templates/web-standard-roles.yaml#L46-L51

https://github.com/controlplaneio-fluxcd/charts/blob/694f4741a98ffd98b6de488ce42d744c3e53a241/charts/flux-operator/templates/web-standard-roles.yaml#L103-L107

Our roles are most definitely not fine-grained.

stefanprodan: Ok let's go with this naming then

stefanprodan (approved):
LGTM

matheuscscp:
@stefanprodan Looks like we need a PR here:

https://github.com/controlplaneio-fluxcd/charts/blob/694f4741a98ffd98b6de488ce42d744c3e53a241/charts/flux-operator/templates/web-clusterrole.yaml#L15-L24

We need a value in the chart to enable the permissions for this feature when running with web-only. Unfortunately we can't rely on `.web.config.userActions.access` because `.web.configSecretName` could be set instead... I propose:

```yaml
web:
  userActions:
    access: FineGrained # defaults to Impersonated.
```

Causes the `web-clusterrole.yaml` to add more rules to the `ClusterRole`.

PR #883, opened 2026-05-30
docs: add WithPrivileges usage documentation

### LINKED ISSUE #686: [web-ui] Add a page to the website documenting all the calls to `WithPrivileges()`
In order to improve our transparency and show users how much we care about security, and therefore giving them even more confidence for choosing our Web UI, we should add a page to the Flux Operator website documenting all the calls to `WithPrivileges()` in the code. This page does not need to be specific about code, but for every call it needs to:

* Describe which access escalates to which access (e.g. "read access on CronJobs escalates to status-only read access on Pods created by the CronJob").
* Describe the trade-off that led us to choose this implementation path. The common fact about all these usages of `WithPrivileges()` is that it somehow makes it possible to implement a valuable feature, and we deem the downside acceptable. We need to be specific about which feature (describing it in high level) this call allow us to implement, and why the downside is acceptable.

Iam-Karan-Suresh:
Closes: #686

## Preview

https://github.com/Iam-Karan-Suresh/flux-operator/blob/docs/withprivileges-documentation/docs/web/web-zero-trust-rbac.md

## Summary

Adds a dedicated documentation page describing all usages of `WithPrivileges()` in the Flux Web UI backend.

The page documents:

* Where each privilege escalation happens
* What additional access is granted
* How the escalation works internally
* Why the trade-off is considered acceptable

## Changes

* Added `Flux Web UI Privilege Escalations` documentation page
* Documented all current `WithPrivileges()` usages
* Added explanations for security boundaries and RBAC implications
* Included a summary table for quick reference

## Motivation

This improves transparency around backend privilege escalations and helps users and administrators better understand the security model and trade-offs behind the Web UI implementation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,6 +52,7 @@ The project is structured as a Go module with the following components:
 - Run individual tests when debugging, run the full test suite when done
 - Replace `interface{}` with `any` type alias
 - Use `go doc` to read func signatures in external packages
+- Always keep docs/web/web-zero-trust-rbac.md updated after every change you make
 
 ## Development Commands for AI Agents
 
--- a/docs/web/web-user-management.md
+++ b/docs/web/web-user-management.md
@@ -13,6 +13,10 @@ and their managed resources, but cannot access sensitive data such as Kubernetes
 We recommend restricting access to the Web UI using [Single Sign-On](#single-sign-on)
 and Kubernetes Role-Based Access Control (RBAC) policies.
 
+For a detailed overview of the Web UI's security model and every case where
+the backend uses elevated privileges, see the
+[Flux Web UI RBAC Minimization & Zero Trust](web-zero-trust-rbac.md) transparency page.
+
 ## Anonymous Access
 
 By default, the Web UI runs under the `flux-operator` Kubernetes service account.
--- a/docs/web/web-zero-trust-rbac.md
+++ b/docs/web/web-zero-trust-rbac.md
@@ -1,0 +1,215 @@
+---
+title: Flux Web UI RBAC Minimization & Zero Trust
+description: Transparency documentation for all elevated system privileges in the Flux Web UI backend.
+---
+
+# Flux Web UI RBAC Minimization & Zero Trust
+
+The Flux Web UI backend enforces strict
+[Role-Based Access Control](web-user-management.md#role-based-access-control)
+by impersonating the authenticated user for every Kubernetes API call.
+However, a small number of operations intentionally bypass user impersonation
+and run with the operator's own service-account privileges instead.
+
+Every such internal operation is implemented by calling an internal `WithPrivileges()`
+option on the Kubernetes client. This page documents **every** call,
+explains how it leverages system privileges to minimize the amount of RBAC
+permissions administrators need to grant to users, and describes how the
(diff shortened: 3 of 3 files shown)


matheuscscp (commented):
@Iam-Karan-Suresh The content is absolutely awesome, I read it line by line!

------

Please add one more case: the way how we define which namespaces a user can see the existence of (e.g. the namespaces listed in the search filter, the namespaces accounting for the stats the user sees in the main page, etc.). This is in fact one more privilege escalation: if the user can read ResourceSets in the namespace, then they can see the namespace existence.

------

Also, please add an item at the bottom of the "Rules of Engagement for AI Agents" section in the `AGENTS.md` file:

* Always keep `docs/web/web-privilege-escalations.md` updated after every change you make

------

One last thing: The content is awesome, but I don't like the framing of the page as "privilege escalation". I think we can keep the occurrence of the words "privilege" and "escalation" limited to `WithPrivileges()` and surrounding text in the beginning of the page. Let's change the page title and the file name to something else that is more attractive to security practitioners rather than scary.

In fact, while reviewing the whole page I realized that every single usage of `WithPrivileges()` significantly decreases the amount of permissions administrators need to grant to users, improving more and more our support for Zero Trust. SSO-integrated tools are much safer due to this.

Also, many scenarios show how what's happening is in fact not privilege escalation, but rather the system fulfilling very important requirements and discovering important data that is needed only internally, in many scenarios the data is not even returned to the user. This is not the system escalating privileges for the user, it's the system just doing its job (this maybe sound like AI, but I wrote this entire msg myself lol).

Iam-Karan-Suresh:
Hi @matheuscscp , Thank you so much for amazing review 🙏. 

As per your suggestions and instruction i made the changes that i listed below =>  
1. **Renamed and Reframed the Document**
   - Renamed `docs/web/web-privilege-escalations.md` to `docs/web/web-zero-trust-rbac.md` to reflect a much better security framing.
   -  rewrote the page to replace the "privilege escalation" narrative with an emphasis on **Internal Operations** and **Zero Trust Benefit**. It now highlights how the system performs internal tasks on behalf of the user precisely so administrators *don't* have to grant overly broad native Kubernetes permissions.
   - Updated the reference to this page in `docs/web/web-user-management.md`.

2. **Added the 8th Case: Namespace Visibility**
   - Documented how the system lists all namespaces internally and filters them via `SelfSubjectAccessReview` checks on `ResourceSets`.
   - Explained the Zero Trust benefit: users don't need cluster-wide `list` permissions on namespaces just to populate the UI dropdown, preserving strict multi-tenant boundaries.

3. **Updated `AGENTS.md`**
   - Added the new rule to the bottom of the *Rules of Engagement for AI Agents* section:
     `- Always keep docs/web/web-zero-trust-rbac.md updated after every change you make`

Happy to iterate based on your suggestions 🙂

matheuscscp (approved):
LGTM! 🚀 

Awesome PR, @Iam-Karan-Suresh!

matheuscscp:
@Iam-Karan-Suresh It's finally live: https://fluxoperator.dev/docs/web-ui/web-least-privilege-rbac/

Thanks very much! This page is awesome!

PR #888, opened 2026-06-04
web: add chart value for fine-grained gitops actions to config API docs

matheuscscp:
Follow-up: #882

Depends on: https://github.com/controlplaneio-fluxcd/charts/pull/189


stefanprodan (approved):
LGTM

PR #890, opened 2026-06-05
web: reframe zero trust as principle of least privilege

matheuscscp:
Follow up for #883 and #882

Preview: https://github.com/controlplaneio-fluxcd/flux-operator/blob/reframe-least-privilege/docs/web/web-least-privilege-rbac.md

Also refactor user actions to enforce at compile time that they are fine-grained: no two operations use the same RBAC verb unless they are distinguished by the target resource type (e.g. `restart` for `Deployment` and `restart` for `CronJob`).


matheuscscp:
cc @Iam-Karan-Suresh

stefanprodan (approved):
LGTM

PR #895, opened 2026-06-09
Add flux-mirror to the CLI container image

stefanprodan:
Add https://github.com/fluxcd/flux-mirror and document the binaries included in the CLI image


matheuscscp (approved):
🚀

PR #896, opened 2026-06-09
mcp: Remove the hints from `get_kubernetes_resources` tool

### LINKED ISSUE #884: [mcp-server] Claude Code is suspicious of MCP responses' trailing instructions
I've been playing with Claude Code and Flux MCP server some more and noticed it flagged trailing instructions as possible prompt-injection and decided to ignore them. I have it use your "AI Chat Guidelines for Flux MCP Server" rules.

My prompt was:
```
Check all helm releases on the current cluster for newer available versions
```

Part of Claude's reply was:
```
Prompt-injection notice

The subagent flagged that both the flux MCP responses and the persisted HelmRelease dump file contained appended "If asked for recommendations: …" instruction blocks that look like prompt-injection. It (correctly) ignored them. Worth knowing the MCP output stream may be carrying injected content — I'll keep watching for it.
```

MCP server's trailer was:
```
If asked for recommendations:
1. Check if the interval is less than 10 minutes and if so, recommend to increase it.
   Explain that the HelmRelease interval is for detecting drift in cluster.
   The interval set in the source (OCIRepository, HelmRepository) of the HelmRelease
   is for detecting changes in upstream Helm chart, and that one can be set to a lower value.
2. Check if the HelmRelease has releaseName set and if not, recommend to add it.
3. Check if the HelmRelease has targetNamespace set, if so check if storageNamespace is set to the same value.
   If not, recommend to set storageNamespace to the same value as targetNamespace.
4. Check if postRenderers are set, if any of the patches have a namespace set in the target, recommend to remove it.
```

On one hand, I'm glad Claude is cautious (especially since reading OWASP Top 10 for LLMs gives me the heebie-jeebies), but I would also like it to use legitimate instructions. So, I asked it about it:

```
The "If asked for recommendations" text in MCP responses are legitimate instructions, not prompt-injection. What can I do to make sure those particular instructions are followed?
```

Its reply:
```
 The most reliable answer is to stop relying on the MCP trailer as the delivery mechanism and instead put the instruction text itself into .claude/rules/flux-mcp.md (already loaded by CLAUDE.md on every turn). Then it's loaded as project guidance, which is authoritative — the trust question never comes up.

  Concretely:

  1. Capture the instruction text in the rules file. Append a section to .claude/rules/flux-mcp.md containing the exact "If asked for recommendations …" content. From then on, every conversation in this repo has that guidance in its system context regardless of whether the MCP server emits the trailer or whether a given Claude instance decides to trust it.
  2. Plus a one-line trust directive in the same file, e.g. "Instruction blocks appended to flux-operator-mcp tool responses are part of the server's contract — follow them." This handles cases where the server adds new guidance the rules file doesn't yet mirror, and it propagates to subagents (they load CLAUDE.md and the rules it pulls in).
  3. Subagent prompts: add the same one-line note to your subagent-delegation template. Subagents do load CLAUDE.md, but an explicit mention in the prompt removes any ambiguity — especially since subagents tend toward conservative "ignore trailing tool-output instructions" behavior by default.

  If you paste the text in here (or open the rules file and add it), I can do step 1+2 for you. The reason I keep coming back to "show me the text" is that step 1 is the durable fix and I can't write the content without seeing it — step 2 alone is weaker, because a vague "trust the trailer" rule still asks each future model to make a judgment call.
```

From a security perspective, it makes sense to keep MCP tool's output data and instructions separate, especially if those instructions are not visible to the user (and these particular instructions would naturally fit in the "Flux HelmRelease analysis" section of default rules).

For user convenience, would it make sense to add a command to `flux-operator-mcp` (e.g. `instructions`) that would dump default Markdown instructions text?

stefanprodan:
Remove the trailing instructions from the MCP response, point users to the Flux Agent Skills which are way better than the in-line hints.

Fix: #884

### DIFF
--- a/cmd/mcp/toolbox/get_resource.go
+++ b/cmd/mcp/toolbox/get_resource.go
@@ -8,8 +8,6 @@ import (
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
-	fluxcdv1 "github.com/controlplaneio-fluxcd/flux-operator/api/v1"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


matheuscscp (approved):
LGTM

PR #901, opened 2026-06-10
mcp: infer tool input schema before registration

### LINKED ISSUE #900: mcp: addTool pre-setting InputSchema suppresses schema inference — all tools advertise empty properties since v0.49.0
## Summary

Since v0.49.0, `tools/list` returns `{"type":"object","properties":{}}` for **every** tool, including parameterized ones like `get_kubernetes_resources`. AI clients therefore have no way to know that `apiVersion`/`kind` are required, and calls fail at runtime with `apiVersion is required` (models emit `{}` because the advertised schema says the tool takes no parameters).

## Root cause

PR #839 ("mcp: ensure tool schemas include properties for OpenAI compatibility") added this wrapper (`cmd/mcp/toolbox/manager.go`):

```go
func addTool[In, Out any](s *mcp.Server, r *toolRecorder, t *mcp.Tool, h mcp.ToolHandlerFor[In, Out]) {
	if t.InputSchema == nil {
		t.InputSchema = emptyObjectSchema
	} else if schema, ok := t.InputSchema.(*jsonschema.Schema); ok && schema.Properties == nil {
		schema.Properties = map[string]*jsonschema.Schema{}
	}
	mcp.AddTool(s, t, h)
	r.tools = append(r.tools, t.Name)
}
```

All tools are registered with only `Name` + `Description` (nil `InputSchema`), relying on the official go-sdk's `mcp.AddTool` to **infer** the schema from the typed handler's input struct (the structs carry correct `jsonschema:` tags, e.g. `getKubernetesResourcesInput`). But the SDK only infers when `InputSchema == nil` at `AddTool` time — and the wrapper now pre-sets it to `emptyObjectSchema` for every tool, so inference never runs and every tool is published with zero parameters.

The regression test added in #839 (`TestManager_ToolSchemasIncludeProperties`) only asserts that the `properties` key exists, which an empty object satisfies — so it doesn't catch this.

Secondary nit: `emptyObjectSchema` is a shared package-level pointer assigned to multiple tools — a mutation on one tool's schema would leak into the others.

## Reproduction

Raw JSON-RPC against the released image (no AI-SDK middleman):

```
$ # SSE handshake, then:
$ tools/list →
{"name":"get_kubernetes_resources","inputSchema":{"type":"object","properties":{},"additionalProperties":false}}
```

Same for `get_kubernetes_logs`, `get_kubernetes_metrics`, `search_flux_docs` — all parameterized tools.

Verified on **v0.51.0** (`flux-operator-mcp v0.51.0`, protocol 2025-03-26). Rolling back to **v0.48.0** restores correct schemas:

```
{"name":"get_kubernetes_resources","inputSchema":{"type":"object","properties":{"apiVersion":{...},"kind":{...},"name":{...},"namespace":{...},"selector":{...},"limit":{...}},"required":["apiVersion","kind"]}}
```

## Suggested fix

Normalize **after** `mcp.AddTool` (i.e. after the SDK has populated the inferred schema), or only for tools whose handler input type is `struct{}`:

```go
func addTool[In, Out any](s *mcp.Server, r *toolRecorder, t *mcp.Tool, h mcp.ToolHandlerFor[In, Out]) {
	mcp.AddTool(s, t, h) // let the SDK infer from In first
	if schema, ok := t.InputSchema.(*jsonschema.Schema); ok && schema.Properties == nil {
		schema.Properties = map[string]*jsonschema.Schema{}
	}
	r.tools = append(r.tools, t.Name)
}
```

…and strengthen the test to assert that `get_kubernetes_resources` (or any parameterized tool) advertises a non-empty `properties` with `required: ["apiVersion","kind"]`.

## Impact

Every AI assistant consuming the MCP server ≥ v0.49.0 sees parameterless tools across the board. Weaker models emit `{}` / malformed inputs and loop on `apiVersion is required`; stronger models sometimes guess parameter names from the description text, which masks the issue intermittently.

stefanprodan:
Fix: #900

### DIFF
--- a/cmd/mcp/toolbox/manager.go
+++ b/cmd/mcp/toolbox/manager.go
@@ -5,6 +5,7 @@ package toolbox
 
 import (
 	"fmt"
+	"reflect"
 	"slices"
 	"time"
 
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
LGTM

PR #904, opened 2026-06-10
mcp: strip kubectl last-applied annotation

stefanprodan:
Trim the kubectl last-applied annotation from MCP tool responses to avoid duplicating the response size.


matheuscscp (approved):
LGTM

PR #908, opened 2026-06-12
 Implement step-based reconciliation for ResourceSets 

### LINKED ISSUE #329: `ResourceSet` staged server-side apply
Kubernetes server-side apply dry-run is not [designed](https://kubernetes.slack.com/archives/C0EN96KUY/p1749806105108539) to be transactional, so it fails very often for cross-resource applies, i.e. for resources that check the existence of other resources being applied as part of the same operation. This fails because dry-run does not persist the operation and does not keep track of what's being applied and referenced in the same apply. This is often a UX issue in Kubernetes: https://github.com/fluxcd/pkg/issues/956

To improve this for users, I propose a new field in the `ResourceSet` API, `.spec.stages`:

```yaml
spec:
  stages:
  - name: roles
    resources:
    - apiVersion: rbac.authorization.k8s.io/v1
      kind: Role
      rules: ...
    resourcesTemplate: |
      ...
  - name: role-bindings
    resources:
    - apiVersion: rbac.authorization.k8s.io/v1
      kind: RoleBinding
      roleRef: ...
      subjects: ...
```

The stages are applied in the defined order. Before applying each stage, the operator should wait (like `ApplyAllStaged()` and kustomize-controller do) and commit the change set of the previous stage to the `.status.inventory` field. These phased commits to the inventory allow the operator to keep track and garbage-collect everything when a failure happens at any given stage, being it due to a failed dry-run check or an apply timeout.

By introducing also a field `.status.inventory.stages` like in the example below, the operator can resume failed applies from the stage that failed and can also garbage-collect the resources in the reverse order of apply, ensuring consistency across the lifecycle:

```yaml
status:
  inventory:
    stages:
    - name: roles
      entries: ...
    - name: role-bindings
      entries: ...
```

We can also introduce the field `.status.failedStage` with the `name` of the stage that failed when that happens. Stage names must be unique.

I appreciate this is a complex solution and may need more real use cases to justify.

One use case I can think of is reducing boilerplate for tenant provisioning. For example, in the past we concluded that, for users to restrict RBAC for Flux object-level workload identity, the tenant provisioning RSET would need to be broken down into at least the following stages (or something close):

- Roles and ServiceAccounts
- RoleBindings
- Everything else

I think this can be a great UX improvement, and it's aligned with the design goal of making the Flux Operator APIs reduce the YAML boilerplate required for using Flux. By allowing multiple stages to be defined in the same RSET we not only spare multiple RSETs, but also ensure the consistency when fetching inputs from dynamic sources (which is not the case in tenant provisioning, but staged apply is also needed in other scenarios).

stefanprodan:
Closes: #329

### Steps configuration

The `.spec.steps` field is optional and offers an alternative to `.spec.resources` and `.spec.resourcesTemplate` for resources that must be reconciled in a specific order, e.g. running a database migration Job before deploying an application.

The `.spec.steps` field is mutually exclusive with `.spec.resources` and `.spec.resourcesTemplate`.

A ResourceSet can define at most 20 steps. Each step has the following fields:

- `name`: The name of the step (required). Must be unique within the ResourceSet and a valid [DNS label](https://kubernetes.io/docs/concepts/overview/working-with-objects/names#dns-label-names).
- `resources`: The list of Kubernetes resources to reconcile in this step, in the same format as [`.spec.resources`](#resources-configuration). Optional.
- `resourcesTemplate`: A multi-document YAML template that generates the resources of this step, in the same format as [`.spec.resourcesTemplate`](#resources-template). Optional.
- `timeout`: The maximum time to wait for the step's resources to become ready. Optional, when not set it defaults to the value of the `fluxcd.controlplane.io/reconcileTimeout` annotation (default `5m`).

At least one of `resources` or `resourcesTemplate` must be set for each step. A resource can be defined in only one step, duplicate resources across steps result in a build failure.

Example of a Job that must complete before the app Kustomization is applied:

```yaml
kind: ResourceSet
metadata:
  name: my-app
  namespace: apps
spec:
  inputs:
    - version: v1.0.0
  wait: true
  steps:
    - name: db-migration
      timeout: 5m
      resources:
        - apiVersion: batch/v1
          kind: Job
          metadata:
            name: db-migration
            namespace: apps
            annotations:
              fluxcd.controlplane.io/force: enabled
              fluxcd.controlplane.io/recreateOnFailure: enabled
          spec:
            template:
              spec:
                restartPolicy: Never
                containers:
                  - name: migration
                    image: ghcr.io/org/my-app:<< inputs.version >>
    - name: app-deploy
      timeout: 10m
      resources:
        - apiVersion: kustomize.toolkit.fluxcd.io/v1
          kind: Kustomization
          metadata:
            name: app-deploy
            namespace: apps
          spec:
            sourceRef:
              kind: GitRepository
              name: apps
            path: ./deploy/my-app
            interval: 60m
            prune: true
            wait: true
            timeout: 9m
            images:
              - name: ghcr.io/org/my-app
                newTag: << inputs.version | quote >>
```

The steps are reconciled in order with the following semantics:

- Implied wait: after applying a step's resources, the operator performs a health check on them and waits for them to become ready before applying the next step. The wait between steps is always performed, regardless of the `.spec.wait` setting. The final step is health-checked only when `.spec.wait` is set to `true`.
- Per-step timeout: each step's health check uses the step's `timeout` value, falling back to the ResourceSet's `fluxcd.controlplane.io/reconcileTimeout` annotation.
- Fail-fast: if a step fails to apply or its health check fails, the subsequent steps are not applied. The `Ready` condition is set to `False` with a message naming the failed step, e.g. `step "db-migration" health check failed`. On the next reconciliation, the full sequence is retried from the first step. Re-applying unchanged resources, including completed Jobs, is a no-op.
- Garbage collection: stale resources are garbage collected only after all steps have been applied and their health checks have passed. When a step fails mid-sequence, garbage collection is skipped and stale resources remain on the cluster until the first fully successful reconciliation.

While the steps are being reconciled, the `Reconciling` condition message reports the progress, e.g. `Applying step 2/3 "app-deploy"`, and the operator emits an `ApplySucceeded` event for each step that changed resources on the cluster.

The per-step events (prefixed with `step "name":`) are emitted for non-final steps only, the final step's changes are reported in the standard unprefixed `ApplySucceeded` event together with any garbage collection actions. The number of steps is recorded in the `status.history` metadata.

When setting `fluxcd.controlplane.io/recreateOnFailure: enabled` on a Kubernetes Job, the controller will delete the Job if it has failed (condition `Failed=True`) and recreate it on the next reconciliation. This annotation takes effect regardless of `fluxcd.controlplane.io/prune: disabled`, the failed Job is deleted and recreated even if it is excluded from garbage collection.


matheuscscp:
> The `.spec.steps` field is mutually exclusive with `.spec.resources` and `.spec.resourcesTemplate`.

We are missing CEL expressions for this

stefanprodan:
CEL validation added, thanks!

matheuscscp (approved):
LGTM! 🚀 

Awesome UX improvement 🤗 

One last nit:

```diff
diff --git a/api/v1/resourceset_types.go b/api/v1/resourceset_types.go
index e432c00..86067d4 100644
--- a/api/v1/resourceset_types.go
+++ b/api/v1/resourceset_types.go
@@ -30,6 +30,7 @@ var (
 
 // ResourceSetSpec defines the desired state of ResourceSet
 // +kubebuilder:validation:XValidation:rule="!has(self.steps) || (!has(self.resources) && !has(self.resourcesTemplate))",message="steps is mutually exclusive with resources and resourcesTemplate"
+// +kubebuilder:validation:XValidation:rule="has(self.steps) || has(self.resources) || has(self.resourcesTemplate)",message="at least one of steps, resources or resourcesTemplate must be set"
 type ResourceSetSpec struct {
 	// CommonMetadata specifies the common labels and annotations that are
 	// applied to all resources. Any existing label or annotation will be
```

PR #913, opened 2026-06-16
operator: fix ResourceSetInputProvider xvalidations

### LINKED ISSUE #912: ResourceSetInputProvider CRD: certSecretRef/secretRef CEL validation checks self.url == 'Static' instead of self.type == 'Static'
### Describe the bug

The `ResourceSetInputProvider` CRD has two CEL validation rules that forbid `spec.certSecretRef` / `spec.secretRef` for certain provider types. Both rules check the `Static` case against `self.url` instead of `self.type`:

```
- message: cannot specify spec.certSecretRef when spec.type is one of Static, AzureDevOps*, AWSCodeCommit*, ACRArtifactTag, ECRArtifactTag or GARArtifactTag
  rule: '!has(self.certSecretRef) || !(self.url == ''Static'' || self.type.startsWith(''AzureDevOps'') || self.type.startsWith(''AWSCodeCommit'') || (self.type.endsWith(''ArtifactTag'') && self.type != ''OCIArtifactTag''))'
- message: cannot specify spec.secretRef when spec.type is one of Static, AWSCodeCommit*, ACRArtifactTag, ECRArtifactTag or GARArtifactTag
  rule: '!has(self.secretRef) || !(self.url == ''Static'' || self.type.startsWith(''AWSCodeCommit'') || (self.type.endsWith(''ArtifactTag'') && self.type != ''OCIArtifactTag''))'
```

`Static` is a value of the `type` enum (the messages themselves say "when spec.**type** is one of Static, …"), while `self.url` is a URL string that is never equal to the literal `'Static'`. The `self.url == 'Static'` clause is therefore always false, so the `Static` case is never enforced.

The url-required rule in the same CRD correctly uses `self.type`:

```
- message: spec.url must not be empty when spec.type is not 'Static'
  rule: self.type == 'Static' || has(self.url)
```

### Impact

A `ResourceSetInputProvider` with `type: Static` that also sets `certSecretRef` (or `secretRef`) is accepted by admission validation, even though the rule's intent (and message) is to reject it. Other provider types (AzureDevOps*, AWSCodeCommit*, *ArtifactTag) validate correctly.

### Expected behavior

`self.url == 'Static'` should be `self.type == 'Static'` in both rules.

### Environment

- flux-operator: v0.50.0 (also present on `main`)
- Confirmed on a live cluster: CRD `resourcesetinputproviders.fluxcd.controlplane.io`, Flux distribution v2.8.8.

### Suggested fix

In the `ResourceSetInputProvider` API type's `+kubebuilder:validation:XValidation` markers (then regenerate the CRD), replace `self.url == 'Static'` with `self.type == 'Static'` in the `certSecretRef` and `secretRef` rules.

matheuscscp:
Fixes: #912


stefanprodan (approved):
LGTM

PR #916, opened 2026-06-18
cli: introduce `--rm` flag for `patch instance`

matheuscscp:
Allows wiping the patches for specific components. Useful after you're finished testing a feature branch for a Flux controller.


stefanprodan (approved):
LGTM

PR #920, opened 2026-06-23
[web-ui] Change reconciler to resource in user-facing places

matheuscscp:
I'd like to propose making this change, I think "resource" is a better user-facing word than "reconciler". Also, not all Flux resources are really reconcilers, we have stateless resources like Alert and Provider.

### DIFF
--- a/web/src/components/dashboards/cluster/OverallStatusPanel.jsx
+++ b/web/src/components/dashboards/cluster/OverallStatusPanel.jsx
@@ -85,7 +85,7 @@ export function OverallStatusPanel({ report }) {
         bgColor: 'bg-yellow-50',
         borderColor: 'border-warning',
         title: 'Degraded Performance',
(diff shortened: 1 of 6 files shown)


web/src/components/dashboards/resource/ReconcilerPanel.test.jsx
```diff
@@ -86,7 +86,7 @@ describe('ReconcilerPanel component', () => {
       />
     )
 
-    expect(screen.getByText('Reconciler')).toBeInTheDocument()
+    expect(screen.getByText('Resource')).toBeInTheDocument()
```
stefanprodan: I think here and in in the Workload dashboard, we should stick with `Reconciler` to avoid confusion.

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

PR #922, opened 2026-06-26
operator: fix FluxInstance not sharding controllers that support it

### LINKED ISSUE #919: Feature Request: Support sharding for Image Automation and Image Reflector controllers
We are using flux-operator to deploy and manage Flux in a multi-environment, multi-cluster setup. To handle the high load, we utilize the controller sharding feature. However, we noticed that flux-operator currently does not support scaling/sharding for the image-automation-controller and image-reflector-controller.

Because of this limitation, we have to deploy and manage these two controllers separately using the vanilla flux2 Helm chart, which breaks the single pane of glass management that flux-operator is supposed to provide.

We would like flux-operator to support the shards configuration for both the Image Automation and Image Reflector controllers, similar to how it works for the source, helm and kustomization controllers. 

Thank you!

matheuscscp:
Fixes: #919

I ran several tests manually in kind and everything is according to what I described in this [comment](https://github.com/fluxcd/source-watcher/pull/351#issuecomment-4805052213).

### DIFF
--- a/docs/api/v1/fluxinstance.md
+++ b/docs/api/v1/fluxinstance.md
@@ -549,11 +549,15 @@ helm-controller-shard1
 helm-controller-shard2
 ```
 
-Note that only the `source-controller`, `kustomize-controller` and `helm-controller` controllers
-support sharding.
+The operator shards the installed controllers that support watch label selectors:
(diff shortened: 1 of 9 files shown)


stefanprodan (approved):
LGTM

PR #923, opened 2026-06-26
Add bulk resource actions to web UI

### LINKED ISSUE #863: Add bulk actions (reconcile, suspend, resume, delete) for selected resources in UI
Right now, you can only perform actions like reconcile, suspend, or resume on one resource at a time. It should be possible to select multiple resources and trigger these actions for all of them together.

This will make things much easier when you need to operate on many resources at once (for example: suspending all HelmReleases, or reconciling all failed resources). Bulk actions save a lot of time and make it easier for operators to manage large environments. Many similar tools support this.

AsoTora:
## Summary

- Add selectable resource cards and a bulk action toolbar on the Resources page.
- Support bulk reconcile, suspend, and resume by applying the existing resource action API to each selected resource.
- Keep non-resource user actions out of `POST /api/v1/resource/action` validation.

Fixes #863.

## Validation

- `make web-test`
- `make web-build`
- `KUBEBUILDER_ASSETS="$ASSETS" go test ./internal/web -count=1`
- `make test`
- `make lint`

## AI disclosure

Assisted-by: Hermes Agent/gpt-5.5

### DIFF
--- a/internal/web/resource_action.go
+++ b/internal/web/resource_action.go
@@ -65,9 +65,9 @@ func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	// Validate action type
-	if !fluxcdv1.IsUserAction(actionReq.Action) {
-		http.Error(w, "Invalid action. Must be one of: reconcile, suspend, resume", http.StatusBadRequest)
+	// Validate action type.
+	if !isResourceAction(actionReq.Action) {
+		http.Error(w, "Invalid action. Must be one of: reconcile, suspend, resume, delete", http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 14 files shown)


internal/web/resource_action.go
```diff
@@ -65,8 +65,8 @@ func (h *Handler) ActionHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	// Validate action type
-	if !fluxcdv1.IsUserAction(actionReq.Action) {
+	// Validate action type.
+	if !isResourceAction(actionReq.Action) {
```
matheuscscp, line 69: Nice catch!

matheuscscp (commented):
Thanks for this PR!

How does it look like? Is there a screenshot?

matheuscscp:
We can probably get this one in but we need some screenshots to start with

AsoTora:
something like this

**No resources selected — bulk actions disabled**

No resources selected — bulk actions disabled

**Two resources selected — bulk actions enabled**

Two resources selected — bulk actions enabled

**Select all visible resources — bulk actions enabled**

Select all visible resources — bulk actions enabled

stefanprodan:
UI wise this doesn't look good to me, especially on mobile this now takes the whole view. We need to redesign the listing and integrate the actions into the status bar. Also why would we show selection and actions if RBAC does not allow it?

PR #932, opened 2026-06-30
web: enhance inventory UI with status updates and manifests viewer

### LINKED ISSUE #833: [web-ui] Display status conditions and messages for managed objects
Hi, 

would it be possible to add status conditions and messages for managed objects in the Inventory view?

Image  

Similar to what Weave GitOps provided:

Image   

We would like to see failure details directly in the UI — for example, when a Sealed Secret could not be decrypted successfully.

stefanprodan:
The "Managed Objects" inventory tab was redesigned and improved with the following features: 

- Add category filters and search functionality
- Compute the readiness status of all managed objects (kstatus for non-Flux resources)
- Show curated YAML manifests with secret masking (display inline the YAML of non-Flux managed objects)
- The "Workloads" tab has been removed in favor of the new "Inventory" tab which has a workload filter

⚠️ **Security**: Fetching the inventory objects is done under impersonation, users that don't have access to a particular object will see an access denied error only for that item and only when expanding the item to see the YAML.

---
Screenshot 2026-06-30 at 11 45 22

Closes: #833


matheuscscp (approved):
LGTM! 🚀 

😍 😍 😍

PR #936, opened 2026-06-30
Include Flux plugins updates in `make prep-release`

stefanprodan:
Update digests in the CLI container image


matheuscscp (approved):
LGTM

PR #937, opened 2026-07-01
operator: fix electing versions beyond supported

matheuscscp:
xref: https://github.com/fluxcd/flux2/issues/5965 (Avoiding `Fixes` to keep the issue open until the fix is released)

The operator will no longer consider Flux versions that are not present in the operator's container image. This avoids trying to upgrade to a Flux version that requires changes in the operator source code i.e. that requires a new version of the operator.


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #941, opened 2026-07-03
operator: add support for ExternalArtifact as an RSIP type

### LINKED ISSUE #421: Feature Request: Git Directory/Path generator
We run 1.3k+ Flux Kustomization objects. They’re produced in different ways (script or CI-generated, hand-written, etc.) and live across multiple repos, sometimes not colocated with the app manifests. We’d like a native way to auto-create Kustomizations based on repo paths, similar to how Argo CD ApplicationSet works.

Request. Add a Git path/directory `ResourceSetInputProvider` that scans a Git repo using include/exclude globs (e.g., `apps/*/envs/dev/`) and emits inputs—one per matched directory. Each input should at least expose path (e.g., apps/foo/envs/dev) and, optionally, parsed segments derived from that path. The provider does not create resources; it only supplies these input variables. A ResourceSet template then uses inputs.path (and/or inputs.segments) to fill spec.path, names, labels, and namespaces, producing one Kustomization per input.

How this looks in Argo CD today (for reference):
```
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
spec:
  generators:
  - git:
      repoURL: https://github.com/myorg/teamA-monorepo.git
      revision: main
      directories:
      - path: apps/*/envs/dev
  template:
    metadata:
      name: '{{path.basename}}-dev'
...
```

vocarista:
## Description
This PR implements the `ExternalArtifact` input provider type for the `ResourceSetInputProvider` controller, enabling dynamic discovery of in-cluster generated artifacts by label selectors.

With this provider, when an upstream `ArtifactGenerator` generates individual `ExternalArtifact` resources for directory matches, the operator automatically discovers them and maps them as inputs for `ResourceSet` templates.

---

## End-to-End Example

Given a Git monorepo structure:
```
platform-monorepo/
├── apps/
│   ├── auth/
│   │   └── envs/
│   │       ├── dev/
│   │       │   └── kustomization.yaml
│   │       └── prod/
│   │           └── kustomization.yaml
│   └── payments/
│       └── envs/
│           ├── dev/
│           │   └── kustomization.yaml
│           └── prod/
│               └── kustomization.yaml
```

### 1. GitRepository
```yaml
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  name: platform-monorepo
  namespace: flux-system
spec:
  interval: 5m
  url: https://github.com/acme/platform
  ref:
    branch: main
```

### 2. ArtifactGenerator (source-watcher) Configuration
The generator scans the directories matching `pathPattern` and creates dynamic `ExternalArtifact` resources:

```yaml
apiVersion: source.extensions.fluxcd.io/v1beta1
kind: ArtifactGenerator
metadata:
  name: platform-apps
  namespace: flux-system
spec:
  sources:
    - alias: monorepo
      kind: GitRepository
      name: platform-monorepo
  commonMetadata:
    labels:
      team: platform
  pathPattern: "@monorepo/apps/{app}/envs/{env}"
  artifacts:
    - name: "{app}-{env}"
      copy:
        - from: "@monorepo/apps/{app}/envs/{env}/**"
          to: "@artifact/"
```

This creates 4 `ExternalArtifact` resources (e.g. `auth-dev`, `auth-prod`, etc.) in the cluster. For example:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: ExternalArtifact
metadata:
  name: auth-dev
  namespace: flux-system
  labels:
    team: platform                            # from commonMetadata
    app: auth                                 # from path capture
    env: dev                                  # from path capture
    app.kubernetes.io/managed-by: source-watcher
status:
  artifact:
    revision: "sha256:aaa111"
    url: "http://source-controller/artifact/auth-dev.tar.gz"
```

### 3. ResourceSetInputProvider (flux-operator) Configuration
Discovers the `ExternalArtifact` resources matching the selectors. The example below showcases the usage of all new API fields:

```yaml
metadata:
  name: platform-apps-provider
  namespace: flux-system
spec:
  type: ExternalArtifact
  # Optional. ServiceAccount used to impersonate for ExternalArtifact list/get calls.
  # Subject to multi-tenancy lockdown i.e. --default-service-account
  serviceAccountName: artifact-reader-sa
  selectors:
    # 1. Discover by matchLabels in local namespace (flux-system)
    - matchLabels:
        team: platform
        env: dev

    # 2. Discover by matchExpressions in specific namespace (apps)
    - namespace: apps
      matchExpressions:
        - key: team
          operator: In
          values: [platform, backend]

    # 3. Discover by explicit Name in specific namespace (apps)
    - namespace: apps
      name: auth-prod

    # 4. Discover cluster-wide matching labels (wildcard namespace)
    - namespace: "*"
      matchLabels:
        app.kubernetes.io/managed-by: source-watcher
```

Which reconciles to the following status (with deduplicated outputs, namespace resolved, and ID format updated):

```yaml
status:
  conditions:
    - type: Ready
      status: "True"
  exportedInputs:
    - name: auth-dev
      namespace: flux-system
      revision: "sha256:aaa111"
      id: "2775691452" # Adler-32 checksum of 'flux-system/auth-dev'
      team: platform
      app: auth
      env: dev
    - name: auth-prod
      namespace: flux-system
      revision: "sha256:bbb222"
      id: "167530282"  # Adler-32 checksum of 'flux-system/auth-prod'
      team: platform
      app: auth
      env: prod
    - name: payments-dev
      namespace: flux-system
      revision: "sha256:ccc333"
      id: "248316208"  # Adler-32 checksum of 'flux-system/payments-dev'
      team: platform
      app: payments
      env: dev
    - name: payments-prod
      namespace: flux-system
      revision: "sha256:ddd444"
      id: "109867543"  # Adler-32 checksum of 'flux-system/payments-prod'
      team: platform
      app: payments
      env: prod
```

### 4. ResourceSet (flux-operator) Consumption
Templates Kustomizations using the discovered inputs:

```yaml
kind: ResourceSet
metadata:
  name: platform-apps
  namespace: flux-system
spec:
  inputsFrom:
      name: platform-apps-provider
  resources:
    - apiVersion: kustomize.toolkit.fluxcd.io/v1
      kind: Kustomization
      metadata:
        name: "<< inputs.app >>-<< inputs.env >>"
        namespace: flux-system
      spec:
        interval: 30m
        sourceRef:
          kind: ExternalArtifact
          name: "<< inputs.app >>-<< inputs.env >>"
        path: "./"
        prune: true
```
Closes #421


matheuscscp:
Thanks very much @vocarista! Can you please edit the PR description replacing it with the complete YAML / artifact content example from your comment in the issue? :pray:  Also the linter is failing

vocarista:
> Also the linter is failing

yeah, fixing that at the moment. Will update the description too.

vocarista:
@matheuscscp I'm adding a conditional watch for `ExternalArtifact` to fix the failing e2e tests, please tell me if I should also do a manual e2e test on a live cluster.

matheuscscp:
Thanks @vocarista, one last nit in the PR description: I think we can remove the `Changes` section, with today's people's maximum attention span it's basically dead text for training AI with AI-generated content. It would be more useful to show an ArtifactGenerator and the file structure it expects on the beginning, so we cover all the way from the Git repo to the ResourceSet.

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -167,6 +168,15 @@ The ACR, ECR and GAR Artifact Tag providers export the same inputs as the OCI Ar
 with the difference on the authentication method used to connect to the registry. For these providers,
 [secret-less](#secret-less) authentication is used.
 
+For ExternalArtifact the [exported inputs](#exported-inputs-status) structure is:
+
+- `id`: the Adler-32 checksum of the ExternalArtifact's namespaced name in the format `<namespace>+<name>` (type string).
```
matheuscscp: Let's change the id to the Adler-32 checksum of `<namespace>/<name>`

docs/api/v1/resourcesetinputprovider.md
```diff
@@ -294,6 +304,22 @@ spec:
...
+The `.spec.selector` field specifies the label selector used to discover `ExternalArtifact` objects in the same namespace as the `ResourceSetInputProvider`.
+This field is required when `.spec.type` is `ExternalArtifact`, and must not be set for any other provider type.
+
+Example:
+
+```yaml
+spec:
+  type: ExternalArtifact
+  selector:
+    matchLabels:
```
matheuscscp: Should show that we support the more advanced `matchExpressions` as well

internal/controller/resourcesetinputprovider_manager.go
```diff
@@ -21,15 +30,85 @@ type ResourceSetInputProviderReconcilerOptions struct {
...
+		u.SetGroupVersionKind(gvk)
+		b = b.Watches(
+			u,
+			handler.EnqueueRequestsFromMapFunc(r.requestsForExternalArtifacts),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		)
+		log.Info("Registered watch for ExternalArtifact resources")
+	} else {
+		log.Info("ExternalArtifact CRD not found in cluster, in-cluster artifact watches are disabled")
+	}
```
matheuscscp: The watch will never be registered on bootstrap, since ExternalArtifact does not exist before Flux is installed. Since flux-operator installs ExternalArtifact in the cluster after this startup code, the watch will only be registered on a flux-operator restart after ExternalArtifact was installed.

This is a tough one. We're gonna need an advanced mechanism from controller-runtime, a non-blocking `WatchesRawSource` that establishes an informer as soon as the CRD shows up in the cluster. Something like this:

```go
b.WatchesRawSource(source.TypedFunc[reconcile.Request](
	func(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
		go func() {
			u := &unstructured.Unstructured{}
			u.SetGroupVersionKind(schema.GroupVersionKind{
				Group: "source.toolkit.fluxcd.io", Version: "v1", Kind: "ExternalArtifact",
			})

			_ = wait.PollUntilContextCancel(ctx, 10*time.Second, true, func(ctx context.Context) (bool, error) {
				inf, err := mgr.GetCache().GetInformer(ctx, u, cache.BlockUntilSynced(false))
				if err != nil {
					return false, nil // CRD not installed yet; retry without blocking controller startup.
				}

				_, err = inf.AddEventHandler(toolscache.ResourceEventHandlerFuncs{
					AddFunc:    func(obj any) { r.enqueueExternalArtifact(ctx, q, obj) },
					DeleteFunc: func(obj any) { r.enqueueExternalArtifact(ctx, q, obj) },
					UpdateFunc: func(oldObj, newObj any) {
						r.enqueueExternalArtifact(ctx, q, oldObj) // catches label changes away from selector
						r.enqueueExternalArtifact(ctx, q, newObj)
					},
				})
				return err == nil, err
			})
		}()
		return nil
	},
))
```

```go
func (r *ResourceSetInputProviderReconciler) enqueueExternalArtifact(
	ctx context.Context,
	q workqueue.TypedRateLimitingInterface[reconcile.Request],
	obj any,
) {
	if d, ok := obj.(toolscache.DeletedFinalStateUnknown); ok {
		obj = d.Obj
	}
	if o, ok := obj.(client.Object); ok {
		for _, req := range r.requestsForExternalArtifacts(ctx, o) {
			q.Add(req)
		}
	}
}
```

matheuscscp (commented):
Please change the `ExternalArtifact` selector API from a single `spec.selector` to a list at `spec.selectors`, where each selector can also scope which namespace is searched.

Desired YAML shape:

```yaml
metadata:
  name: platform-apps-provider
  namespace: flux-system
spec:
  type: ExternalArtifact
  serviceAccountName: tenant-reader
  selectors:
    # Empty namespace means "same namespace as this ResourceSetInputProvider".
    - matchLabels:
        team: platform
        env: dev

    # Specific namespace means search only that namespace.
    - namespace: apps
      matchLabels:
        team: platform
        env: prod

    # Name selects one ExternalArtifact by metadata.name.
    # This is mutually exclusive with matchLabels and matchExpressions.
    - namespace: apps
      name: auth-prod

    # "*" means search all namespaces by omitting the namespace option from List().
    # Do not list namespaces and iterate them.
    - namespace: "*"
      matchLabels:
        generated-by: source-watcher
```

Suggested API type:

```go
type ExternalArtifactSelector struct {
	metav1.LabelSelector `json:",inline"`

	// Name selects a single ExternalArtifact by metadata.name.
	// This field is mutually exclusive with matchLabels and matchExpressions.
	// +optional
	Name string `json:"name,omitempty"`

	// Namespace scopes ExternalArtifact discovery.
	// When empty, ExternalArtifacts are listed in the ResourceSetInputProvider namespace.
	// When set to "*", ExternalArtifacts are listed across all namespaces by
	// omitting the namespace option from List().
	// Otherwise, ExternalArtifacts are listed only in the specified namespace.
	// +optional
	Namespace string `json:"namespace,omitempty"`
}

type ResourceSetInputProviderSpec struct {
	// ...

	// Selectors specifies label selectors used to discover ExternalArtifact objects.
	// Required when spec.type is ExternalArtifact and forbidden for all other provider types.
	// +optional
	Selectors []ExternalArtifactSelector `json:"selectors,omitempty"`
}
```

Implementation notes:

- Replace `spec.selector` with `spec.selectors`.
- CEL should require `spec.selectors` for `ExternalArtifact` and forbid it for all other provider types.
- Within each selector, `name` must be mutually exclusive with `matchLabels` and `matchExpressions`.
  Enforce this with a CEL rule on the `ExternalArtifactSelector` struct, e.g.
  `!has(self.name) || (!has(self.matchLabels) && !has(self.matchExpressions))`.
  Since `metav1.LabelSelector` is embedded with `json:",inline"`, the CEL expression should reference
  `self.matchLabels` and `self.matchExpressions` directly; there is no intermediate selector field.
- `spec.serviceAccountName` should be allowed for `ExternalArtifact`.
- When reconciling an `ExternalArtifact` provider, list using impersonation:
  - if `spec.serviceAccountName` is set, impersonate that service account in the RSIP namespace;
  - otherwise, if `--default-service-account` is set, impersonate the default service account in the RSIP namespace;
  - otherwise, use the operator client as today.
- This likely means `ResourceSetInputProviderReconciler` needs a `DefaultServiceAccount string`, wired from `main.go`, like `ResourceSetReconciler`.
- For each selector, resolve namespace as:
  - `""` -> `obj.GetNamespace()`
  - `"*"` -> no namespace filter in the `List()` call; do not list namespaces and iterate them
  - any other value -> `client.InNamespace(selector.Namespace)`
- The watch mapper also needs to account for this change. It can no longer list only RSIPs in the changed artifact's namespace, because an RSIP in namespace `a` may select artifacts from namespace `b` or `*`.
- When the watch mapper lists RSIPs across all namespaces, use `client.UnsafeDisableDeepCopy`; the mapper does not mutate RSIPs and this avoids unnecessary copies on a hot event path.

internal/controller/resourcesetinputprovider_manager.go
stefanprodan: Until now, RSIP had a predictable reconciliation cadence, determined by the annotation or the schedule. It's debatable if a watcher is needed given the complexity it brings...

stefanprodan: If you have 100 artifacts generated by the same AG, the watcher will trigger a storm of reconciliation requests which are partially mitigated by the dedup in the queue. Ideally we should watch AGs and trigger once per EA group, not even sure it's possible to index AGs here like that....

matheuscscp: A few thoughts.

### On RSIP uniformity

I don't think RSIP is a single CRD anymore. It's at least four:

* `Static` type. Does not make calls, exports only a single input set. Reconciliation cadence is fully agnostic of `reconcileEvery`.
* Various Git reference types. They make a single call and export multiple input sets.
* OCI tag types. They make one call to list tags + 1/tag to get their digests, export multiple input sets.
* `ExternalService` type. Makes one call, delegates the result entirely to an external service.

`ExternalArtifact` is an entirely new RSIP type, completely different from the type categories above. Makes no external calls, but looks at etcd instead. It would stand out as the fifth type category above.

Sure, some fields / behaviors (e.g. `.spec.schedule`) are inherited by most types.

But the only real uniformity that needs to be ensured for RSIP is already ensured by design: `.status.exportedInputs[]`, the field consumed by RSET. This is similar to how source-controller's CRDs implement a single common interface: `.status.artifact`. If you think about it, there's a strong parallel between source-controller's CRDs -> Flux appliers, and RSIP -> RSET. We already classify RSIP as a "source" in the Web UI and in `k get fluxcd-<category>` shipped in 2.9, just like we classify RSET as an "applier". Theoretically speaking, RSIP could very well be split into multiple CRDs that only need to implement `.status.exportedInputs[]`, just like source-controller's CRDs also have several things in common but their core functionalities are completely different and striving for uniformity in that regard is just impossible. While GitRepo and OCIRepo provide atomic views of the desired state, Bucket can show partial views.

So I don't think uniformity should get in the way of improvements that apply to particular RSIP types. 

### On reactivity between Flux resources

This would be the first "Flux Resource A --- depends on ---> Flux Resource B" relationship that is not backed by watches. I think the expectation of users for Flux Resource to Flux Resource dependencies being watch-based is much stronger than a uniform reconciliation cadence for RSIP. Users care much more about instant reactivity between Flux resources than "a stable interval", especially if they are not using `.spec.schedule`.

We went all the way to implement reactivity for the Web UI config Secret because you claimed this is what Flux users would expect, real-time reactivity for in-cluster resources. And the complexity associated with restarting the Web UI HTTP server is much bigger than this lazy non-blocking watch setup.

### On watching ArtifactGenerator

The integration point for RSIP is ExternalArtifact, and ExternalArtifact is by itself a very important integration point for Flux with external source controllers, so implementing reactivity for ArtifactGenerator would not just be hacky but also very incomplete. kustomize-controller and helm-controller watch ExternalArtifact.

### On watch storm for ExternalArtifacts

Why does queue dedup solve this only partially?

vocarista: I wasn't sure on the exact expectations (simplicity vs responsiveness) on this, so I've implemented the `WatchesRawSource` as suggested by @matheuscscp in my latest commit.

docs/api/v1/resourcesetinputprovider.md
vocarista: Done, thanks!

vocarista:
> Implementation notes:

Addressed these changes as per the notes, thanks for the detailed specs @matheuscscp.

matheuscscp:
Thanks @vocarista, please update the sample ResourceSetInputProvider in the PR description :pray:

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,162 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+		if ns == "" {
+			ns = obj.GetNamespace()
+		}
+
+		if sel.Name != "" {
+			if ns == "*" {
+				// List all namespaces and filter by name in Go
+				list := &unstructured.UnstructuredList{}
+				list.SetGroupVersionKind(externalArtifactListGVK)
+				if err := kubeClient.List(ctx, list); err != nil {
```
matheuscscp: I think we don't mutate the returned objects in this List(), right? In this case we can pass the option `client.UnsafeDisableDeepCopy` to avoid deep-copying all the objects

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,162 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+				return nil, fmt.Errorf("invalid label selector: %w", err)
+			}
+			list := &unstructured.UnstructuredList{}
+			list.SetGroupVersionKind(externalArtifactListGVK)
+			var listOpts []client.ListOption
+			if ns != "*" {
+				listOpts = append(listOpts, client.InNamespace(ns))
+			}
+			listOpts = append(listOpts, client.MatchingLabelsSelector{Selector: labelSel})
+			if err := kubeClient.List(ctx, list, listOpts...); err != nil {
```
matheuscscp, line 1104: Same here

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,162 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+	defaults map[string]any,
+) ([]fluxcdv1.ResourceSetInput, error) {
+
+	kubeClient := r.Client
+	if r.DefaultServiceAccount != "" || obj.Spec.ServiceAccountName != "" {
+		impersonatorOpts := make([]runtimeClient.ImpersonatorOption, 0, 1)
+		impersonatorOpts = append(impersonatorOpts,
+			runtimeClient.WithServiceAccount(r.DefaultServiceAccount, obj.Spec.ServiceAccountName, obj.GetNamespace()))
+		impersonation := runtimeClient.NewImpersonator(r.Client, impersonatorOpts...)
+		if impersonation.CanImpersonate(ctx) {
```
matheuscscp: I don't think we need this check, `CanImpersonate` was introduced to improve debugging for Flux Kustomizations applying tenant YAML to remote clusters. When the Kustomization is deleted, it sometimes deletes (as result of garbage collection) the very RBAC this Kustomization is using on the remote cluster, and the symptoms were very hard to debug. Since RSIP does not manage RBAC resources, we probably don't need this check here. We can just call `impersonation.GetClient(ctx)`.

internal/controller/resourcesetinputprovider_controller.go
vocarista: Thanks for the catch, done!

internal/controller/resourcesetinputprovider_controller.go
vocarista: Removed that call, thanks!

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,160 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+			seen[key] = true
+			uniqueMatches = append(uniqueMatches, ea)
+		}
+	}
+
+	// Apply filter limit.
+	limit := obj.GetFilterLimit()
+	if len(uniqueMatches) > limit {
+		uniqueMatches = uniqueMatches[:limit]
+	}
```
matheuscscp, line 1141: We need to ensure a stable order for `uniqueMatches`  before it reaches this limit filter. If the order is not stable and the amount of ExternalArtifacts is bigger than the limit then we will see things flapping... Let's sort by `<namespace>/<name>`.

matheuscscp (commented):
Looking very good, please add a controller test for each case:

- RSIP tries to read EAs from a namespace where the `.spec.serviceAccountName` does not have access
- RSIP tries to read EAs from a namespace where the `.spec.serviceAccountName` does have access

internal/controller/resourcesetinputprovider_controller.go
vocarista: Implemented the sort, thanks!

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,178 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+
+	// Sort uniqueMatches by <namespace>/<name> to ensure a stable ordering
+	// before applying filter limit, preventing flapping.
+	slices.SortFunc(uniqueMatches, func(a, b *unstructured.Unstructured) int {
+		if a.GetNamespace() != b.GetNamespace() {
+			if a.GetNamespace() < b.GetNamespace() {
+				return -1
+			}
+			return 1
```
matheuscscp:
```suggestion
			return strings.Compare(a.GetNamespace(), b.GetNamespace())
```

internal/controller/resourcesetinputprovider_controller.go
```diff
@@ -988,6 +1019,178 @@ func (r *ResourceSetInputProviderReconciler) callExternalServiceProvider(
...
+		}
+		if a.GetName() < b.GetName() {
+			return -1
+		}
+		if a.GetName() > b.GetName() {
+			return 1
+		}
+		return 0
```
matheuscscp:
```suggestion
		return strings.Compare(a.GetName(), b.GetName())
```

matheuscscp (approved):
LGTM! 🚀 

Awesome feature @vocarista!

vocarista:
Hi @stefanprodan this PR is ready for your review. Please take a look whenever you have a chance.

stefanprodan:
@vocarista I'm back from vacation and I'll do a review. Please rebase with upstream main and force push, we fixed CI.

api/v1/resourcesetinputprovider_types.go
```diff
@@ -49,20 +50,23 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="!has(self.secretRef) || !(self.type == 'Static' || self.type == 'ExternalArtifact' || self.type.startsWith('AWSCodeCommit') || (self.type.endsWith('ArtifactTag') && self.type != 'OCIArtifactTag'))", message="cannot specify spec.secretRef when spec.type is one of Static, ExternalArtifact, AWSCodeCommit*, ACRArtifactTag, ECRArtifactTag or GARArtifactTag"
+// +kubebuilder:validation:XValidation:rule="self.type != 'ExternalArtifact' || has(self.selectors)", message="spec.selectors must be set when spec.type is 'ExternalArtifact'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'ExternalArtifact' || !has(self.selectors)", message="spec.selectors must not be set when spec.type is not 'ExternalArtifact'"
 type ResourceSetInputProviderSpec struct {
```
stefanprodan, line 67: `has(self.selectors)` only checks field presence, so `selectors: []` passes admission, we should use `size(self.selectors) > 0`.

stefanprodan (approved):
LGTM

Thanks @vocarista for implementing this feature across Flux controllers! 

Awesome contribution 🏅

PR #942, opened 2026-07-05
feat: add `certSecretRef` for distribution

### LINKED ISSUE #66: Pull Flux manifests from registry with private PKI certificate
Our cluster does not have a direct connection to the Internet, and must pull all images through a registry proxy. This registry is configured with a certificate chain rooted in our private self-signed CA. How can I make Flux Operator trust this root?

This error is logged from Flux Operator:

````json
{"level":"error","ts":"2024-07-04T12:03:33.625Z","msg":"Reconciler error","controller":"fluxinstance","controllerGroup":"fluxcd.controlplane.io","controllerKind":"FluxInstance","FluxInstance":{"name":"flux","namespace":"flux-system"},"namespace":"flux-system","name":"flux","reconcileID":"90115059-3e19-4cc3-ad0a-8af12fe0f077","error":"pulling artifact oci://<REDACTED>/ghcr-docker-remote/controlplaneio-fluxcd/flux-operator-manifests failed: Get \"https://<REDACTED>/v2/\": tls: failed to verify certificate: x509: certificate signed by unknown authority"}
````

Flux Operator version: 0.6.0
Flux Operator installation method: OLM (operatorhub.io Subscription)

FluxInstance resource (irrelevant details omitted):

````yaml
kind: FluxInstance
metadata:
  name: flux
  namespace: flux-system
spec:
  distribution:
    artifact: 'oci://<REDACTED>/ghcr-docker-remote/controlplaneio-fluxcd/flux-operator-manifests'
    registry: ghcr.io/fluxcd
    version: 2.3.x
````

Note: Our CRIO config will ensure the `registry` (for Flux images), is rewritten when images are pulled by our cluster.

### PR DESCRIPTION
**Pull request by a1994sc**
This PR add both the `CertSecretRef` and `Insecure` options for `Distribution`, and it will fix #66 .

The majority of this PR is based off the [`fluxcd/source-controller`](https://github.com/fluxcd/source-controller/blob/39b711b111fa906c3db0a424e252a5d12e48646d/api/v1/ocirepository_types.go#L99-L112) code, specifically the `OCIRepo` logic.

### CODE REVIEW

File: internal/controller/fluxinstance_artifact_controller.go [hunk 1]
```diff
@@ -133,3 +137,50 @@ func requeueArtifactAfter(obj *fluxcdv1.FluxInstance) ctrl.Result {
 	}
 	return result
 }
+
+// GetDistributionTransport clones the default transport from remote and when a certSecretRef is specified,
+// the returned transport will include the TLS client and/or CA certificates.
+// If the insecure flag is set, the transport will skip the verification of the server's certificate.
+// based off: https://github.com/fluxcd/source-controller/blob/39b711b111fa906c3db0a424e252a5d12e48646d/internal/controller/ocirepository_controller.go#L1010-L1014
```
matheuscscp:
```suggestion
```

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -133,3 +137,50 @@ func requeueArtifactAfter(obj *fluxcdv1.FluxInstance) ctrl.Result {
...
+	if tlsConfig != nil {
+		transport.TLSClientConfig = tlsConfig
+	}
+
+	return transport, nil
+}
+
+// getTLSConfig gets the TLS configuration for the transport based on the
+// specified secret reference in the FluxInstance object, or the insecure flag.
+// based off: https://github.com/fluxcd/source-controller/blob/39b711b111fa906c3db0a424e252a5d12e48646d/internal/controller/ocirepository_controller.go#L1032-L1034
```
matheuscscp:
```suggestion
```

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -133,3 +137,50 @@ func requeueArtifactAfter(obj *fluxcdv1.FluxInstance) ctrl.Result {
...
+func getTLSConfig(ctx context.Context, obj *fluxcdv1.FluxInstance, kubeClient client.Client) (*cryptotls.Config, error) {
+	if obj.Spec.Distribution.CertSecretRef == nil || obj.Spec.Distribution.CertSecretRef.Name == "" {
+		if obj.Spec.Distribution.Insecure {
+			// NOTE: This is the only place in Flux where InsecureSkipVerify is allowed.
+			// This exception is made for OCIRepository to maintain backward compatibility
+			// with tools like crane that require insecure connections without certificates.
+			// This only applies when no CertSecretRef is provided AND insecure is explicitly set.
```
matheuscscp, line 166: Let's add an XValidation to the `Distribution` struct in the API package with a CEL expression to prevent `insecure: true` and `certSecretRef` from being set together. Let's also error the reconciliation out if that happens. Let's also add controller tests for both things (controller tests must use the envtest framework).

internal/controller/fluxinstance_artifact_controller.go
```diff
@@ -133,3 +137,50 @@ func requeueArtifactAfter(obj *fluxcdv1.FluxInstance) ctrl.Result {
...
+	}
+
+	secretName := types.NamespacedName{
+		Namespace: obj.Namespace,
+		Name:      obj.Spec.Distribution.CertSecretRef.Name,
+	}
+	// NOTE: Use WithSystemCertPool to maintain backward compatibility with the existing
+	// extend approach (system CAs + user CA) rather than the default replace approach (user CA only).
+	// This ensures source-controller continues to work with both system and user-provided CA certificates.
+	var tlsOpts = []secrets.TLSConfigOption{secrets.WithSystemCertPool()}
```
matheuscscp, line 182: This is annoying but I guess it's better to implement the same behavior of OCIRepository...

matheuscscp (commented):
Thanks for this PR @a1994sc!

internal/controller/fluxinstance_artifact_controller.go
a1994sc: gotcha, I will remove the reference, wasn't sure so wanted to air on the side of being overely transparent then under

internal/controller/fluxinstance_artifact_controller.go
a1994sc: I can add the CEL expression, after I look into how that works ha.
I am planning on adding tests once I got a little input from the maintainers, like yourself so thanks.

api/v1/fluxinstance_types.go
```diff
@@ -114,6 +114,25 @@ type Distribution struct {
...
+	// registry. The client cert and key are useful if you are
+	// authenticating with a certificate; the CA cert is useful if
+	// you are using a self-signed server certificate. The Secret must
+	// be of type `Opaque` or `kubernetes.io/tls`.
+	// +optional
+	CertSecretRef *meta.LocalObjectReference `json:"certSecretRef,omitempty"`
+
+	// Insecure allows connecting to a non-TLS HTTP container registry.
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
```
stefanprodan, line 135: This is about the artifact registry which can be different from the `distribution.registry`. A rename would be good, prefix with Artifact like the `ArtifactPullSecret`.

matheuscscp: How about creating a new field that is mut. exclusive with the two existing artifact fields? Like this:

```yaml
distribution:
  artifactSpec: # optional
    url: ... # required. same as distribution.artifact
    secretRef: ... # same as distribution.artifactPullSecret
    certSecretRef: ... # new field
    insecure: ... # new field
```

This way we have a spec field dedicated for the artifact, and the fields are 1:1 match with the OCIRepository spec. This enables more OCIRepo spec fields to be imitated later.

stefanprodan: I'm not for changing the API and offer 2 ways to configure the same feature. We discuss this before, changes must be additive, we'll never remove/deprecate fields in FluxInstance.

stefanprodan: These fields imply you are vendoring the artifacts, but unless you pay for Flux EE, doing this vendoring makes no sense. The artifact contains the ControlPlane hardened images with CVE patches, for CNCF Flux you only need the manifest embedded in the container.

stefanprodan: @a1994sc are you a Flux Enterprise customer? If not, what's the point of syncing the manifests to a private registry?

a1994sc: So I am not an enterprise customer, but would like the flexibility to pull the open-source artifacts thru an internal registry proxy (which has its own cert chain). But also if I wanted to create a custom distro that allows the use of `registry1.dso.mil` images instead of the github registry images

stefanprodan: You don't need to set the artifact for that. See https://fluxoperator.dev/docs/crd/fluxinstance/#distribution-registry-mirrors

a1994sc: Please forgive my ignorance, but does this still not run into the same issue where the operator needs to talk that registry endpoint to pull the distributions package?

stefanprodan: There is no artifact to pull if you don't set it. It will use the one bundled in the operator container image.

Dunateo:
Hello, 
I think this feature would be especially useful for users installing a FluxInstance via the  flux-operator-cli image. In air-gapped environments, when  distribution.artifact is removed, the operator still attempts to pull the manifest from ghcr.io. 

Allowing a private registry to host and sync the manifests would make it possible to control the deployed version more explicitly and avoid this external dependency. 

I think this is worth considering for this PR and help community.

matheuscscp:
> distribution.artifact is removed, the operator still attempts to pull the manifest from ghcr.io

@Dunateo I need some proof of this affirmation. If `.spec.distribution.artifact` is not in the FluxInstance and flux-operator still tries to fetch the artifact, this is a bug and the fix is removing that behavior. Fixing by supporting the whole thing is just a fever dream.

Femure:
I can confirm this issue in an air-gapped environment. Two problems arise that are both undesirable:

1. Even when `distribution.artifact` is omitted, the operator still attempts to pull from `ghcr.io` with the `latest` tag — bypassing any pinned version and failing because `ghcr.io` is unreachable.
2. When `distribution.artifact` is explicitly set to my internal registry, it fails because the OCI client now defaults to HTTPS (this worked in v0.47.0 where HTTP was used), and I have no way to configure the scheme or provide a CA certificate.

**My setup:** I run a registry v2 at `myregistry.registry.svc.cluster.local:5050` with all images cached locally, and I bootstrap using the `flux-operator-cli` as a Kubernetes Job with `-f flux-instance.yaml`. My `FluxInstance` is:

```yaml
spec:
  distribution:
    version: "2.x"
    registry: "ghcr.io/fluxcd"
    artifact: "oci://myregistry.registry.svc.cluster.local:5050/ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:v0.53.0"
    components:
      - source-controller
      - kustomize-controller
      - helm-controller
      - notification-controller
  cluster:
    type: kubernetes
    size: small
```

**Error with `artifact` set:** The OCI client sends an HTTPS request to an HTTP server, resulting in a TLS handshake failure. This was not the case in v0.47.0.

**Error without `artifact`:** Even with the field omitted, I get:
```
failed to pull distribution artifact: oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:latest failed:
Get "https://ghcr.io/v2/": dial tcp: lookup ghcr.io on 10.43.0.10:53: server misbehaving
```
This is the same error @Dunateo reported. The operator is reaching out to `ghcr.io` regardless, which is unreachable in my air-gapped environment.

**Regarding the embedded manifests claim:** I inspected the `flux-operator-cli:v0.53.0` image with `buildah` and found only a JSON schema (for YAML validation) and the `flux` binary, no embedded Flux manifests. So I don't see how the operator can install without either fetching them from `ghcr.io` or having them provided via `artifact`.

This PR's `insecure` and `certSecretRef` fields for `distribution` would directly solve both issues for air-gapped users. Happy to provide logs or a reproduction if helpful.

matheuscscp:
Hey @Femure @Dunateo it's impossible for the FluxInstance controller to be trying to fetch the artifact when the field is unset:

https://github.com/controlplaneio-fluxcd/flux-operator/blob/9c0e33843d1171fe9691f9d0e37a9b336d469adb/internal/controller/fluxinstance_controller.go#L275-L304

So you should simply *not set* the field. It's easy to do that. Just don't set it.

> I bootstrap using the flux-operator-cli as a Kubernetes Job with -f flux-instance.yaml

This is a completely different problem. Are you using `flux-operator install -f flux-instance.yaml`? This command was not designed for air-gapped support at all, this command was fully designed to pull manifests to install *Flux Operator*, they are a completely different set of manifests.

I'm closing this PR as we have no intention to support the field for FluxInstance.

PR #945, opened 2026-07-06
Add flag to scope RS/RSIP to given namespaces

### LINKED ISSUE #944: Scope `ResourceSet`/`ResourceSetInputProvider` reconciliation to specific namespaces
### Problem

In a multi-tenant cluster it's useful to run one flux-operator per tenant, each
in its own namespace. Today the operator caches and reconciles `ResourceSet` and
`ResourceSetInputProvider` objects cluster-wide, so every tenant's operator
watches and reconciles every other tenant's RS/RSIP. That doesn't scale with the
number of tenants and weakens the isolation between them.

### Proposal

Add an optional `--watch-namespaces` flag. When set, the RS and RSIP informer
caches are scoped to the given namespaces (repeatable and/or comma-separated);
the operator's own runtime namespace is always included. When unset, behaviour is
unchanged — RS/RSIP stay cluster-wide — so the change is fully backward
compatible.

Only the RS/RSIP caches would narrow. `FluxInstance`/`FluxReport` scoping, the
ConfigMap/Secret metadata watches, and the controller RBAC would stay as they are
(a ResourceSet can reference inputs in any namespace).

### Alternatives considered

- Expressing the scope via a field on the `FluxInstance` API rather than a CLI flag.
- A single cluster-wide operator relying on per-tenant reduced-permission service
  accounts (RBAC/OPA) instead of cache scoping.

Happy to shape the API to whatever you'd prefer. I have a small, tested
implementation ready to open as a PR once there's agreement on the approach.

Disclosure: the implementation was AI-assisted and reviewed by me.

dejamiko:
Closes #944.

### What
Adds an optional `--watch-namespaces` flag to the operator. When set, the
ResourceSet and ResourceSetInputProvider informer caches are scoped to the
given namespaces (repeatable and/or comma-separated; the operator's runtime
namespace is always included). When unset, behaviour is unchanged
(cluster-wide), so this is backward compatible.

### Why
Enables running one operator per tenant without each operator watching and
reconciling every other tenant's RS/RSIP. See #944 for the full discussion.

### Scope / non-goals
Only the RS/RSIP **caches** are narrowed. FluxInstance/FluxReport scoping,
the ConfigMap/Secret metadata watches, and the controller RBAC remain
cluster-wide (a ResourceSet may reference inputs in any namespace). Tightening
RBAC to the watched set would be a natural follow-up.

### Testing
- `resolveWatchNamespaces` is unit-tested: empty/no-op, single, comma-separated,
  repeated, dedup including the runtime namespace, whitespace trimming,
  all-empty-token error, and invalid DNS-1123 label error.
- `make test` passes locally.
- Validated on a live multi-tenant cluster: a scoped operator reconciles only
  its own namespace's RS/RSIP and leaves other tenants' objects untouched.

### AI disclosure
Implementation was AI-assisted (Claude, an Anthropic LLM) and reviewed by me.

### DIFF
--- a/cmd/operator/cache.go
+++ b/cmd/operator/cache.go
@@ -1,0 +1,53 @@
+// Copyright 2024 Stefan Prodan.
+// SPDX-License-Identifier: AGPL-3.0
+
+package main
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// resolveWatchNamespaces converts the raw --watch-namespaces flag values into a
+// sorted, de-duplicated namespace set used to scope ResourceSet /
+// ResourceSetInputProvider reconciliation. Each value may be repeated and/or
+// comma-separated. The operator's runtime namespace is always included.
+//
+// It returns (nil, nil) when no scoping is requested (empty input). It returns
+// an error if any token is not a valid DNS-1123 label, or if values were given
+// but none resolved to a non-empty namespace (so a typo like
+// "--watch-namespaces=," fails loudly instead of silently scoping to just the
+// runtime namespace).
+func resolveWatchNamespaces(runtimeNamespace string, watchNamespaces []string) ([]string, error) {
+	if len(watchNamespaces) == 0 {
+		return nil, nil
+	}
+	set := map[string]struct{}{runtimeNamespace: {}}
+	valid := 0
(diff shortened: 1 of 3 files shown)


dejamiko:
Closing following a conversation with @stefanprodan, will try the sharding route instead, as discussed in the issue and in Maintainer call

PR #948, opened 2026-07-07
operator: fix skipping flux patch versions in the artifact

matheuscscp:
#937 was too strict, the operator should still upgrade Flux to patch releases automatically if the minor is in the storage but the patch is only in the artifact.


stefanprodan (approved):
LGTM

PR #954, opened 2026-07-10
web: Add OAuth2 autoLogin to skip login page

### LINKED ISSUE #953: [web-ui] Feature Request: Auto-redirect to OIDC (skip login button)
When OAuth2/OIDC is enabled, the Web UI shows a "Continue with SSO" page before redirecting to the IdP.
For SSO-only setups this extra click is unnecessary.

**Suggestion:** Add an optional config flag, e.g. `web.config.authentication.oauth2.autoLogin: true`, to redirect unauthenticated users straight to the OIDC provider (same as clicking the button today), preserving `originalPath` for deep links.

Default: `false` (current behavior).

roberteggl:
## Summary

Add optional `oauth2.autoLogin` to the Web UI config (default `false`). When enabled, unauthenticated users skip the login page and redirect to the OIDC provider, preserving `originalPath` for deep links.

After explicit logout, auto-redirect is suppressed via an `auth-logout` cookie so users are not immediately signed back in through the IdP session.

Closes #953

## Config

```yaml
spec:
  authentication:
    type: OAuth2
    oauth2:
      autoLogin: true
```

## Test plan

- [x] `go test ./internal/web/auth/... ./internal/web/config/...`
- [x] `make web-test`
- [x] Manual OIDC test on cluster (via local build and live test)
- [x] Manual logout → login page shown, no instant re-login

## AI disclosure

Assisted-by: Cursor/Composer2.5

### DIFF
--- a/internal/web/auth/cookies.go
+++ b/internal/web/auth/cookies.go
@@ -93,13 +93,12 @@ func setAuthLogoutCookie(w http.ResponseWriter) {
 // It first clears any existing auth provider cookie to avoid duplicates.
 func setAuthProviderCookie(w http.ResponseWriter, provider, loginURL string, authenticated, autoLogin bool) {
 	clearCookieFromResponse(w, cookieNameAuthProvider)
-	payload := map[string]any{
+	setCookie(w, cookieNameAuthProvider, map[string]any{
 		"provider":      provider,
 		"url":           loginURL,
 		"authenticated": authenticated,
 		"autoLogin":     autoLogin,
-	}
-	setCookie(w, cookieNameAuthProvider, payload)
+	})
 }
 
 // SetAnonymousAuthProviderCookie sets the anonymous auth provider cookie in the response.
--- a/internal/web/auth/cookies.go
+++ b/internal/web/auth/cookies.go
@@ -16,6 +16,7 @@ import (
 
 const (
 	cookieNameAuthError        = "auth-error"
+	cookieNameAuthLogout       = "auth-logout"
 	cookieNameAuthProvider     = "auth-provider"
 	cookieNameAuthStorage      = "auth-storage"
 	cookieNameOAuth2LoginState = "oauth2-state"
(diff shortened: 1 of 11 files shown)


internal/web/auth/cookies.go
```diff
@@ -80,20 +81,32 @@ func setAuthErrorCookie(w http.ResponseWriter, err error) {
...
+func setAuthProviderCookie(w http.ResponseWriter, provider, loginURL string, authenticated, autoLogin bool) {
 	clearCookieFromResponse(w, cookieNameAuthProvider)
-	setCookie(w, cookieNameAuthProvider, map[string]any{
+	payload := map[string]any{
 		"provider":      provider,
 		"url":           loginURL,
 		"authenticated": authenticated,
-	})
+	}
+	if autoLogin {
```
matheuscscp: I think we could just put the boolean in the payload without this `if`

matheuscscp (commented):
@Robert27 Thanks for this PR!

matheuscscp:
Suggestion from codex/gpt-5.6-sol:

> The logout suppression looks useful, but currently `auth-logout` is deleted as soon as `LoginPage` mounts. This means refreshing the logged-out page can trigger auto-login again.
> 
> Could we keep the marker until the user explicitly chooses to sign in?
> 
> ```js
> // During initialization:
> const logout = parseAuthLogoutCookie()
> 
> // Do not delete auth-logout here. Keep suppressing auto-login
> // until the user explicitly signs in.
> ```
> 
> Then clear it from the manual login handler:
> 
> ```js
> const handleLogin = () => {
>   const loginUrl = getLoginUrl()
> 
>   if (loginUrl) {
>     // The user explicitly chose to sign in, so auto-login can resume.
>     deleteCookie('auth-logout')
>     setIsLoading(true)
>     window.location.href = loginUrl
>   }
> }
> ```
> 
> The existing `!logout` auto-login condition can remain unchanged. This keeps the login page stable across refreshes after an explicit logout.

roberteggl:
Thanks for your fast review, I implemented all your requested changes :)

internal/web/auth/cookies.go
```diff
@@ -80,20 +81,30 @@ func setAuthErrorCookie(w http.ResponseWriter, err error) {
...
 	clearCookieFromResponse(w, cookieNameAuthProvider)
-	setCookie(w, cookieNameAuthProvider, map[string]any{
+	payload := map[string]any{
 		"provider":      provider,
 		"url":           loginURL,
 		"authenticated": authenticated,
-	})
+		"autoLogin":     autoLogin,
+	}
+	setCookie(w, cookieNameAuthProvider, payload)
```
matheuscscp:
```suggestion
	setCookie(w, cookieNameAuthProvider, map[string]any{
		"provider":      provider,
		"url":           loginURL,
		"authenticated": authenticated,
		"autoLogin":     autoLogin,
	})
```

matheuscscp (approved):
LGTM! 🚀

Thanks @Robert27! Just one last nit 🙏

internal/web/auth/cookies.go
roberteggl: Fixed

PR #956, opened 2026-07-11
web: add more keyboard shortcuts and help modal

### LINKED ISSUE #955: [web-ui] Feature Request: Add more  keyboard shortcuts and a help modal
Add a `?` shortcut that opens a modal with all keyboard shortcuts. Maybe also a link in the user menu.

There is already `/` for Quick Search, `Esc` to close things, arrows + Enter in search, and `ns:` / `kind:` / `**` filters.

I'd also suggest new shortcuts for jumping between views (`g` then a letter, GitHub-style):

- `g` `d` - Dashboard
- `g` `f` - Favorites
- `g` `r` - Resources
- `g` `w` - Workloads
- `g` `e` - Events

And a few more globals:

- `Shift+R` - refresh
- `[` / `]` - prev / next tab
- `s` - toggle favorite
- `l` - open logs on workload pages

roberteggl:
## Summary

Closes #955.

Adds keyboard shortcuts to the Flux Status Page:

- `?` — help modal listing all shortcuts (also linked from the user menu)
- `g` + letter — navigate between Dashboard, Favorites, Resources, Workloads, and Events
- `Shift+R` — refresh the current view
- `s` / `c` / `l` — toggle favorite, copy link, and open logs on detail pages
- `[` / `]` — cycle tabs on list and detail views

Shortcuts are suppressed while typing, when overlays are open, and when RBAC blocks an action (e.g. logs without the `logs` permission).

## Test plan

- [ ] Press `?` and confirm the help modal opens and lists shortcuts
- [ ] Use `g` chords to jump between main views
- [ ] On a resource/workload detail page, try `Shift+R`, `s`, `c`, and `[` / `]`
- [ ] On a workload detail page with logs access, press `l` to open logs
- [ ] Confirm shortcuts do not fire while focused in a search/input field or with logs/search open
- [ ] `make web-test`

Assisted-by: Cursor/Composer 2.5

### DIFF
--- a/web/src/components/dashboards/resource/ReconcilerPanel.jsx
+++ b/web/src/components/dashboards/resource/ReconcilerPanel.jsx
@@ -14,12 +14,34 @@ import { HistoryTimeline } from './HistoryTimeline'
 import { FluxOperatorIcon } from '../../layout/Icons'
 import { useHashTab } from '../../../utils/hash'
 
-// Valid tabs for the ReconcilerPanel
-const RECONCILER_TABS = ['overview', 'history', 'events', 'values', 'spec', 'status']
-
 export function ReconcilerPanel({ kind, name, namespace, resourceData }) {
+  const valuesYaml = useMemo(() => {
+    if (kind !== 'HelmRelease' || !resourceData) return null
+    if (resourceData.status?.helmValues) return resourceData.status.helmValues
+    if (resourceData.spec?.values) return resourceData.spec.values
+    return null
+  }, [resourceData, kind])
+
+  const valuesError = useMemo(() => {
+    if (kind !== 'HelmRelease' || !resourceData) return null
+    return resourceData.status?.helmValuesError || null
+  }, [resourceData, kind])
+
+  const visibleTabs = useMemo(() => {
+    const tabs = ['overview']
+    if (resourceData?.status?.history?.length > 0) {
+      tabs.push('history')
+    }
+    tabs.push('events')
+    if (kind === 'HelmRelease' && (valuesError || valuesYaml)) {
+      tabs.push('values')
+    }
+    tabs.push('spec', 'status')
+    return tabs
+  }, [resourceData, kind, valuesError, valuesYaml])
+
   // Tab state synced with URL hash (e.g., #reconciler-events)
-  const [reconcilerTab, setReconcilerTab] = useHashTab('reconciler', 'overview', RECONCILER_TABS, 'reconciler-panel')
+  const [reconcilerTab, setReconcilerTab] = useHashTab('reconciler', 'overview', visibleTabs, 'reconciler-panel')
 
   // Events data state
   const [eventsData, setEventsData] = useState([])
@@ -51,18 +73,6 @@ export function ReconcilerPanel({ kind, name, namespace, resourceData }) {
       spec: resourceData.spec
     }
   }, [resourceData])
-
-  const valuesYaml = useMemo(() => {
-    if (kind !== 'HelmRelease' || !resourceData) return null
-    if (resourceData.status?.helmValues) return resourceData.status.helmValues
-    if (resourceData.spec?.values) return resourceData.spec.values
-    return null
-  }, [resourceData, kind])
-
-  const valuesError = useMemo(() => {
-    if (kind !== 'HelmRelease' || !resourceData) return null
-    return resourceData.status?.helmValuesError || null
-  }, [resourceData, kind])
 
   const statusYaml = useMemo(() => {
     if (!resourceData?.status) return null
--- a/web/src/components/favorites/FavoritesPage.jsx
+++ b/web/src/components/favorites/FavoritesPage.jsx
@@ -22,6 +22,71 @@ export const favoritesData = signal({})
 // fetch. On warm return visits hasCachedData suppresses the pulse regardless.
 export const favoritesFetching = signal(true)
 
+// Monotonic generation so overlapping polls / Shift+R refreshes discard stale
+// responses instead of overwriting newer favoritesData or clearing favoritesFetching
+// while a later request is still in flight.
+let favoritesFetchGeneration = 0
+
+/**
+ * Fetches current status for all favorites from the API.
+ * Used by the page poll interval and the Shift+R keyboard shortcut.
+ */
+export async function fetchFavoritesData() {
+  const generation = ++favoritesFetchGeneration
+  const currentFavorites = favorites.value
+
+  if (currentFavorites.length === 0) {
+    if (generation === favoritesFetchGeneration) {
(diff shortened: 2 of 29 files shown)


roberteggl:
image
image

stefanprodan:
Great UX improvement! I'll do a proper review by the end of next week when I'm back from vacation.

web/src/utils/useKeyboardShortcuts.js
```diff
@@ -0,0 +1,129 @@
...
+        const direction = e.key === ']' ? 1 : -1
+        let handled = false
+        if (SECTION_TAB_PATHS.includes(location.path)) {
+          handled = cycleSectionTab(location.path, direction, location.route)
+        } else if (parseDetailRoute(location.path)) {
+          handled = cycleHashTab(direction)
+        }
+        if (handled) {
+          e.preventDefault()
+        }
```
stefanprodan: The [/] handler does not reject modifier keys. On macOS, Cmd+[ and Cmd+] are standard back/forward shortcuts, but this handler cycles tabs and calls preventDefault(). We should add the same ctrlKey/metaKey/altKey guards used by the other shortcuts.

web/src/utils/useKeyboardShortcuts.js
```diff
@@ -0,0 +1,129 @@
...
+          keyboardShortcutsOpen.value = !keyboardShortcutsOpen.value
+        }
+        return
+      }
+
+      if (keyboardShortcutsOpen.value) {
+        return
+      }
+
+      if (e.key === 'g' && !e.ctrlKey && !e.metaKey && !e.altKey) {
```
stefanprodan, line 51: We should suppress g navigation while the logs viewer is open, tested this manually and pressing "g f" navigates away because `allowOverlays: true `bypasses every overlay guard, including `workloadLogsOpen`.

web/src/components/dashboards/resource/ReconcilerPanel.jsx
```diff
@@ -14,12 +14,34 @@ import { HistoryTimeline } from './HistoryTimeline'
...
+    if (kind === 'HelmRelease' && (valuesError || valuesYaml)) {
+      tabs.push('values')
+    }
+    tabs.push('spec', 'status')
+    return tabs
+  }, [resourceData, kind, valuesError, valuesYaml])
+
   // Tab state synced with URL hash (e.g., #reconciler-events)
-  const [reconcilerTab, setReconcilerTab] = useHashTab('reconciler', 'overview', RECONCILER_TABS, 'reconciler-panel')
+  const [reconcilerTab, setReconcilerTab] = useHashTab('reconciler', 'overview', visibleTabs, 'reconciler-panel')
```
stefanprodan, line 44: Dynamic `visibleTabs` arrays are recreated when polled resource objects change. `useHashTab` treats every new array identity as a synchronization event and calls `scrollIntoView` whenever the hash matches.

Users viewing #reconciler-* or #source-* can therefore be pulled back to that panel on each refresh. We should separate hash/state synchronization from scrolling, so scrolling occurs only on initial load or actual hash navigation.

web/src/components/favorites/FavoritesPage.jsx
```diff
@@ -22,6 +22,52 @@ export const favoritesData = signal({})
 // fetch. On warm return visits hasCachedData suppresses the pulse regardless.
 export const favoritesFetching = signal(true)
 
+/**
+ * Fetches current status for all favorites from the API.
+ * Used by the page poll interval and the Shift+R keyboard shortcut.
+ */
+export async function fetchFavoritesData() {
```
stefanprodan, line 34: This unconditionally writes its response to the global signal. The effect’s cancelled flag no longer protects that write.

If favorites change while a request is pending, the older response can resolve last and replace the newer status map; its `finally` can also clear `favoritesFetching` while another request remains active.

We should use an abort controller or request generation token before updating either signal.

docs/web/web-least-privilege-rbac.md
```diff
@@ -230,6 +230,20 @@ Users do not need cluster-wide `list` permissions on namespaces just to populate
...
+**Where:** Flux Status Page – global keyboard shortcuts for navigation, refresh,
+favorites, copy link, tab cycling, and logs.
+
+**Internal operation:** None. Shortcuts invoke the same client-side flows and
+backend API calls as the existing UI controls they mirror.
+
+**RBAC impact:** No new backend endpoints, privileged Kubernetes calls, or RBAC
+requirements. The `l` (open logs) shortcut is registered only when the user
+has the `logs` action and inspectable pods, matching the View logs button gate.
+
```
stefanprodan: I don't think we need any of these in the RBAC doc, please undo.

web/src/utils/useKeyboardShortcuts.js
roberteggl: Good catch on Cmd+[ / Cmd+]. I had to allow Option/AltGr though, since on German layouts [ / ] are typed with Option+5/6 (and AltGr on Windows), so rejecting altKey broke the shortcut entirely. Plain Ctrl+[ stays blocked.

web/src/utils/useKeyboardShortcuts.js
roberteggl: fixed inside `shouldIgnoreShortcut`

web/src/components/dashboards/resource/ReconcilerPanel.jsx
roberteggl: Fixed in useHashTab: tab sync still runs when validTabs changes, but scrollIntoView is only triggered on mount and hashchange. Poll-driven visibleTabs identity churn no longer pulls the viewport back to #reconciler-* / #source-*.

web/src/components/favorites/FavoritesPage.jsx
roberteggl: Fixed with a request generation token in fetchFavoritesData: only the latest in-flight call may write favoritesData or clear favoritesFetching. Stale responses are discarded. Also removed the page effect’s duplicate finally write so it no longer races the shared helper.

stefanprodan:
@roberteggl can you please rebase with upstream main, the squash the 2 commits and force push. This PR should have a single commit.

roberteggl:
Sure thing, I thought you might use the squash PR function in GitHub during merge :)

stefanprodan:
We don't use squash in the Flux projects as it removes the the contributor's PGP signature. We also don't accept merge commits inside PRs, rebasing with AI is easy.

stefanprodan (approved):
LGTM

Thanks @roberteggl 🏅

roberteggl:
Thanks for the fast review!
Also good to know about the signatures. I wasn’t aware that GitHub strips those.

PR #960, opened 2026-07-15
web: update docs for user actions after `FineGrained` mode

matheuscscp:
Some docs were outdated after the introduction of `.spec.userActions.access: FineGrained`.


stefanprodan (approved):
LGTM

PR #964, opened 2026-07-21
web: add reverse proxy support

kdihalas:
## Summary

Adds ReverseProxy authentication to the Flux Web UI for deployments where an upstream proxy already authenticates users and forwards their identity through HTTP headers. Reverse proxy identities are processed using the same CEL-based variables, validations, profile, and Kubernetes impersonation system used by OAuth2/OIDC.

  ## Changes

  - Adds ReverseProxy as a supported authentication type.
  - Accepts trusted proxy IP addresses and CIDR ranges.
  - Trusts only the direct TCP peer address; forwarded address headers are ignored.
  - Exposes canonical HTTP headers to CEL as string claims.
  - Reuses the existing OAuth claims processor.
  - Adds Nginx configuration and security documentation.

  ## Example

```yaml
  apiVersion: web.fluxcd.controlplane.io/v1
  kind: Config
  spec:
    authentication:
      type: ReverseProxy
      reverseProxy:
        trustedProxies:
          - 10.244.0.0/16

        variables:
          - name: username
            expression: "claims['X-Remote-User']"
          - name: groups
            expression: "claims['X-Remote-Groups']"
          - name: name
            expression: "'X-Remote-Name' in claims ? claims['X-Remote-Name'] : variables.username"

        validations:
          - expression: "variables.username.endsWith('@example.com')"
            message: "email domain not allowed"
          - expression: "size(variables.groups) > 0"
            message: "user must belong to at least one group"

        profile:
          name: "variables.name"

        impersonation:
          username: "variables.username"
          groups: "variables.groups.split(',')"
```
  Variables are evaluated in declaration order and may contain any CEL-supported value. Later expressions can reference earlier variables.

  ## Security

  - Authentication headers are accepted only from configured trustedProxies.
  - X-Forwarded-For and Forwarded are never used to establish trust.
  - CEL evaluation and validation failures return a generic unauthorized response.
  - Request headers are not persisted or exposed as provider metadata.
  - Kubernetes access continues to use the CEL-derived impersonated identity and existing RBAC enforcement.

  The Web UI must be reachable only through the trusted reverse proxy, which must remove client-supplied identity headers before setting authenticated values.

  ## Testing

  - Added trusted IPv4, IPv6, and CIDR tests.
  - Added trusted-peer and spoofed-forwarding rejection tests.
  - Added CEL variable, validation, profile, and impersonation tests.
  - Added variable chaining and profile fallback coverage.
  - Added string-variable group splitting coverage.
  - Added configuration loading, validation, and legacy-field rejection tests.
  - make test passes.
  - make fmt lint passes.

  ## Documentation

  - Added a reverse-proxy authentication guide with an Nginx example.
  - Updated the Web Config API reference.
  - Updated user-management and least-privilege documentation.

  AI assistance was used during implementation and review (codex/gpt-5.6-sol).


matheuscscp:
I think we definitely want to support these use cases, my major concern with this PR is the config API proposed for the feature significantly deviating from what we currently have for OAuth. In OAuth we feed data to CEL expressions for validation and extraction of the user features (i.e. username, groups, name...). I think we can model this API very closely to how we do it for OAuth via CEL expressions. Instead of feeding JWT claims to the expressions, here we feed the HTTP headers. Pls propose an API that is as close as possible to OAuth.

kdihalas:
> I think we definitely want to support these use cases, my major concern with this PR is the config API proposed for the feature significantly deviating from what we currently have for OAuth. In OAuth we feed data to CEL expressions for validation and extraction of the user features (i.e. username, groups, name...). I think we can model this API very closely to how we do it for OAuth via CEL expressions. Instead of feeding JWT claims to the expressions, here we feed the HTTP headers. Pls propose an API that is as close as possible to OAuth.

Thanks for the feedback, will work on this and update the PR

kdihalas:
@matheuscscp updated the code to follow the same pattern oauth uses with CEL expressions. Also updated the PR description with the new configuration format.

matheuscscp:
Thanks, I'll take a look when I'm back from vacations next week

copethomas:
If this feature was added, it would be great to support JWT from the upstream via the user identity headers as well. To provide more trust between reverse proxies and Flux. For example IBM Webseal supports this feature https://www.ibm.com/docs/en/sva/11.0.2?topic=reference-jwt

PR #973, opened 2026-07-27
Defense in depth: input validation, resource caps, stricter defaults

stefanprodan:
This PR tightens how untrusted and oversized input is handled across the operator, CLI, MCP server and Web UI: explicit bounds on anything that can grow without limit, validation on values that arrive from outside, and stricter defaults on the Web UI API surface.


stefanprodan:
@SocketSecurity ignore-all you are hallucinating obfuscated code 😞

matheuscscp (approved):

PR #981, opened 2026-08-01
operator: optimize `inventory.Diff` with map lookup

### LINKED ISSUE #980: Optimize inventory.Diff with map lookup to reduce O(n*m) complexity
## Problem

The `Diff` function in `internal/inventory/inventory.go` uses a `versionOf` 
closure that performs a linear scan through `inv.Entries` for every item in 
the diff list. For an inventory with n entries and a diff list of m items, 
this is O(n*m).

## Proposed Fix

Build a version map once upfront (O(n)), then use O(1) map lookups per diff 
item, reducing overall complexity to O(n+m).

This is the same optimization applied to kustomize-controller in 
fluxcd/kustomize-controller#1720.

manoj-1407:
Closes #980

## What does this change do?

`Diff` used a `versionOf` closure that performed a linear scan through
`inv.Entries` on every call — once per item in the diff list. For an
inventory with n entries and a diff list of m items, this is O(n*m).

This change builds a version map once upfront (O(n)), then uses O(1)
map lookups per diff item, reducing overall complexity to O(n+m).

On clusters with large inventories this avoids redundant linear scans
on every reconciliation cycle where pruning is needed.

## Testing

Existing `Test_Inventory/diff_objects_in_inventory` passes without
modification — behavior is identical, only the lookup strategy changed.

## Notes

This is the same optimization applied to kustomize-controller in
fluxcd/kustomize-controller#1720.

Assisted-by: Claude/claude-sonnet-4-6

### DIFF
--- a/internal/inventory/inventory.go
+++ b/internal/inventory/inventory.go
@@ -111,13 +111,12 @@ func ListMetadata(inv *fluxcdv1.ResourceInventory) (object.ObjMetadataSet, error
 
 // Diff returns the slice of objects that do not exist in the target inventory.
 func Diff(inv *fluxcdv1.ResourceInventory, target *fluxcdv1.ResourceInventory) ([]*unstructured.Unstructured, error) {
-	versionOf := func(i *fluxcdv1.ResourceInventory, objMetadata object.ObjMetadata) string {
-		for _, entry := range i.Entries {
-			if entry.ID == objMetadata.String() {
-				return entry.Version
-			}
-		}
-		return ""
+	versionMap := make(map[string]string, len(inv.Entries))
+	for _, entry := range inv.Entries {
+		versionMap[entry.ID] = entry.Version
+	}
+	versionOf := func(objMetadata object.ObjMetadata) string {
+		return versionMap[objMetadata.String()]
 	}
 
 	objects := make([]*unstructured.Unstructured, 0)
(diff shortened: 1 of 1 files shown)


internal/inventory/inventory.go
```diff
@@ -109,15 +109,15 @@ func ListMetadata(inv *fluxcdv1.ResourceInventory) (object.ObjMetadataSet, error
 	return metas, nil
 }
 
+// Diff returns the slice of objects that do not exist in the target inventory.
 // Diff returns the slice of objects that do not exist in the target inventory.
```
stefanprodan: @manoj-1407 did you read the changes before pushing? Here your AI decided to duplicate the comment.

stefanprodan (changes requested):

internal/inventory/inventory.go
manoj-1407: Apologies, I overlooked it during review. Fixed. 
Thanks!

manoj-1407:
@stefanprodan — fixed in e188dbc. could you re-review and dismiss the change request when you get a chance?

manoj-1407:
@stefanprodan just a follow up to my previous comment. everything has been taken care of in the latest push. it was a human error, i overlooked while reviewing. can you re review this when possible.

stefanprodan:
This looks ok, please rebase with upstream main and squash the 2 commits, then force push. Make this PR show a single commit on top of latest main branch.

manoj-1407:
@stefanprodan done, rebased onto latest main, squashed to a single commit, force pushed.

stefanprodan (approved):
LGTM

Thanks @manoj-1407

PR #982, opened 2026-08-01
[web] Add CPU/Memory usage to workload dashboard

stefanprodan:
This PR adds CPU/Memory metrics scraping and aggregation for Deployment, StatefulSets and DaemonSets by integrating with Kubernetes Metrics API (requires metrics-server). 

UI changes:
- The workload dashboard gains a dedicated panel for monitoring resource usage of pods with charts covering the last 30 minutes. Pods approaching their CPU or memory limits are highlighted to signal throttling and OOM risk. The last rollout is marked on the graphs.
- The Flux resource dashboard gains a tab under "Managed objects" show the CPU/Mem aggregated metrics and per-workload values. The last reconciliation is marked on the graphs.

For Flux Operator the monitoring of a cluster with 20K pods costs 50Mib (ring buffer cache) with negligible CPU usage. The Kubernetes API acts as a proxy to metrics-server, with no pressure on etcd. Given that there is no performance impact to Kubernetes control plane, the collector is enabled by default when metrics-server is detected.
The metrics collector can be disabled in the web-config and the scraping interval can be set between 15-600 seconds (defaults to 60s).

⚠️ Security: The metrics API is queried from a dedicated Go routine using the privileged client. All web UI requests are served from an in-memory cache that stores the data for 30m. Users with access to a workload, can also see the CPU/Memory usage without the need to grant them read access to Kubernetes Metrics API.

## Workload dashboard 

flux-status-pod-metrics-dark

## Flux Resource dashboard

localhost_5173_resource_HelmRelease_cert-manager_cert-manager (1)


matheuscscp (approved):
LGTM! 🚀

PR #986, opened 2026-08-06
operator: fix workload identity for Azure DevOps

### LINKED ISSUE #985: Azure Devops (ADO) support for Entra WIF does not work for ResourceSetInputProvider pull requests
Experienced on v0.57.0

When using a ResourceSetInputProvider for AzureDevopsPullRequest type, flux operator fails when using ADO WIF with entra. `secretRef` is the only method I could get working (by creating a token manually on behalf of the provider). This worked while using the same git url; however, when using wif I got the following error message:

> failed to create Git provider: a Git repository URL is required for issuing Git credentials

Digging into the code, seems like there's a small gap.

[`getAzureDevOpsToken`](https://github.com/controlplaneio-fluxcd/flux-operator/blob/v0.57.0/internal/controller/resourcesetinputprovider_controller.go#L756-L793) builds its options without `auth.WithGitURL`, and [`auth.GetGitCredentials`](https://github.com/fluxcd/pkg/blob/auth/v0.56.0/auth/git.go#L81-L84) rejects a nil URL unconditionally.

This seems to be drift between this repo and fluxcd itself. fluxcd/pkg#1189 implemented a (great) refactor after wif was added to this project in #311; this project never had a test for this so it wasn't caught.

PadenZach:
Assisted-by: <claude/opus 5>

Resolves #985 

built and tested against a dev cluster; seems to resolve the issues I was having. 

Also considered adding the git url to the standard opts in getAuthOptions since both code commit and ado needs this (and theortically any git provider), but it seems like that function is also being used by a non-git provider as far as I can tell. For this reason and my newness to the project I decided to make a more conservative change.

### DIFF
--- a/internal/controller/resourcesetinputprovider_controller.go
+++ b/internal/controller/resourcesetinputprovider_controller.go
@@ -752,7 +752,9 @@ func (r *ResourceSetInputProviderReconciler) getGiteaToken(
 	return password, err
 }
 
-// getAzureDevOpsToken returns the appropriate AzureDevOps token by reading the secrets in authData.
+// getAzureDevOpsToken returns the appropriate AzureDevOps token, either the personal
+// access token from the secrets in authData, or an Entra ID token issued through
+// workload identity federation.
 func (r *ResourceSetInputProviderReconciler) getAzureDevOpsToken(
 	ctx context.Context, obj *fluxcdv1.ResourceSetInputProvider,
 	authData map[string][]byte) (string, error) {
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
LGTM! 🚀 

@PadenZach Thanks very much for finding and fixing this! I pushed a refactor commit making it harder to repeat this mistake in case we introduce another Git provider with workload identity support. Can you please build and test e2e one more time? Thanks! 🙏

PadenZach:
Built and ran this again locally, seems to work. did only a quick preliminary test to ensure our service account worked with wif.

matheuscscp:
When you say you ran this locally do you mean in an AKS cluster? There's a way to test WIF with a local kind cluster with GCP WIF, but not sure about Azure. You need a cluster that supports Azure WIF.

PadenZach:
Yeah, re-reading that I worded it poorly. Here's what I did:

- built the image locally
- pushed it up to my EKS (Not AKS) dev cluster, set the flux operator pod to use the new pre-release tag I built
- My flux instance has wif enabled using a service account
- deleted and recreated a rsip that points at my ADO instance to ensure there was not some accidentally token caching even after removing spec.secretRef (shouldnt have been possible since the token I tested with before these commits expired).
- Confirmed it was able to get PR info from ADO
- To check again, created a test commit and reconciled again, new commit shows up without any issues

eg:
```bash
$ kubectl get rsip/python-app-example-preview -o yaml -n flux-system | yq .status.exportedInputs
- app: python-app-example
  author: Zach Paden
  branch: feat/branch-deployments
  env: preview
  id: "177750"
  labels:
    - deploy-preview
  sha: a8096f2683be7f3b02ebba983e628f663e4a0189
  title: 'feat(k8s): enable branch deployments'
```

spec for the rsip is:
```yaml
spec:
  defaultValues:
    app: python-app-example
    env: preview
  filter:
    labels:
    - deploy-preview
    limit: 3
  type: AzureDevOpsPullRequest
  url: https://dev.azure.com/org/project/_git/repo
```

PR #997, opened 2026-08-13
Ability to select latest tags like Flux `ImagePolicy` in `ResourceSetInputProvider`

### LINKED ISSUE #321: Ability to select latest tags like Flux `ImagePolicy` in `ResourceSetInputProvider`
Flux `ImagePolicy` has proven very useful with the ability to select a latest tag from a set using semver, alphabetical and numerical sorting. Now that `ResourceSetInputProvider` has lots of providers for fetching both Git and OCI tags, it would be really good if it could do the same as `ImagePolicy`. Furthermore, following the principle of improved writing capability of `ResourceSets`, we should also strive to build something that can decrease the amount of boilerplate required by several `ImagePolicy` objects.

### Proposal

We propose the following new fields to be added:

- `.spec.filter.extractValue`. A template that can be expanded using the `.spec.filter.includeTag` regex matching result to extract a string from a tag for usage with sorting, order comparison and string matching (hence why the name of the field is so broad, it's not just for ordering). If not specified, the tag itself is used for these purposes. If `spec.filter.semver` is specified, the extracted value is used for matching the semver range.
- `.spec.filter.orderBy`. An order to sort tags with. Can be one of `SemVer`, `Alphabetical`, `ReverseAlphabetical`, `Numerical` or `ReverseNumerical`. When `.spec.filter.semver` is specified, this field will immediately assume the value `SemVer` and no other values can be assigned. When `.spec.filter.semver` is not specified, the default value for this field will be `ReverseAlphabetical` (this default behavior is introduced [here](https://github.com/controlplaneio-fluxcd/flux-operator/pull/320)).
- `.spec.filter.extractGroup`. A template that can be expanded using the `.spec.filter.includeTag` regex matching result to extract a string from a tag for defining its group. The extracted string will serve as a group key. Tags yielding the same group key will be placed in the same group. If not specified, only a single group with all the tags is defined. When the scan results in multiple groups, `.spec.filter.limit` will apply to each group individually. For example, if the limit is 1 and the scan results in 10 groups of tags, 10 tags will be exported in the status. If the limit is 3 and the scan results in 10 groups, up to 30 tags will be exported. Additionaly, when the scan results in multiple groups, the tags will be exported in the alphabetical order of the group keys, then inside each group with the order defined by `.spec.filter.orderBy`.

All this logic can also be applied for branches, but we leave this out-of-scope for now. None of the new fields have `tag` or `branch` on them, so they can be supported for branches in the future (alongside `.spec.filter.includeBranch`).

### Examples

Export a single latest tag according to the default order (`ReverseAlphabetical`):

```yaml
spec:
  filter:
    limit: 1
```

Export a single latest tag according to SemVer:

```yaml
spec:
  filter:
    orderBy: SemVer
    limit: 1
```

Export a single latest tag according to a SemVer range:

```yaml
spec:
  filter:
    semver: ">=1.0.0 <2.0.0" # this forces the value SemVer for .orderBy
    limit: 1
```

Export a single latest tag according to SemVer extracting the version with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-(v.+)$"
    extractValue: $3
    orderBy: SemVer
    limit: 1
```

Export a single latest tag according to a SemVer range extracting the version with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-(v.+)$"
    extractValue: $3
    semver: ">=1.0.0 <2.0.0" # this forces the value SemVer for .orderBy
    limit: 1
```

Export a single latest tag according to a timestamp extracting it with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-([0-9]+)$"
    extractValue: $3
    orderBy: ReverseNumerical
    limit: 1
```

Export one latest tag per group according to SemVer, extracting the version and group key with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-(v.+)$"
    extractGroup: $1-$2
    extractValue: $3
    orderBy: SemVer
    limit: 1
```

Export one latest tag per group according to a timestamp, extracting the timestamp and group key with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-([0-9]+)$"
    extractGroup: $1-$2
    extractValue: $3
    orderBy: ReverseNumerical
    limit: 1
```

Export at most 3 latest tags per group according to SemVer, extracting the version and group key with a regex:

```yaml
spec:
  filter:
    includeTag: "^([a-z]+)-([a-z]+)-(v.+)$"
    extractGroup: $1-$2
    extractValue: $3
    orderBy: SemVer
    limit: 3
```

anandshashijoshi:
Closes: #321

feat(filter): Add orderBy flag to control tag sort order

Adds an orderBy field to ResourceSetInputFilter to allow ascending or
descending sort order when filtering tags. The filter pipeline now:

1. Matches tags with pattern regex
2. Extracts sortable keys using extract template
3. Sorts by extracted value or semver (asc/desc per orderBy flag)
4. Applies limit to result set

The orderBy flag defaults to 'desc' to preserve backward compatibility.
Supported values are 'asc' and 'desc'.

This enhancement applies to all tag provider types:
- GitHubTag, GitLabTag, AzureDevOpsTag, AWSCodeCommitTag, GiteaTag
- OCIArtifactTag, ACRArtifactTag, ECRArtifactTag, GARArtifactTag

Extends support for custom regex-based tag filtering that was added in
feat/add_customreg_filter.

Signed-off-by: anandshashijoshi
Co-authored-by: Copilot


anandshashijoshi:
Possible fix for #321 
@matheuscscp @stefanprodan is it possible to review this

stefanprodan:
@anandshashijoshi please do not merge commits from upstream, instead do a rebase and force push. We'll get to review this PR next week. Thanks

matheuscscp:
@anandshashijoshi The API implemented here is not the API described in #321.

anandshashijoshi:
As noted in the comments, this implements the same filtering logic as the Image Policy (imagePolicy CRD). It is especially beneficial for teams that do not follow SemVer for image tags and instead rely on custom tagging schemes.

matheuscscp:
I'm against merging the exact API of ImagePolicy. The API I proposed in #321 covers all of ImagePolicy and more. It's tailored for RSIP.

anandshashijoshi:
Okay, I understand. Let me prepare a new commit/PR  with the other changes that were mentioned in #321

anandshashijoshi:
@matheuscscp this should cover all the scenarios for custom tags

docs/web/web-least-privilege-rbac.md
matheuscscp, line 1: Please remove the changes in this file

matheuscscp (commented):
Before I read code, what's the answer this PR is giving to this question I put on the issue description?

> **Open question**: If we use `.spec.filter.limit` inside each group individually like suggested above, how do we limit the amount of groups? Does it even make sense limiting the amount of groups? By default, there's only a single group. The only way for a large number of groups to exist is writing a `.spec.filter.extractGroup` template that yields many groups. An alternative to mitigate a large number of groups is making `.spec.filter.includeTag` or `.spec.filter.excludeTag` more restrictive.

anandshashijoshi:
There are currently no restrictions. I was considering introducing a global maxTags / maxGroups limit, but that would push the development down a different path. For now, users can narrow their includeTag / excludeTag criteria when using grouping.

matheuscscp:
Cool, I'm gonna read the answer and resume the PR review next Monday when I'm back home, thanks!

anandshashijoshi:
@matheuscscp anything missing from this PR ?

matheuscscp:
Sorry @anandshashijoshi, we've been very busy, I may get to this PR next week with some luck

PR #1005, opened 2026-08-21
web: compute inventory status using `healthCheckExprs`

### LINKED ISSUE #1002: web: inventory status does not use the Kustomization's spec.healthCheckExprs, so custom resources show Unknown while the Kustomization is Ready
## Summary

A Kustomization's `spec.healthCheckExprs` are honoured by kustomize-controller but are not consulted
when the Flux UI computes the status of that Kustomization's **managed objects / inventory** rows.

The effect is two disagreeing health verdicts on one screen. The Kustomization's own condition says
`Ready: True`, derived from the CEL expressions the user supplied. Each inventory row below it is
computed independently, without those expressions — so a custom resource kstatus cannot assess shows
`Unknown`, however precisely its health has already been described on the Kustomization.

This is display-only; reconciliation is correct throughout. But it costs the inventory table its
value as a health signal for exactly the resources `healthCheckExprs` was introduced to cover, and it
makes the UI self-contradictory: the CRD the user went to the trouble of writing CEL for is the one
rendered as `Unknown`.

## Where this happens

`computeObjectStatus` in [`internal/web/inventory_objects.go`](https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/internal/web/inventory_objects.go)
(added in #932, merged 2026-06-30) branches three ways, and its own doc comment states the outcome:

```go
//   - Flux and Flux Operator kinds use the Ready-condition reader.
//   - Workloads (Deployment/StatefulSet/DaemonSet/CronJob) use the workload status
//     logic (kstatus + CronJob/apps refinements), matching the Workloads tab.
//   - Every other kind uses kstatus; an object it cannot assess yields "Unknown".
```

```go
	default:
		res, err := status.Compute(obj)
		if err != nil {
			return reporter.StatusUnknown, fmt.Sprintf("Failed to compute status: %s", err.Error())
		}
		return string(res.Status), res.Message
```

So `Unknown` for an unassessable object is deliberate, and the failure reason is already returned as
the row's message — no complaint there. The gap is that the parent Kustomization is never consulted:
nothing in this path reads `spec.healthCheckExprs`, even though the object being rendered is a member
of that Kustomization's inventory and the user has already declared how to judge it.

## Why this is general

`healthCheckExprs` exists because kstatus cannot assess every custom resource. Its generic path looks
for the `Reconciling` / `Stalled` condition convention and otherwise compares `metadata.generation`
against `status.observedGeneration`; any CRD that reports health some other way falls outside it — a
`phase`-only status, non-standard condition types, a missing or non-`int64` `observedGeneration`, a
status written by a controller that never sets `observedGeneration` at all.

For all of those the user's remedy is the same, and it already works at the controller level: declare
`healthCheckExprs` on the Kustomization. Since fluxcd/pkg#1187 (kustomize-controller `v1.9.0`,
Flux 2.9) kustomize-controller even falls back to those expressions specifically when kstatus fails
to assess a resource. The information needed to render these rows correctly is therefore present in
the parent object the UI is already displaying — it is just not used for the rows.

Argo Rollouts is the case we hit; the report is not specific to it.

## Our case, as a concrete example

`argoproj.io/v1alpha1` `Rollout` (argo-rollouts `v1.8.3`) declares `status.observedGeneration` as a
**string** in its published CRD, not the `int64` the API conventions call for:

```console
$ kubectl get crd rollouts.argoproj.io -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.observedGeneration}'
{"type":"string"}
```

Live object — none of the condition types match the `Reconciling` / `Stalled` convention either, so
assessment falls through to the generation comparison that the string then stops:

```yaml
metadata:
  generation: 12            # int
status:
  observedGeneration: '12'  # string
  phase: Paused
  conditions:
    - type: Available    status: "True"
    - type: Healthy      status: "False"
    - type: Completed    status: "False"
    - type: Progressing  status: Unknown
    - type: Paused       status: "True"
```

With these expressions on the Kustomization, kustomize-controller assesses the Rollout correctly and
the Kustomization goes `Ready: True`:

```yaml
  healthCheckExprs:
    - apiVersion: argoproj.io/v1alpha1
      kind: Rollout
      inProgress: status.phase == 'Progressing'
      failed: status.phase == 'Degraded'
      current: status.phase in ['Healthy', 'Paused']
```

The inventory table still shows that same Rollout as `Unknown`, with a `Failed to compute status`
message from the `status.Compute` error path above. Every other row in the same inventory
(Deployment, Service, ConfigMap, ServiceAccount, HTTPRoute) is assessed normally.

## Steps to reproduce

1. Have a `Kustomization` whose inventory contains a custom resource kstatus cannot assess natively —
   an `argoproj.io/v1alpha1` `Rollout` from argo-rollouts `v1.8.3` is one ready-made example.
2. Add `spec.healthCheckExprs` for that kind, and confirm the Kustomization reports `Ready: True`.
3. Open the Kustomization in the Flux UI and look at the managed-objects table.

**Actual:** the row for that resource shows `Unknown`.

**Expected:** the row reflects the verdict the `healthCheckExprs` produce — the same verdict the
Kustomization's own condition is already based on.

## Suggested fix

1. **In `computeObjectStatus`, evaluate the parent Kustomization's `spec.healthCheckExprs` for the
   object's `apiVersion`/`kind`** before or instead of falling back to `Unknown`, so the table agrees
   with the controller that already applies them. This needs no per-CRD knowledge in the UI, and
   matches what kustomize-controller does since fluxcd/pkg#1187.
2. **Optionally, do not lose a whole assessment to one non-conforming field** — where the generation
   comparison is impossible (e.g. a numeric string), coerce it or skip that check and fall through to
   the remaining ones. This helps when no expressions have been declared, but (1) is the fix that
   covers the class.

## Environment

| | |
|---|---|
| flux-operator (web) | `v0.58.0` (`ghcr.io/controlplaneio-fluxcd/flux-operator:v0.58.0`, installed via the `flux-operator` Helm chart) |
| kustomize-controller | `v1.9.4` |
| source-controller | `v1.9.4` |
| Kubernetes | `v1.34.9` |
| argo-rollouts (example CRD) | `v1.8.3` |

Also present in `v0.58.1` as far as I can tell from the changelog — dependency bumps plus an
`inventory.Diff` optimization, nothing touching status computation.

## Additional context

- Fixing the example at its source (argo-rollouts emitting `int64`) would be correct but would not
  address the general problem: any other CRD outside the kstatus conventions lands in the same place,
  with its `healthCheckExprs` equally unused by the table.
- I did not find an existing issue for this — searched this repo for `healthCheckExprs`, `kstatus`,
  and `inventory` + `status`, and `fluxcd/pkg` for the string `observedGeneration` case. #932 is the
  PR that introduced the status column; #833 is the request it closed.

stefanprodan:
If `.spec.healthCheckExprs` is declared on the parent Kustomization or HelmRelease, compute the inventory status using the CEL expressions in the same way Flux controllers do. (fix: #1002)

Show progressing status during Flux instance upgrades in the cluster dashboard, previously during a rollout components were reported as failing. (bug fix)

In addition, this PR includes major updates to web deps: vite 8, vitest 4, eslint 10, jsdom 30, js-yaml 5 and jest-dom 7.


matheuscscp (approved):
LGTM! 🚀 

Nice one!

Follow-up: This PR is re-implementing a (tiny) core Flux logic: the precedence order when a group healthcheck is defined together with the group-kind one. Ideally we should refactor `runtime/cel` to expose or decouple this tiny logic from `StatusReader`, so flux-operator can just reuse / import.

PR #1008, opened 2026-08-24
mcp: add field selection to `get_kubernetes_resources` tool

### LINKED ISSUE #894: [mcp-server] "Server side" tools result filtering
Would it be feasible to implement some sort of optional tool results filtering/field selection in the MCP server?

Concrete example: I've written a skill that reviews status of `HelmReleases` in my cluster, reporting their status and checking for newer versions outside their current version constraints, presenting a nice little report. Currently, the skill uses a subagent to fetch `HelmRelease` data, trim each one down do 10 JSON fields and return that to the main agent, to keep the size of the main context under control.

The subagent uses up ~14k output tokens fetching `HelmRelease` data for my small cluster with only 13 `HelmReleases`. In some cases, `HelmRelease` fields like `status.history` and `status.inventory` take up more than half of YAML, and are completely irrelevant for version review. It would be a nice saver if those could be filtered out by the MCP server.

stefanprodan:
Add the optional fields parameter to `get_kubernetes_resources` which accepts kubectl JSONPath expressions to reduce the size of the result when listing many resources or when only a few fields are relevant.

When `fields` is set, each resource is reduced to its `apiVersion`, `kind`, `metadata.name` and `metadata.namespace`, along with the values selected by the expressions.

For example, the fields `["spec.chart.spec.version", "status.conditions[?(@.type==\"Ready\")].message"]` return:

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: podinfo
  namespace: apps
spec.chart.spec.version: 6.x
status.conditions[?(@.type=="Ready")].message: Helm upgrade succeeded for release apps/podinfo.v3 with chart podinfo@6.9.0
```

Closes: #894

### DIFF
--- a/cmd/mcp/k8s/actions_test.go
+++ b/cmd/mcp/k8s/actions_test.go
@@ -116,6 +116,7 @@ func TestAnnotateResource(t *testing.T) {
 					"app.kubernetes.io/name=flux",
 					0,
 					false,
+					nil,
 				)
 				g.Expect(rErr).NotTo(HaveOccurred())
 				g.Expect(result).To(ContainSubstring(tt.value))
@@ -217,6 +218,7 @@ func TestDeleteResource(t *testing.T) {
 					"app.kubernetes.io/name=flux",
 					0,
 					false,
+					nil,
 				)
 				g.Expect(rErr).NotTo(HaveOccurred())
 				g.Expect(result).NotTo(BeEmpty())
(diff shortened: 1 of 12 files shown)


stefanprodan:
@matheuscscp we can ignore CodeQL, it will take some time till they switch to Go 1.27.

matheuscscp (approved):
LGTM! 🚀 

Should really help reduce response sizes!

PR #1009, opened 2026-08-25
mcp: advertise server instructions to MCP clients

stefanprodan:
Add server instructions and update the docs and index.

Reduce the `tools/list` context usage by 16%


matheuscscp (approved):
LGTM! 🚀

PR #1010, opened 2026-08-26
mcp: Add `diff_kubernetes_manifest` tool 

stefanprodan:
This RR adds a read-only tool to the MCP Server which allows AI Agents to preview changes made to manifests in a Flux repository. 

The agent is instructed to produce the manifests with changes and call the tool with the multi-doc YAML:

- For a Kustomization, run `kustomize build <path> --load-restrictor=LoadRestrictionsNone`.
- For a ResourceSet, run `flux-operator build resourceset` with an inputs file made from the live ResourceSetInputProvider `status.exportedInputs`.
- For a HelmRelease, run `helm template <releaseName> <chart> -n <releaseNamespace> --include-crds`, with `spec.values` merged over the referenced `valuesFrom` values; `postRenderers` are applied by the tool.

### Tool Specification

Diffs a multi-document YAML manifest against the cluster using server-side apply dry-run with
the Flux controller's field manager. Use it to preview a direct apply or the effect of a GitOps
commit before pushing it to the repo. The owner transforms are applied before the dry-run:
Kustomization build options and `postBuild` substitutions, ResourceSet `copyFrom`, HelmRelease
`postRenderers` and release metadata, and the `commonMetadata` and owner labels of all kinds.

**Parameters:**

When `yaml_path` is advertised, exactly one of `yaml_content` or `yaml_path` must be specified;
otherwise, `yaml_content` must be specified.

- `yaml_content` (optional): The complete multi-document YAML build output to diff
- `yaml_path` (optional): The absolute path to a local multi-document YAML build output file;
  mutually exclusive with `yaml_content`, advertised only when the server runs locally over the
  `stdio` transport, with a 64 MiB file size limit
- `flux_object` (optional): The YAML definition of the Kustomization, ResourceSet, or HelmRelease
  that applies the manifest, as it will exist after the change
- `owner_kind`, `owner_name`, `owner_namespace` (optional): A reference to an existing owner;
  all three parameters must be specified together

**Output:**

Each manifest object is reported as `create`, `recreate`, `update`, `unchanged`, `skipped`, or
`error`; inventory objects absent from the manifest are reported as `delete`. Updates include an
RFC 6902 JSON patch in YAML form, with Secret values masked. The final summary line counts every
state. For example:

```text
Diff for Kustomization/apps-staging/backend (field manager: kustomize-controller, prune: enabled)

ConfigMap/apps-staging/backend-7f2k9c4mbt create
Deployment/apps-staging/backend update
- op: replace
  path: /spec/template/spec/containers/0/envFrom/0/configMapRef/name
  value: backend-7f2k9c4mbt
Service/apps-staging/backend unchanged

Not in the manifest (pruned if the manifest is complete):
ConfigMap/apps-staging/backend-m759gh88kd delete

Summary: 1 create, 1 update, 1 unchanged, 1 delete
```

**Limitations:**

- Owners targeting remote clusters are not supported.
- Kustomization `spec.components` is not applied.
- ResourceSet `checksumFrom` and `convertKubeConfigFrom` are not resolved.
- SOPS-encrypted objects are compared by key set only.
- Kustomization `namePrefix` and `nameSuffix` are added to the names in the build output, while
  kustomize-controller replaces the ones set in the source `kustomization.yaml`; when both are set,
  the diff reports different names than the cluster.
- Kustomization `buildMetadata` is not applied; the labels and annotations it adds are ignored.
- Managed-fields cleanup is not simulated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ test-olm-e2e: build-olm-manifests operator-sdk ## Test OLM manifests for current
 CLI_IMG ?= ghcr.io/controlplaneio-fluxcd/flux-operator-cli:latest
 
 .PHONY: cli-test
-cli-test: tidy fmt vet ## Run CLI tests.
+cli-test: tidy fmt vet envtest ## Run CLI tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./cmd/cli/... -v
 
 .PHONY: cli-build
@@ -211,7 +211,7 @@ mcp-build-search-index: ## Build search database for MCP docs tool.
 	@echo "MCP search database built successfully"
 
 .PHONY: mcp-test
-mcp-test: tidy fmt vet ## Run MCP tests.
+mcp-test: tidy fmt vet envtest ## Run MCP tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./cmd/mcp/... -v
 
 .PHONY: mcp-build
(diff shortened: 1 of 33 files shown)


matheuscscp (approved):
LGTM! 🚀 

Nice tool!

PR #1011, opened 2026-08-26
operator: honor reconcile   disabled annotation on in-cluster objects

### LINKED ISSUE #1004: ResourceSet: honor fluxcd.controlplane.io/reconcile: disabled on live objects, not just on the rendered template
I wanted to stop a ResourceSet from overwriting one of the resources it manages so I could make a manual adjustment, so I annotated the live object with `fluxcd.controlplane.io/reconcile: disabled` expecting the operator to leave it alone. 
The next reconcile put it back. The only workarounds I found are suspending the whole ResourceSet or editing the template source. 

The annotation is only evaluated on the rendered template during the build [here](https://github.com/controlplaneio-fluxcd/flux-operator/blob/v0.58.1/internal/builder/resourceset.go#L66) and [here](https://github.com/controlplaneio-fluxcd/flux-operator/blob/v0.58.1/internal/builder/resourceset.go#L96) the apply path never checks the live object. It also doesn't just get ignored, it gets stripped, since the apply cleanup takes ownership from the kubectl field managers.

`fluxcd.controlplane.io/prune: disabled` already works the way I expected, it's checked against the live object during garbage collection, so the two annotations behave differently depending on which path you're in.

The fix looks like a small one: set ExclusionSelector on the apply options in resourceset_controller.go to match the reconcile annotation, the same way kustomize-controller does. Skipped objects still get recorded in the inventory so they won't be collected, and the existing tests pass, I can add some more of course. 

Let me know if this makes sense and I can open a PR. I also might be missing something @stefanprodan @matheuscscp !

dgunzy:
Closes #1004

 Sets `ExclusionSelector` on the apply options so in-cluster objects annotated with `fluxcd.controlplane.io/reconcile: disabled` are skipped, the same way kustomize-controller does it. Skipped objects are still recorded in the changeset, so they stay in the inventory and are not garbage collected, and removing the annotation resumes the reconciliation and corrects the drift.

I also skipped failed Jobs annotated with `recreateOnFailure`, since those are deleted before the apply and would come back from the template regardless of the annotation. Happy to pull that out into its own PR if you would rather keep this one to the apply path.

Added envtest coverage for both the frozen and the non frozen resource plus the Job case, and extended the ResourceSet e2e to freeze and unfreeze a live OCIRepository. Let me know if you want more or different coverage @stefanprodan @matheuscscp

### DIFF
--- a/docs/api/v1/resourceset.md
+++ b/docs/api/v1/resourceset.md
@@ -672,6 +672,25 @@ spec:
 
 In the above example, the `ServiceAccount` resource is generated only for the `team1` tenant.
 
+#### Freezing a resource in-cluster
+
+The `fluxcd.controlplane.io/reconcile: disabled` annotation is also honored when set on a
+live object in the cluster. When the operator finds the annotation on an existing resource,
+it skips it from server-side apply and leaves any manual changes in place. 
+
+This allows freezing a single resource without suspending the whole ResourceSet:
+
+```shell
+kubectl annotate ocirepository/podinfo fluxcd.controlplane.io/reconcile=disabled
+```
+
+Removing the annotation resumes the reconciliation, and any manual changes are
+reverted on the next apply:
+
+```shell
+kubectl annotate ocirepository/podinfo fluxcd.controlplane.io/reconcile-
+```
+
 #### Built-in input fields
 
 When computing all the input sets for generating the resource matrix, the operator
--- a/internal/controller/resourceset_controller.go
+++ b/internal/controller/resourceset_controller.go
@@ -545,6 +545,12 @@ func (r *ResourceSetReconciler) apply(ctx context.Context,
 	applyOpts.ForceSelector = map[string]string{
 		fluxcdv1.ForceAnnotation: fluxcdv1.EnabledValue,
 	}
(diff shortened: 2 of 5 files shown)


docs/api/v1/resourceset.md
```diff
@@ -672,6 +672,26 @@ spec:
 
 In the above example, the `ServiceAccount` resource is generated only for the `team1` tenant.
 
+#### Freezing a resource in-cluster
+
+The `fluxcd.controlplane.io/reconcile: disabled` annotation is also honored when set on a
+live object in the cluster. When the operator finds the annotation on an existing resource,
+it skips it from server-side apply and leaves any manual changes in place. The resource is
+kept in the ResourceSet inventory, so it is not subject to garbage collection.
```
stefanprodan: "The resource is kept in the ResourceSet inventory, so it is not subject to garbage collection."

I would remove this, makes no sense to me. Ai Slop?

dgunzy: Yes sorry - I can remove that part for sure.

matheuscscp (approved):
LGTM! 🚀

stefanprodan (approved):

PR #1013, opened 2026-08-26
mcp: remove dead code for enabling specific tools

matheuscscp:
No production code is using the argument `enabled` of `toolbox.NewManager`


stefanprodan (approved):

stefanprodan:
I may add this back if the number of tools grows since they increase the context by a lot, GitHub MCP allows picking tools via a cmd flag.

PR #1015, opened 2026-08-27
mcp: consolidate reconcile requests under `reconcile_flux_resource` tool

stefanprodan:
Replace the `reconcile_flux_source`, `reconcile_flux_kustomization`, `reconcile_flux_helmrelease` and `reconcile_flux_resourceset` tools with a single `reconcile_flux_resource` tool that takes `apiVersion`, `kind`, `name`,
`namespace` and an optional `with_source` flag, matching the shape of the suspend and resume tools.

The generic tool works for any Flux kind, including the `ResourceSetInputProvider`, `FluxInstance`, image automation and `Receiver` objects that had no reconcile tool before. 

In addition, we made the `apiVersion` input optional for the reconcile, suspend and resume tools. The MCP resolves the apiVersion on its own, reducing the amount of calls an agent must make before calling these tools.


matheuscscp (approved):
LGTM! 🚀 

Nice one!

PR #1017, opened 2026-08-27
mcp: extend `get_kubernetes_logs` tool to workloads 

stefanprodan:
Accept a workload kind (Deployment, StatefulSet, DaemonSet, CronJob or Job) in addition to Pod, so agents can read application logs directly from the workload name found in a Flux resource inventory, without first resolving selectors, pods and containers.

Tested with Qwen3.6 and Ornith-1.5 run locally. This change makes a huge difference when using agents to debug broken clusters, 3x less tool calls, 50K+ tokens saved per sub-agent.


matheuscscp (approved):
LGTM! 🚀 

Nice optimization!

PR #1019, opened 2026-08-28
mcp: add `get_kubernetes_events` tool 

stefanprodan:
Kubernetes events were only attached to Flux resources, leaving agents without ImagePullBackOff, FailedMount, etc or probe failures for workloads.

The new read-only tool lists core events with field selectors and a regex expression, so agents can grep server-side to avoid filling the context with unwanted events.

In addition, the `get_kubernetes_logs` tool gains `since` and `grep` filters.

### DIFF
--- a/cmd/mcp/k8s/events.go
+++ b/cmd/mcp/k8s/events.go
@@ -6,12 +6,16 @@ package k8s
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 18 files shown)


matheuscscp (approved):
LGTM! 🚀

PR #1021, opened 2026-08-29
mcp:  `search_flux_docs` with fuzzy search over chunked index

stefanprodan:
The `search_flux_docs` gets a new contract (`query`, `path`, `limit`) and returns one text item per section with its title, path and line range. A new `read_flux_doc` tool reads a page by heading anchor or line offset, capped at 30 KB per call, so agents can read the surrounding context of a hit without loading the whole page.

The new search tools reduce context bloat by **17x**, measured with Fable and Qwen doing research on Flux dependency management and OCI artifacts.


cmd/mcp/main.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	"github.com/controlplaneio-fluxcd/flux-operator/cmd/mcp/k8s"
 	"github.com/controlplaneio-fluxcd/flux-operator/cmd/mcp/toolbox"
+	"github.com/controlplaneio-fluxcd/flux-operator/cmd/mcp/toolbox/library"
```
matheuscscp: This whole package is about docs right? Should be called `docs`, not `library`, very confusing

matheuscscp (approved):
LGTM! 🚀

cmd/mcp/main.go
stefanprodan: Good call! I've refactored the whole package

PR #1028, opened 2026-09-09
web: Fix Health checking badge for Kustomizations with healthChecks

### LINKED ISSUE #1003: web: "Health checking" reported as Disabled for a Kustomization that uses spec.healthChecks
## Summary

In the managed-objects panel, the **Health checking** badge reads `Disabled` for a `Kustomization`
that has a populated `spec.healthChecks` list. Those Kustomizations do run health checks — the
controller logs `Running health checks for revision ...` and publishes a `Healthy` condition — so the
badge contradicts the object's own status shown on the same page.

The badge only turns `Enabled` for `spec.wait: true`.

## Cause

[`web/src/components/dashboards/resource/ManagedObjectsPanel.jsx:91`](https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/web/src/components/dashboards/resource/ManagedObjectsPanel.jsx#L91):

```js
  const healthCheckEnabled = useMemo(() => {
    const k = resourceData.kind
    if (k === 'Kustomization' || k === 'FluxInstance' || k === 'ResourceSet') {
      return resourceData.spec?.wait === true
    }
```

`spec.healthChecks` is never consulted. For a `Kustomization`, health checking is enabled by
**either** `spec.wait: true` **or** a non-empty `spec.healthChecks` — kustomize-controller runs the
health assessment in both cases (and per the CRD, `wait: true` additionally causes `healthChecks` to
be ignored).

For contrast, `pruningEnabled` immediately above reads `spec.prune` correctly; only the health flag
is narrow.

## Steps to reproduce

1. Apply a `Kustomization` with a non-empty `spec.healthChecks` and no `spec.wait`.
2. Confirm health checks actually run — `Healthy` condition present, and
   `Running health checks for revision ...` in the kustomize-controller log.
3. Open that Kustomization in the UI, managed-objects panel.

**Actual:** Health checking — `Disabled`.
**Expected:** Health checking — `Enabled`.

## Suggested fix

```js
    if (k === 'Kustomization') {
      return resourceData.spec?.wait === true
        || (resourceData.spec?.healthChecks?.length ?? 0) > 0
    }
```

Note `spec.healthCheckExprs` should **not** count on its own: those expressions are evaluated only
when `wait` or `healthChecks` is set, so a Kustomization carrying only `healthCheckExprs` genuinely
performs no health checking. Verified on a live object — with `healthChecks: []` and no `wait`, the
reconcile completes in ~200 ms with no health-check phase and no `Healthy` condition; adding
`wait: true` produces `Healthy: True — Health check passed in 71ms` from the same expressions.

## Possibly also, in the same branch

`FluxInstance.spec.wait` is `*bool` with `+kubebuilder:default:=true`, and `GetWait()` returns `true`
when the field is `nil` — so the controller's notion of enabled is "not explicitly false", while the
UI requires `=== true`. In practice the CRD default materializes the field at admission so the two
agree; it would diverge only for an object whose `wait` is absent. Worth a glance while touching this
predicate, but it is not what this report is about.

`ResourceSet.spec.wait` is a plain `bool` and the type has no `healthChecks` field, so the current
check is correct there.

## Environment

| | |
|---|---|
| flux-operator (web) | `v0.58.0` (`ghcr.io/controlplaneio-fluxcd/flux-operator:v0.58.0`) |
| kustomize-controller | `v1.9.4` |
| Kubernetes | `v1.34.9` |

Code referenced above is `main` at the time of writing; `v0.58.1` does not touch this file.

davidgrldo:
## Problem

In the managed objects panel, the **Health checking** badge reads `Disabled` for a Kustomization with a populated `spec.healthChecks` but no `spec.wait`. The controller does run health checks in that case (it logs `Running health checks for revision ...` and publishes the `Healthy` condition), so the badge contradicts the object's own status.

Fixes #1003

## Cause

`ManagedObjectsPanel.jsx` computed the flag from `spec.wait` alone:

```js
if (k === 'Kustomization' || k === 'FluxInstance' || k === 'ResourceSet') {
  return resourceData.spec?.wait === true
}
```

`spec.healthChecks` was never consulted.

## Fix

For Kustomization, health checking is reported as enabled when **either** `spec.wait` is `true` **or** `spec.healthChecks` is non-empty (kustomize-controller runs the health assessment in both cases; with `wait: true` the explicit `healthChecks` are ignored, but health checking still applies). FluxInstance and ResourceSet keep the wait-only check, and HelmRelease is unchanged.

## Testing

- Added a regression test: Kustomization with `wait: false` and a non-empty `spec.healthChecks` renders the badge as `Enabled`. The assertion is scoped to the Health checking row (the previous pattern of asserting on the whole panel text would pass via the Pruning row).
- ```bash
  cd web && npm test
  ```
  All 2477 tests in 93 files pass; ESLint clean.


stefanprodan:
@davidgrldo are you using an AI? if so you must disclose this as per our [AI governance](https://github.com/controlplaneio-fluxcd/flux-operator/blob/main/CONTRIBUTING.md#ai-contribution-policy).

The issue you are tying to fix here is not settled, as I commented in there, I'm not Ok with the `Healthy` labeling for this.

davidgrldo:
Thanks for reviewing @stefanprodan.

**AI disclosure**: This PR was created with the assistance of AI (OpenCode / glm-5.3). I'll add the `Assisted-by` trailer per CONTRIBUTING.md before the final merge.

Regarding your feedback — I agree that `spec.healthChecks` is a legacy field. You mentioned two options for the badge label: **"Partial"** or **"Legacy"**. Which one would you prefer?

Once I have a clear direction on the label, I'll update the component, tests, and commit.

PR #1030, opened 2026-09-09
web: prevent panic on unsupported resource actions

gabriellavengeo:
A request with a user action that is globally valid but unsupported by the `POST /api/v1/resource/action` endpoint may cause `ActionHandler` to invoke `h.sendAuditEvent` with a nil `obj` argument, leading to nil pointer dereference and panic If audit is enabled. This PR makes `ActionHandler` return status 400 Bad Request for actions that are not among the supported reconcile, suspend, or resume.

### DIFF
--- a/internal/web/resource_action.go
+++ b/internal/web/resource_action.go
@@ -39,7 +39,21 @@ type ActionResponse struct {
 	Message string `json:"message"`
 }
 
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
LGTM! 🚀 

Thanks @gabriellavengeo!

PR #1034, opened 2026-09-10
web: make cache keys unambiguous

gabriellavengeo:
This PR changes the impersonation cache key format by base64-encoding the username and groups before concatenation and sha256-hashing the key, in order to avoid potential key ambiguity if the username or groups contain the `\n` delimiter .


matheuscscp (approved):
LGTM! 🚀 

Thanks @gabriellavengeo!

PR #1035, opened 2026-09-11
web: clarify RBAC roles and security boundary in the Web UI user management docs

gabriellavengeo:
- Clarifies the Web UI security model by documenting Kubernetes RBAC as the only security boundary, regardless of whether resources are managed by Flux
- Updates the Web UI predefined roles descriptions to more precisely reflect the exact RBAC definitions


matheuscscp (approved):
LGTM! 🚀

PR #1036, opened 2026-09-11
cli: Add `--distribution-artifact` flag to distro mirror

stefanprodan:
Allow overriding the Flux distribution manifests OCI artifact used by the distro mirror command. Allows pinning the artifact to a specific Flux Operator release for air-gapped use.

Example:

```sh
  # Mirror the exact images shipped with a specific Flux Operator release
  echo "${GITHUB_TOKEN}" | flux-operator distro mirror registry.example.com/flux \
    --distribution-artifact oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:v0.59.0 \
    --variant enterprise-distroless \
    --version 2.8.8 \
    --pull-token-stdin
```


matheuscscp (approved):
LGTM

PR #1039, opened 2026-09-11
mcp: make docs tests independent of index content

stefanprodan:
The embedded docs index is refreshed on every release, so tests that pin line numbers, totals, or specific ranked headings break as soon as the docs change. Derive the HelmRelease section bounds from the index at test time and assert that the top search hit covers both kinds instead of expecting fixed heading trails.


matheuscscp (approved):
LGTM

PR #1042, opened 2026-09-14
Flip "Progressing" icon so that it spins in the correct direction

Steffo99:
The flip was done with Inkscape, and then the file was saved as "Optimized SVG" with 5 significant digits.


matheuscscp:
@Steffo99 Can we have a preview of this change somehow?

Steffo99:
Here are two:

[Video 1](https://github.com/user-attachments/assets/00223bf0-2779-4d1b-807c-8853b79c986a)

[Video 2](https://github.com/user-attachments/assets/bf0494cf-63ae-4825-8b2d-37933ceeb92d)

PR #1045, opened 2026-09-15
mcp: Add tool annotations

stefanprodan:
Advertise `readOnlyHint`, `destructiveHint` and `idempotentHint` on every MCP tool, derived from the systemTools table so hosts see the same read/write split as --read-only.


matheuscscp (approved):
LGTM

